PR #116, opened 2015-04-29
Fix inline jumps

asandroq:
The cpp define for inline jumps was misspelled in gambit.h.in and
completely gone from config.h.in


feeley:
Thanks for the pull request but I had already fixed this a few minutes ago!

PR #118, opened 2015-04-30
Default character encoding for file paths is UTF-8.

jgesswein:
All file functions use the same macro for determining the type for a file path. The type on non-Windows systems is now "utf8" instead of "native". Function "directory-files" now works fine on OS X with the path "äöü".

There are also some macros for process names that might be considered as well (e.g. ___STREAM_OPEN_PROCESS_CE_SELECT). These names are probably also file names.

### DIFF
--- a/lib/os.c
+++ b/lib/os.c
@@ -1685,14 +1685,12 @@ ___SCMOBJ modification_time;)
 
 #ifdef USE_utimes
 
-#define ___TIMES_PATH_CE_SELECT(latin1,utf8,ucs2,ucs4,wchar,native) native
-
   if ((e = ___SCMOBJ_to_NONNULLSTRING
              (___PSA(___PSTATE)
               path,
               &cpath,
               1,
-              ___CE(___TIMES_PATH_CE_SELECT),
+              ___CE(___PATH_CE_SELECT),
               0))
       == ___FIX(___NO_ERR))
     {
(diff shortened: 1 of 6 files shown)


feeley:
Thanks for you pull request.  I have committed my own fix, which I had started working on a few days ago.  I'm sorry for the time you spent on your fix, but your help in improving Gambit is greatly appreciated!  I hope you can review my fix and let me know if there are remaining issues.

gambiteer:
Isn't there a way to link closing an issue to the commit that "fixes" it?  That would be nice instead of searching the list of recent commits.

jgesswein:
Hi Marc!

> Am 2015-05-04 um 16:49 schrieb Marc Feeley <notifications@github.com>:
> 
> Thanks for you pull request. I have committed my own fix, which I had started working on a few days ago. I'm sorry for the time you spent on your fix, but your help in improving Gambit is greatly appreciated! I hope you can review my fix and let me know if there are remaining issues.

I had a look at the differences for your fix and as far as I can see this should work.

However when I read about the option for selecting the encoding I thought this translates into a variable that is initialised with the system’s encoding (similar to those for terminal and file encoding).  All routines would refer to that variable.  This allows to change the encoding if the default is not correct for some reason or some stuff is not encoded properly.

BTW: Does it make sense to you to break down those files with system-specific code into system-spefic files?  For instance “os_files.c” could be split into “os_files_posix.c” and “os_files_windows.c” (and may-be other variants).  I think this would improve readability of system-specific code.

Cheers, Jürgen

PR #120, opened 2015-05-04
Improve AArch64 handling.

hrw:
AArch64 (64-bit ARMv8) is different architecture that ARM so let's treat
it that way.

It can also be big or little endian so check for little endian one is
added.


hrw:
Note that it still fails to run under Fedora.

hrw:
feeley: __aarch64__ is defined by gcc (iirc also by clang), __arm64__ is iOS compiler. No idea is there other environment supporting those cpus.

PR #148, opened 2015-07-27
** NOT A PULL REQUEST ** Sent out to discuss adding syntactic closure(hygienic) handling in compiler / interpreter

wsxiaoys:
Hi Marc I recently started reading the compiling sub-module in both compiler(_env.scm/_ptree*.scm) and interpreter (_eval.scm), and is interested in adding hygienic.

A roughly roadmap would be like:
* Document environment handling in both compiler / interpreter.
* Adding macro syntactic environment
* Adding syntactic closure transforming phase.
* Converting macros to its hygienic implementation (let, letrec, case ...)

As this is quite a complicated task I would like to have your permission and help :)


feeley:
Hello Meng.  Sorry for not responding sooner.  Can you explain how your pull request is supposed to fit in your plan to add hygiene?  I think the main issue is the addition of a new kind of "identifier" (to replace symbols) so that the name clashes possible with plain symbols can be avoided.  The function `##cte-lookup` (in lib/_eval.scm) needs to accept such identifiers for the "name" parameter.  Also, this might be a good time to improve the environment representation (currently a chain of bound names) so that it scales well for large files (a tree or hash table would be better).

wsxiaoys:
Hi Marc,
Sorry for not making it clear, this branch is mostly an experiment comparing different implementation of hygienic macros. So let's ignore that for now..

I mostly finished reading on compiling environment code in lib/_eval.scm, and still investigating the runtime environment handling and the code generation part. After that I plan to draft documents on following aspects:
1. Environment representation
In a hygienic situation, Compile time environment is mapping from Identifier to variable binding or macro binding at arbitrary level. 
Gambit's compile time environment(ctx) is simpler. All macro frame are added after top-level ctx, no macro binding in local frame.

2. Macro handling
Macro need to be able to access the lexical environment when invoking. Currently gambit's macro system doesn't allow such information.

Another issue is that I'm a bit confusing how macro is handled in an compiler (define-macro/define-syntax _ptree1.scm). From what i've read it looks like macro expansion is done with eval so not sure how these code are used.

PR #161, opened 2015-08-17
Fixes java field initialisation

astlouisf:
Adds two univ-defs transformations.

univ-defs-separate-field-initialisation splits fields into a declaration and an assignment statement.

univ-use-initialisation-methods groups each univ-defs-inits entry and generate a thread of initializing methods.

__init0() {
  /* assign */
 ___init1()
}


astlouisf:
I thought that, since we were working on the limit of method size, it would have been a better idea to limit the size of the size of the <clinit> (class init) method .

Anyways, I'm currently encapsulating the code inside classes and generating individual init calls inside static:

```
static class Gambit____init0 {
  static void init() {
    /* field initialisation */
  }
}

static class Gambit____init1 {
  static void init() {
    /* field initialisation */
  }
}
 /* ... */
static {
  Gambit____init0.init();
  Gambit____init1.init();
}
```

So except if you really want the option to not generate class, I will probably just push the class version.

feeley:
Are you using this approach because the limit on the number of constants is on each class?

Also, in many cases a Scheme program will have few constants, say less than 1000.  So it would be good to optimize this case to avoid creating very small (or empty?) initialization classes.  When the number of constants is small, the main class should contain the initializations (as was the case before).

Finally, the triple underscore is not very visually appealing and it doesn't eliminate the possibility of name clashes.  The module name should also be included in the name of the class (similar to the naming of basic block methods).

astlouisf:
> Are you using this approach because the limit on the number of constants is on each class?

Yes.

> Also, in many cases a Scheme program will have few constants, say less than 1000.

The problem isn't the number of Scheme constant but what Java will add to its constant pool.
From my understanding of [the java constant pool](http://blog.jamesdbloom.com/JVMInternals.html#constant_pool), each basic block initialization comes with its number of constants added to the main class constant pool. And we generate a lot of them.


> When the number of constants is small, the main class should contain the initializations (as was the case before).

This code already does it (just need to tweak the numbers!)
```
    (if (and (eq? (target-name (ctx-target ctx)) 'java)
             (> (+ (length (univ-defs-fields defs))
                      (length (univ-defs-inits defs)))
                100))
        (begin
         (set! defs (univ-defs-separate-field-initialisation ctx defs))
         (set! defs (univ-defs-use-initialisation-methods ctx defs 50))))
```
I've put the threshold to 100 but we could bump it to 1000. Then again, its hard to strike the right balance. Currently, univ-lib can't be compiled to use `fast-bignum-algorithm` because the initialization of the precomputed powers of 10 busts the method size limit when generating one initialization method per `univ-defs-init`. 


>Finally, the triple underscore is not very visually appealing and it doesn't eliminate the possibility of name clashes. The module name should also be included in the name of the class (similar to the naming of basic block methods).

Will do.

PR #163, opened 2015-08-24
Test for $target_os should be lowercase

jasperla:
This unbreaks OpenBSD/mips64 again after the spello was introduced in https://github.com/feeley/gambit/pull/29 .


juanfra684:
@feeley Can you commit this PR? The patch is obviously correct and was tested on OpenBSD/mips64 :)

feeley:
Sorry about that... it went off my radar.

PR #176, opened 2016-01-31
unit test for fx- and ##fx-

eval-apply:
Please let me know if this looks ok to you and I'll go ahead and attempt some more when I have a chance.


eval-apply:
Yes you are right sorry, I was rushing this last night.  I'll fix it and rename as you suggest.

eval-apply:
I've added a few more tests for some more fixnum procedures, please let me know what you think.

feeley:
Nice! Thanks!

PR #177, opened 2016-02-05
Fixnum tests

eval-apply:
I think that this should be all of the remaining public fixnum procedures, let me know if I've missed any and I'll update this pull request.  I tried to cover all the suitable cases I could think of but again if I've missed any let me know and I'll update them.

### DIFF
--- a/tests/unit-tests/01-fixnum/fxabs.scm
+++ b/tests/unit-tests/01-fixnum/fxabs.scm
@@ -1,0 +1,19 @@
+(include "#.scm")
+
+(check-eqv? (##fxabs 0) 0)
+(check-eqv? (##fxabs 1) 1)
+(check-eqv? (##fxabs 100) 100)
+(check-eqv? (##fxabs -100) 100)
+
+(check-eqv? (fxabs 0) 0)
+(check-eqv? (fxabs 1) 1)
+(check-eqv? (fxabs 100) 100)
+(check-eqv? (fxabs -100) 100)
+
+(check-tail-exn fixnum-overflow-exception? (lambda () (fxabs ##min-fixnum)))
+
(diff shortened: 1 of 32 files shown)


feeley:
Thanks for the many unit tests!  I've accepted the merge without requesting small changes because all of the unit tests are valid.  However, there is one thing I would like you to fix for all fixnum unit tests, the fact that floating point integers, such as 5.0 and 0.0, are never tested (they should result in type errors).

eval-apply:
Not a problem, for the tests I just submitted I added type-exceptions like this:

```scheme
(check-tail-exn type-exception? (lambda () (fxmodulo 1 0.5)))
```

Is that sufficient or would you prefer that the offending type be in both argument positions?

```scheme
(check-tail-exn type-exception? (lambda () (fxmodulo 1 0.5)))
(check-tail-exn type-exception? (lambda () (fxmodulo 0.5 1)))
```

feeley:
Ideally both arguments should be tested.  But as I mentionned you should also check for flonum integers.  For example:

```
(check-tail-exn type-exception? (lambda () (fxmodulo 1 0.5)))
(check-tail-exn type-exception? (lambda () (fxmodulo 0.0 1)))
```

So here both parameter positions are checked for flonums, and also in one case for integer flonum.

eval-apply:
Makes sense, I'll go through and ensure that all fixnum tests cover all the cases you suggest.

After which I was thinking of moving onto the flonum tests unless there is an area you would prefer done first.

feeley:
Sounds good.  Yes please move on to the flonum primitives.  There it is important to test limit cases such as division by 0.0 and by -0.0 and by +inf.0 and -inf.0 .

PR #191, opened 2016-03-19
Remove ws2_32 check other than MinGW

okuoku:
Do not link against Winsock2(ws2_32) to prevent conflict with Cygwin libc.

Should fix #99.

We can also remove
```
               AC_SEARCH_LIBS(timeBeginPeriod,winmm)
```
and
```
  AC_CHECK_HEADERS(ws2tcpip.h)
```
but these are not blocker for building on Cygwin.

PR #192, opened 2016-03-19
Disable nanosleep() on os_io.c for Cygwin

okuoku:
It seems nanosleep() may return ECHILD on Cygwin x86_64.

Cygwin build can be fixed with #191 and it would produce working `gsi` but to make test passing, we have to do this patch..

This patch make `make check` pass on Cygwin 2.4.1 x86_64.

### DIFF
--- a/lib/os.h
+++ b/lib/os.h
@@ -275,7 +275,14 @@
 #endif
 
 #endif
+#ifdef __CYGWIN__
+/*
+ * Cygwin 2.4.1 x86_64 would return ECHILD during nanosleep().
+ * Avoid it during I/O wait.
+ */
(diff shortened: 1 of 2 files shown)


okuoku:
...no. Gambit issues wait(2) inside `sigchld_signal_handler` and it clobber `errno` set by nanosleep(). This fix is not correct. CLOSE.

Signal handlers should save `errno` on any syscalls as POSIX says:
> Note in particular that even the "safe" functions may modify errno; the signal-catching function, if not executing as an independent thread, should save and restore its value in order to avoid the possibility that delivery of a signal in between an error return from a function that sets errno and the subsequent examination of errno could result in the signal-catching function changing the value of errno.

Issue #109, opened 2015-03-18
Bug on OS X, due to linker's incomplete support of thread local storage class


thewhimer:
I tried to use Gambit-C with the 'multiple-vms’ option enabled, to harness full cpu power by SMP. But ‘compile-file’ on REPL failed with the message:

ld: illegal thread local variable reference to regular symbol ____tls_ptr for architecture x86_64
clang: error: linker command failed with exit code 1 (use -v to see invocation)

After some experiments, it seems that the OS X linker does not properly support dynamic linking of externally defined TLS variables. This can be verified easily with the following code.

// test.c
extern __thread int tls_id;
int get_tls_id()
{
    return tls_id;
}

On OS X, both gcc and clang failed to produce a dynamically loadable object file.
(OS X 10.10.1, Xcode 6.2, clang-600.0.57; gcc-4.8 intalled from homebrew.)
$ gcc-4.8 -shared test.c -o libfoo.so
$ gcc-4.9 -shared test.c -o libfoo.so
$ clang -bundle test.c -o libfoo.so

These three give the same error:

Undefined symbols for architecture x86_64:
  "_tls_id", referenced from:
      _get_tls_id in test-9d6352.o
     (maybe you meant: _get_tls_id)
ld: symbol(s) not found for architecture x86_64

While on Linux (ArchLinux 3.18.6 X86_64, gcc 4.9.2), it worked. The produced library linked fine with main.c (as follows), execution gives 1.

// main.c
#include <stdio.h>

__thread int tls_id;

extern int get_tls_id();

int main ()
{
    int id = 0;
    tls_id = 1;
    id = get_tls_id();
    printf(“tls id is %d\n”, id);
    return 0;
}

It might be tested in configure script, and should fallback to pthread_get/setspecific (POSIX) if not supported.

feeley:
That's a bummer!  pthread_getspecific is much less efficient...  I wonder if Apple is working on a fix.

jncornett:
This is actually not limited to OS X. LLVM doesn't support extern __thread variables. You can, however use C++11 thread_local type variables just fine. I filed a bug with LLVM in December, but I don't think anyone has looked at it for whatever reason: https://llvm.org/bugs/show_bug.cgi?id=25821

thewhimer:
This should be easy to fix, a helper function could be used to retrieve the external TLS variable for processor state, with minimal performance cost.

[...]

PR #195, opened 2016-04-02
Fix issue #109, always use getter/setter to access thread local processor state.

thewhimer:
Fix issue #109, use getter/setter to access tls_ptr, thus avoid linking external TLS variable.

To remove the minimal overhead of using a getter when TLS is supported, ___GET_PSTATE() could have an internal definition (i.e. when compiling the gambit library).

### DIFF
--- a/include/gambit.h.in
+++ b/include/gambit.h.in
@@ -5376,22 +5376,16 @@ if (___TYP((___temp=(src)))==___tSUBTYPED&&___SUBTYPE(___temp)==___FIX(___sPROMI
 
 #ifdef ___THREAD_LOCAL_STORAGE_CLASS
 
-/* Use thread local storage class (this is probably fast) */
-
 extern ___THREAD_LOCAL_STORAGE_CLASS void *___tls_ptr;
 
 #define ___FAST_GET_PSTATE
-#define ___GET_PSTATE() ___CAST(___processor_state,___tls_ptr)
-#define ___SET_PSTATE(ps) ___tls_ptr = ___CAST(void*,ps)
 
-#else
+#endif
 
-/* Use thread local storage getter/setter (this is probably slow) */
+/* Always use getter/setter to access thread local storage */
 
 #define ___GET_PSTATE() ___CAST(___processor_state,___get_tls_ptr())
 #define ___SET_PSTATE(ps) ___set_tls_ptr(___CAST(void*,ps))
-
-#endif
 
 #endif
 
@@ -9565,12 +9559,10 @@ typedef struct ___global_state_struct
        ___P((___WORD volatile *ptr),
             ());
 #endif
-#ifndef ___THREAD_LOCAL_STORAGE_CLASS
     void *(*___get_tls_ptr) ___PVOID;
     void (*___set_tls_ptr)
        ___P((void *ptr),
             ());
-#endif
     void (*___disable_heartbeat_interrupts) ___PVOID;
     void (*___enable_heartbeat_interrupts) ___PVOID;
 #endif
@@ -10869,12 +10861,10 @@ ___IMP_FUNC(___WORD,___emulated_fetch_and_clear_word)
    ___P((___WORD volatile *ptr),
         ());
 #endif
-#ifndef ___THREAD_LOCAL_STORAGE_CLASS
 ___IMP_FUNC(void *,___get_tls_ptr) ___PVOID;
 ___IMP_FUNC(void,___set_tls_ptr)
    ___P((void *ptr),
         ());
-#endif
 #endif
 
 #ifndef ___INCLUDED_FROM_OS_TIME
--- a/lib/mem.c
+++ b/lib/mem.c
@@ -1122,6 +1122,7 @@ ___SCMOBJ ___find_global_var_bound_to
         (val)
 ___SCMOBJ val;)
 {
+  ___processor_state ___ps = ___PSTATE;
   ___SCMOBJ sym = ___NUL;
   int i;
 
--- a/lib/os_thread.c
+++ b/lib/os_thread.c
@@ -287,15 +287,19 @@ ___WORD volatile *ptr;)
 
 #ifdef ___THREAD_LOCAL_STORAGE_CLASS
 
-
 ___THREAD_LOCAL_STORAGE_CLASS void *___tls_ptr;
 
-
-#else
+#endif
 
 
 void *___get_tls_ptr ___PVOID
 {
+#ifdef ___THREAD_LOCAL_STORAGE_CLASS
+
+  return ___tls_ptr;
+
+#else
+
 #ifdef ___USE_POSIX_THREAD_SYSTEM
 
   return pthread_getspecific (___thread_mod.tls_ptr_key); /* ignore error */
@@ -315,6 +319,8 @@ void *___get_tls_ptr ___PVOID
 
 #endif
 #endif
+
+#endif
 }
 
 void ___set_tls_ptr
@@ -322,6 +328,12 @@ void ___set_tls_ptr
         (ptr)
 void *ptr;)
 {
+#ifdef ___THREAD_LOCAL_STORAGE_CLASS
+
+  ___tls_ptr = ptr;
+
+#else
+
 #ifdef ___USE_POSIX_THREAD_SYSTEM
 
   pthread_setspecific (___thread_mod.tls_ptr_key, ptr); /* ignore error */
@@ -341,10 +353,9 @@ void *ptr;)
 
 #endif
 #endif
-}
-
 
 #endif
+}
 
 
 /*---------------------------------------------------------------------------*/


thewhimer:
AppVeyor build failed on 1 of 4 configuration, x86_64 msvc, with reason:
```
config.status: creating prebuilt/windows/build-phase2
config.status: creating include/config.h
cd . && autoconf configure.ac > configure && chmod 755 configure
/bin/sh: autoconf: command not found
make: *** [configure] Error 127
Command exited with code 2
```

vyzo:
doesn't this add overhead for platforms that support tls from libraries just fine?
It would be better to have __get_tls_ptr be a macro that gets defned to use an implementation function only on MacOSX.

feeley:
The failure on appveyor is unrelated to this issue... for some reason on the 4th configuration the makefile thinks the "configure" file is stale and tries to recreate it, but autoconf is not installed.  If one of you knows how to avoid this problem, please propose a fix.

Concerning issue #109, I don't like the fix you propose because it always uses the (slow) get_tls_ptr function even on systems where direct access to the tls_ptr is possible.  I have a fix that will adapt the implementation to the particular platform.  Give me a day to implement it.

gambiteer:
> On Apr 2, 2016, at 1:28 PM, Marc Feeley <notifications@github.com> wrote:
> 
> The failure on appveyor is unrelated to this issue... for some reason on the 4th configuration the makefile thinks the "configure" file is stale and tries to recreate it, but autoconf is not installed. If one of you knows how to avoid this problem, please propose a fix.
> 
> 

Install autoconf and list autoconf as a dependence for Gambit.

Brad

okuoku:
> If one of you knows how to avoid this problem, please propose a fix.

Since [Git won't care about files' datetime](http://stackoverflow.com/questions/2458042/restore-files-modification-time-in-git), you have to `touch` *every* generated files after checkout (or install and do autoconf before build as @gambiteer mentioned).

thewhimer:
With my fix, when TLS is supported, __get_tls_ptr becomes ```void * ___get_tls_ptr(void) {return __tls_ptr;}```.
The overhead is only several instructions. Below is ___get_tls_ptr dissembled in such case, with only 5 extra instructions.

(Mac OSX, x86_64, clang-703.0.29)

```
000000010002b3c0 <____get_tls_ptr>:
   10002b3c0:  55                      push   %rbp
   10002b3c1:  48 89 e5                mov    %rsp,%rbp

   10002b3c4:  e8 4f 4a a1 00          callq  100a3fe18 <_____20___gambitgsc+0x378>
   10002b3c9:  48 8d 3d 40 5b e7 00    lea    0xe75b40(%rip),%rdi        # 100ea0f10 <____tls_ptr>
   10002b3d0:  ff 17                   callq  *(%rdi)
   10002b3d2:  48 8b 00                mov    (%rax),%rax

   10002b3d5:  5d                      pop    %rbp
   10002b3d6:  c3                      retq   
   10002b3d7:  66 0f 1f 84 00 00 00    nopw   0x0(%rax,%rax,1)
   10002b3de:  00 00 
```

Note that the two 'callq' instruction is generated by compiler to get the actual TLS variable.
To prove this, here's the orignal version of ____alloc_rc, directly accessing the TLS variable.

```
0000000100015990 <____alloc_rc>:
   100015990:	55                   	push   %rbp
   100015991:	48 89 e5             	mov    %rsp,%rbp
   100015994:	48 83 ec 40          	sub    $0x40,%rsp
   100015998:	48 89 7d c8          	mov    %rdi,-0x38(%rbp)

   10001599c:	e8 77 c4 a2 00       	callq  100a41e18 <_____20___gambitgsc+0x378>
   1000159a1:	48 8b 7d c8          	mov    -0x38(%rbp),%rdi
   1000159a5:	48 89 7d f0          	mov    %rdi,-0x10(%rbp)
   1000159a9:	48 8d 3d 60 d5 e8 00 	lea    0xe8d560(%rip),%rdi        # 100ea2f10 <____tls_ptr>
   1000159b0:	ff 17                	callq  *(%rdi)
   1000159b2:	48 8b 00             	mov    (%rax),%rax

   1000159b5:	48 89 45 e8          	mov    %rax,-0x18(%rbp)
   1000159b9:	48 8b 45 f0          	mov    -0x10(%rbp),%rax
   1000159bd:	48 83 c0 20          	add    $0x20,%rax
   1000159c1:	48 89 c7             	mov    %rax,%rdi
   1000159c4:	e8 77 54 01 00       	callq  10002ae40 <____alloc_mem>
   1000159c9:	48 89 45 e0          	mov    %rax,-0x20(f%rbp)
   1000159cd:	48 83 7d e0 00       	cmpq   $0x0,-0x20(%rbp)
   1000159d2:	0f 84 76 00 00 00    	je     100015a4e <____alloc_rc+0xbe>

... (more, omitted)

```

And as I mentioned, if such minimal overhead is to be avoided in the core library, then the macro could be defined as the old definition when compiling the gambit libray. I didn't add it because I had not found a macro definition separating core library compilation and user land compilation.

But, in my opinion, when outside the core library (compiling user modules), a better API is a better worth than saving several instruction.

thewhimer:
I do aware that the extra call/ret instructions are a little more costly than normal instructions. But they are cheap non the less.

So, with my argument above:
@vyzo  The performance hit is negligible.
@feeley  I think it's a reasonable fix. But it would be great if there's a better fix.

vyzo:
The performance hit is not negligible -- access to tls_ptr is frequent enough to make a difference (eg: memory allocation).

thewhimer:
@vyzo I am aware of that. Let's say ___get_tls_ptr() would take twice the time of directly accessing TLS variable (adding to the two callq instructions), how much performance degradation would that be? It should be negligible according to  Amdahl's Law, even it it called many times.

Doing micro optimizations would be a good exercise. But I'd rather time go to feature implementation and tooling improvement. Correctness, Robustness, Great Features, Performance, and maybe Neatness, come first, then micro optimization.

Anyway, I still thought it as a reasonable fix, if not ideal.

vyzo:
Sorry, I replied on the issue instead of the pull request.
As I commented there:
The problem is that it's in the fastpath; twice as slow on something there makes for measurable performance variations and noone likes performance regressions that can be avoided.
Why don't you simply make it a macro that expands to a function call on systems that need it and a direct access otherwise?

thewhimer:
I got your point. I am probably too permissive about performance degradations, or maybe because it is my fix. :-)

I did not make it a specific case of OS X, because I thought:
1. that would seems like a hack, while a unified one seems good.
2. I prefer a function as the interface, rather than directly accessing an external variable, in the user land.

feeley:
Fixed by commit dbbb8f58dcd89a76e17e83d74672a59fe41ed4af.  The file gambit.h.in contains this explanation:

```
/*
 * Some target platforms have linkers that don't allow referencing a
 * thread local storage variable from another dynamically linked file.
 * This happens on OS X when using clang (but not when using FSF gcc).
 * This issue can be avoided by having the dynamically linked file
 * call the ___get_tls_ptr in the Gambit runtime system which accesses
 * the thread local storage.  However, doing this slows down access to
 * the processor state, which needs to be as fast as possible.
 *
 * The presence of this issue for the C compiler being used could be
 * determined by ifdefs or the configure script.  However, this would
 * prevent dynamically loading code compiled by a different C
 * compiler.  For example, the Gambit runtime system and interpreter
 * could have been compiled by FSF gcc and a file dynamically loaded by
 * the interpreter could have been compiled by clang.  This situation
 * is common when Gambit is installed using a prebuilt installer
 * (typically built with FSF gcc) and the user's platform is
 * configured with a different C compiler (for example Xcode's clang
 * on OS X).
 *
 * The current solution is to assume that thread local storage
 * variables cannot be referenced from another dynamically linked
 * file.  If it is known that the C compilers being used don't have
 * this issue, then the --enable-dynamic-tls option should be passed
 * to the configure script, i.e.:
 *
 *   ./configure --enable-dynamic-tls
 */
```

thewhimer:
@feeley Thanks for the fix. But a minor problem, make check still fails. I had to comment out ```#ifndef ___PRIMAL``` as below.

```
#ifndef ___DYNAMIC_TLS

#undef ___DEFINE_THREAD_LOCAL_STORAGE_GETTER_SETTER
#define ___DEFINE_THREAD_LOCAL_STORAGE_GETTER_SETTER

#ifdef ___SHARED
// #ifndef ___PRIMAL
#undef ___USE_THREAD_LOCAL_STORAGE_GETTER_SETTER
#define ___USE_THREAD_LOCAL_STORAGE_GETTER_SETTER
// #endif
#endif

#endif

#else
```

Because of ```DEFS_DYN=" -D___PROFILE -D___DYNAMIC"``` in gambcomp-C, ```___PRIMAL``` is always defined when compiling dynamically loadable objects.

```
/*
 * Define symbols appropriate for dynamic loading.
 */

#ifdef ___DYNAMIC
#undef ___PRIMAL
#define ___PRIMAL
#undef ___LIBRARY
#define ___LIBRARY
#undef ___SHARED
#define ___SHARED
#undef ___BIND_LATE
#define ___BIND_LATE
#endif
```

And, could you explain a bit about ```___PRIMAL/___LIBRARY/___SHARED/___BIND_LATE```, perhaps add some comment in gambit.h.in?

thewhimer:
My configure options:
``./configure --prefix=/usr/local/opt/gambit-scheme --enable-profile --enable-multiple-vms --enable-multiple-threaded-vms```

feeley:
The problem is elsewhere.  The definition of ```___DYNAMIC``` should not cause ```___PRIMAL``` to be defined.  This has now been fixed by commit 540bdedad9a01ef141e8c5848d4533c06980419a .

PR #206, opened 2016-08-31
Unit tests for the following string procedures:

eval-apply:
string?
string<?
string>?
string=?
string-set!
string-ref
string-length
string-fill!
string-copy
string-ci=?

### DIFF
--- a/tests/unit-tests/09-string/#.scm
+++ b/tests/unit-tests/09-string/#.scm
@@ -1,0 +1,1 @@
+(include "../#.scm")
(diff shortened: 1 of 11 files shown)


feeley:
Thanks for the unit tests.  May I suggest you look at the existing unit tests for vectors in tests/unit-tests/07-vector because they cover all the operations on vectors, and the operations on strings are very similar (just the type of element changes).  Maybe you can adapt one of those files.  The unit tests for string<?, string=?, string-ci=?, etc are needed however because there is no equivalent for vectors.

eval-apply:
Sure no problem I can have a look at adapting the vector tests.

eval-apply:
As an aside, would you prefer them collated in one file like the vector tests as well or one file per procedure?

feeley:
It would be best to mirror the tests for the vector types, just for consistency.  In general it is best to separate the tests out into individual files, so that it is easier to identify the cause of problems.

PR #250, opened 2017-05-03
Adds primitives for univ backend and unit tests

316k:


feeley:
It looks good. The only missing part is the file gsc/tests/03-fxprim/fxwrapasp.scm to test ```##fxwraparithmetic-shift?```. Please add that and then I'll merge the pull request.

PR #252, opened 2017-05-06
Universal backend: adds more primitive definitions and improves some unit tests

316k:


feeley:
The universal backend implements the runtime library at two levels.  The higher-level functionality is implemented in Scheme in lib/_univlib.scm and the low-level functionality is implemented by the target code generator and Gambit linker (which is capable of generating code for functionality that is requested by one or more of the linked modules).  An interesting example is ```println``` which is implemented in two variants.  The full-featured ```println``` is defined in lib/_io.scm (included in lib/_univlib.scm).  A basic ```println``` is also defined in gsc/_t-univ-2.scm to make it more convenient to do simple output (for example in unit tests) without having to link with the universal library.  Here is how the basic ```println``` is defined in gsc/_t-univ-2.scm, as two branches of a case statement:

```
    ((println)
     (rts-method
      'println
      '(public)
      'noresult
      (list (univ-field 'obj 'str))
      "\n"
      '()
      (lambda (ctx)
        (let ((obj (^local-var 'obj)))
          (case (target-name (ctx-target ctx))
            ((js)
             (^if (^prop-index-exists?
                   "function () {return this;}()"
                   (^str "console"))
                  (^expr-statement
                   (^call-prim (^member (^global-var 'console) 'log)
                               obj))
                  (^expr-statement
                   (^call-prim "print"
                               obj))))
            ((python)
             (^expr-statement (^call-prim "print" obj)))
            ((ruby php)
             (^ (^expr-statement (^call-prim "print" obj))
                (^expr-statement (^call-prim "print" "\"\\n\""))))
            ((java)
             (^expr-statement (^call-prim (^member (^member 'System 'out) 'println) obj)))
            (else
             (compiler-internal-error
              "univ-rtlib-feature println, unknown target")))))))

    ((glo-println)
     (univ-defs-combine
      (univ-jumpable-declaration-defs
       ctx
       #t
       (gvm-proc-use ctx "println")
       'entrypt
       '()
       '()
       (univ-emit-fn-body
        ctx
        "\n"
        (lambda (ctx)
          (^ (^expr-statement
              (^call-prim
               (^rts-method-use 'println)
               (^call-prim
                (^rts-method-use 'tostr)
                (^getreg 1))))
             (^setreg 1 (^void-obj))
             (^return
              (^cast*-jumpable (^getreg 0)))))))
      (rts-init
       (lambda (ctx)
         (^setglo 'println
                  (^this-mod-jumpable
                   (gvm-proc-use ctx "println")))))))
```

The case labels ```println``` and ```glo-println``` are identifiers for "runtime system features" that the linker can generate.  The first branch defines a target language printing function that prints a target language string followed by a newline. The second branch defines the Scheme global variable ```println``` and initializes it to a procedure that converts the parameter (Scheme object) into a target language string and calls the target language printing function.  When the Gambit compiler compiles a program containing a reference to the Scheme global variable ```println``` it will automatically create a dependency to the feature ```glo-println``` (this is done in an ad-hoc way for the global variables ```println```, ```real-time-milliseconds``` and ```##exit-process``` by the procedure ```univ-glo-dependency``` defined in gsc/_t-univ-1.scm).  This will cause the linker to generate the code for the ```glo-println``` feature, which in turn creates a dependency to the ```println``` and ```tostr``` features (thanks to the ```(^rts-method-use 'println)``` and```(^rts-method-use 'tostr)```) so the linker will also generate the code for the ```println``` feature.  So a Scheme program not using ```println``` will not include the code for these features.

The same strategy should be used to implement ```##flscalbn```, and specifically a ```ldexp``` feature that is conditionally linked by the Gambit linker.  So in gsc/_t-univ-4.scm there should be:

```
(define (univ-emit-float-scalbn ctx expr1 expr2)
  (case (target-name (ctx-target ctx))

    ((js)
     (^call-prim (^rts-method-use 'ldexp) expr1 expr2))

    ...))
```

and in gsc/_t-univ-2.scm there should be:

```
    ((ldexp)
     (rts-method
      'ldexp
      '(public)
      'f64
      (list (univ-field 'f64 'mantissa)
            (univ-field 'f64 'exponent))
      "\n"
      '()
      (lambda (ctx)
        (case (target-name (ctx-target ctx))
          ((js)
           "
            var steps = Math.min(3, Math.ceil(Math.abs(exponent) / 1023));
            var result = mantissa;
            for (var i = 0; i < steps; i++)
              result *= Math.pow(2, Math.floor((exponent + i) / steps));
            return result;
           ")
          (else
           (compiler-internal-error
            "univ-rtlib-feature ldexp, unknown target"))))))
```

feeley:
Note that it would be preferable to use the style:

```
(^call-prim (^member 'Math 'sinh) expr)
```

instead of

```
(^ "Math.sinh(" expr ")")
```

feeley:
The test ```(< 1518500250 ##max-fixnum)``` should really be ```(##fixnum? 1518500250)```.

feeley:
Hmmm... this is strange:

```
   ((php)
     'cos)
```

Please make sure that the compiler unit tests work on all targets before doing a pull request.

316k:
The parts looking like :

```
    ((php)
     'asin)
```

are part of :

```scheme
  (^call-prim
   (case (target-name (ctx-target ctx))

     ((js ruby java)
      (^member 'Math 'asin))

    ((python)
     (^member 'math 'asin))

    ((php)
     'asin)

    (else
     (compiler-internal-error
      "univ-emit-float-asin, unknown target")))
   expr)
```

Which will be expanded as :
```
(^call-prim
     'asin
   expr)
```

And works properly with the PHP target.

feeley:
OK. I missread the code.

PR #253, opened 2017-05-06
Add support for unit/stress tests of PHP and Ruby

abdelq:
```bash
% make ut-php
% make ut-ruby
% make st-php
% make st-ruby
```
are now valid targets.

The project's `.gitignore` seems to be inconsistent in regards to the gamcomp files, as `grep gambcomp .gitignore` yielded:

```
bin/gambcomp-C
bin/gambcomp-C.unix
bin/gambcomp-C.bat
bin/gambcomp-C.bat.windows
```

So I generalized it to include the other languages.


feeley:
Has this been tested on Unix and Windows?

abdelq:
Tested on:

Arch Linux
```
gcc (GCC) 6.3.1 20170306
ruby 2.4.1p111 (2017-03-22 revision 58053) [x86_64-linux]
PHP 7.1.4 (cli) (built: Apr 25 2017 09:48:36) ( NTS )
```

Windows 10
```
gcc (GCC) 5.4.0
ruby 2.3.3p222 (2016-11-21 revision 56859) [x86_64-cygwin]
PHP 7.0.18 (cli) (built: Apr 20 2017 18:12:23) ( NTS )
```

Using cygwin, it produces the same result as Javascript and Python, which should/will be investigated and fixed in a separate PR.

feeley:
I see that the ut-python, ut-php and ut-ruby targets cause lots of warnings to be output for each unit test.  This is because lib/_univlib.scm only defines ```##os-device-directory-read```, etc for JavaScript with

```
(macro-case-target
 ((js)

(##inline-host-declaration
"
var fs = require('fs');
...
```

so to avoid the warnings an "else" clause for the ```macro-case-target``` should be added that has dummy definitions for the not yet implemented functions.

PR #256, opened 2017-05-09
Universal backend: shrink the stack in the poll() function

316k:


feeley:
Nice use of ```univ-stack-resizable?```.  However you need something in the "else" branch of the if... a ```(^)``` to generate no code when the stack is not resizable.  Has this been tested on all targets?

316k:
The code has been tested with javascript, php, python and ruby.

The case that would have failed here is the java target, which I'm not sure how to test. Is the procedure to build Java executables documented somewhere ?

feeley:
This should get you started:

```
% cat test.scm
(println 42)
% gsc -link -flat -target java -namespace "" test.scm
% javac test_.java test.java
% java test_
42
```

A few things here are not "ideal"... the "-namespace" option set to empty should not be needed.  The namespace defaults to "G_" (prefix on Java classes created by Gambit), but that would require the generated file to be named G_test_ for the "java" program to start the correct main.  Also the trailing underscore is not intuitive.  Perhaps all of this can be masked by creating a .jar file (can the entry point be set explicitly?).  In any case, a -exe option should be added to automate all of this for the Java back-end.

316k:
Sorry, actually this commit generates an error with the python backend I didn't catch.

The stack is defined as an (array scmobj), but it is more specifically an extensible-array-literal, which is an array everywhere except in python (where it is a dict() of {addr: value})

I don't see an obvious way of efficiently shrinking that without going through each key and testing whether it is (< . (+ sp 1)), should the stack shrinking be disabled in python's case ?

feeley:
It can be disabled for now, but some solution has to be found eventually.  This is needed for Java also.

316k:
What's the reason behind this representation ? Wouldn't a simple array work the same as in js/ruby/php ?

feeley:
In Python arrays don't extend automatically, but dicts do.  In Java a fixed size array is used, but a solution for stack overflow has to be found.

316k:
I'm guessing the g_poll function should check if the stack has to be expanded.
If we were to change python's stack representation to an array, Java and python could use the same  algorithm for that.

feeley:
Expanding the stack by copying it is not efficient, and that is the only option with Java and Python. What should be done is to call ```heapify_cont``` to transfer the stack frames to the heap.  This frees up the space in the stack so new frames can be allocated.  That's the general idea but the details have to be worked out (in particular ```heapify_cont``` can only be called with a parameter that is a ReturnPoint so that the frame layout is known).

PR #258, opened 2017-05-10
Adds float eqv? tests and fixes some tests for the universal backend

316k:


316k:
The (eqv? -0.0 0.0) for PHP is hackish, but I don't think there is any other way to test it.

This makes a few more unit tests pass with make ut-php, but it seems there is still a problem with literal -0.0 values in Scheme code.

gambiteer:
Not-a-numbers, as generated by various run-time libraries, are not all equivalent to each other (they can have different signs and mantissas).  I don't know how to deal with checking eqv? of NaNs in all these languages, but I thought I'd point this out.

316k:
Are all +nan.0 eqv? in Scheme ? As long as it is the case, checking for `x !=x and y != y` should return true when comparing two +nan.0

It is the case for Racket http://docs.racket-lang.org/reference/numbers.html, although this doesn't mean it's standard or it is the way Gambit does it.

gambiteer:
On 05/10/2017 03:08 PM, Nicolas Hurtubise wrote:
> Are all +nan.0 eqv? in Scheme ? As long as it is the case, checking for
> |x !=x and y != y| should return true when comparing two +nan.0
>
> It is the case for Racket
> http://docs.racket-lang.org/reference/numbers.html, although this
> doesn't mean it's standard or it is the way Gambit does it.


In the C backend eqv? on flonums is defined by ##fleqv?, which expands to

#ifdef ___BUILTIN_64BIT_INT_TYPE
#define 
___F64EQV(x,y)(___ps->f64_temp1=(x),___ps->f64_temp2=(y),*___CAST(___U64*,&___ps->f64_temp1)==*___CAST(___U64*,&___ps->f64_temp2))
#else
#define 
___F64EQV(x,y)(___ps->f64_temp1=(x),___ps->f64_temp2=(y),___U64_hi32((*___CAST(___U64*,&___ps->f64_temp1)))==___U64_hi32((*___CAST(___U64*,&___ps->f64_temp2)))&&___U64_lo32((*___CAST(___U64*,&___ps->f64_temp1)))==___U64_lo32((*___CAST(___U64*,&___ps->f64_temp2))))
#endif

(from gambit.h), which compares flonums bitwise (and which is another 
example of breaking strict aliasing in C, blah, I guess I didn't note 
that at the time).

So if two NaNs generated by a runtime have different mantissa "payloads" 
(which used to be true in Apple's runtime for various elementary 
functions, and may still be true of some runtimes), then they will not 
be eqv? in the C backend.

I think this behavior is correct.

Brad

316k:
Hmm... I don't think it is possible to do that in the languages targeted by the universal backend.

The current JavaScript implementation uses Object.is(1, 2), so it depends on how the interpreter handles different NaNs. I haven't seen a case where node.js returns false with that case.

In Java, http://docs.oracle.com/javase/6/docs/api/java/lang/Double.html#doubleToRawLongBits(double) could be the thing to use, but : "If the argument is NaN, the result is 0x7ff8000000000000L.", so that wouldn't help.

Maybe there is something to do with PHP's [pack](http://php.net/pack) function or python's struct.pack, but I doubt Ruby has anything that would do that.


This is definitely trickier than I expected. A simple solution would be to ignore NaNs in the eqv? implementations. That would mean (eqv? +nan.0 +nan.0) is an unreliable test and will lead to different results with different backends.

In JavaScript, I believe it would generally return #t, with other languages, testing against == will always return #f.

gambiteer:
On 05/10/2017 03:47 PM, Nicolas Hurtubise wrote:
> Hmm... I don't think it is possible to do that in the languages targeted
> by the universal backend.

That's OK.  Assuming that there is only one NaN is probably a "good 
enough" approximation for the non-C backends.

Brad

316k:
Ok, I'll assume only one NaN in the universal backend, but I'll remove the unit test I added for `(eqv? +nan.0 +nan.0)`.

gambiteer:
On 05/10/2017 03:55 PM, Nicolas Hurtubise wrote:
> Ok, I'll assume only one NaN in the universal backend, but I'll remove
> the unit test I added for |(eqv? +nan.0 +nan.0)|.

The literal "+nan.0" may always return one NaN, I don't know, but please 
leave the test, it's good to check that a NaN is eqv? to itself even 
though it isn't = to itself.

Or maybe it should be changed to something like

(let ((x (fl/ 0. 0.))) (eqv? x x))

Brad

feeley:
For ```univ-emit-float-eqv?``` this is probably shorter/faster:

```
    ((php)
     (^eq? (^call-prim "strval" expr1)
           (^call-prim "strval" expr2)))

    ((python)
     (^if-expr (^= expr1 expr2)
               (^= (^call-prim (^member 'math 'copysign) (^float 1.0) expr1)
                   (^call-prim (^member 'math 'copysign) (^float 1.0) expr2))
               (^and (^!= expr1 expr1)
                     (^!= expr2 expr2))))
```

feeley:
The ruby ```x.angle``` thing for detecting negative zero is hilarious!  I would expect the ```x.equal?(y)``` form to be more efficient:

```
irb(main):003:0> 0.0.equal?(0.0)
=> true
irb(main):004:0> 0.0.equal?(-0.0)
=> false
irb(main):005:0> -0.0.equal?(0.0)
=> false
```

Note that NANs compare *not* equal, so NANs have to be handled in another branch.

316k:
It is more efficient indeed, but you're missing a test case :

```
irb(main):004:0> -0.0.equal?(-0.0)
=> false
```

Not sure why, but this is not a syntax trick :

```
irb(main):006:0> a = -0.0
=> -0.0
irb(main):007:0> b = -0.0
=> -0.0
irb(main):008:0> a.equal?(b)
=> false
```

There is always the option :

```
a.equal?(b) or (-a).equal?(-b)
```

Which is faster when comparing 0.0 with 0.0, but slower than the hilarious method whenever a -0.0 is used. Test results here -> https://gist.github.com/316k/a01b9ba7e352cd00a1beb8985c1fe992

I would go with the assumption that -0.0 is not a common value and use that anyway, it'll be faster in most cases.

gsc/_t-univ-3.scm
```diff
@@ -2104,7 +2104,9 @@
                (^float-floor expr)))
 
     ((ruby)
-     (^ expr ".truncate"))
+     (^if-expr (^ "[0, Float::INFINITY, -Float::INFINITY, Float::NAN].include?(" expr ")")
+               expr
+               (^call-prim (^member expr 'truncate))))
```
feeley: This test is too expensive... better would be:
```
  (^if-expr (^float-finite? expr) (^call-prim (^member expr 'truncate)) expr)
```

But for this to work, the function ```univ-emit-float-finite?``` needs to be augmented with the case:
```
  ((ruby)
   (^call-prim (^member expr 'finite?)))
```

316k: There is still the -0.0 issue.

```
irb(main):001:0> -0.0.truncate
=> 0
```

gsc/_t-univ-3.scm
```diff
@@ -2781,9 +2783,12 @@ tanh
...
-                          ;; 0.0.angle => 0, -0.0.angle => 3.1415...
-                          (^= (^call-prim (^member expr1 'angle)) (^call-prim (^member expr2 'angle)))))
+                    (^parens
+                     (^and (^= expr2 (^float 0.0))
+                           ;; -0.0 vs -0.0 is handled with : a.equal?(b) or (-a).equal?(-b)
+                           (^parens
+                            (^or (^call-prim (^member expr1 'equal?) expr2)
+                                 (^call-prim (^member (^parens (^- expr1)) 'equal?) (^- expr2)))))))
                (^and (^!= expr1 expr1)
                      (^!= expr2 expr2))))
```
feeley: A bit better:
```
     (^if-expr (^= expr1 expr2)
               (^= (^call-prim (^member expr1 'angle))
                   (^call-prim (^member expr2 'angle)))
               (^and (^float-nan? expr1)
                     (^float-nan? expr2)))
```

And this case should be added to ```univ-emit-float-nan?```:
```
    ((ruby)
     (^call-prim (^member expr 'nan?)))
```

gsc/_t-univ-3.scm
feeley: Forgot about -0... so:

```
  (^if-expr (^and (^float-finite? expr) (^!= expr (^float 0.0)))
            (^call-prim (^member expr 'truncate))
            expr)
```

tests/unit-tests/00-check/check-eqvp.scm
```diff
@@ -12,7 +12,7 @@
 
 ;; Equivalent NaNs should be eqv?
 (let ((x (fl/ 0. 0.)))
-  (eqv? x x))
+  (check-eqv? x x))
```
feeley: I don't expect the NaN to be negative. In your example, I'm worried the compiler will do Common Subexpression Elimination (CSE) and convert that back to your original case.  So how about:

```
 (let* ((x (fl/ 0.0 0.0)) (y (fl+ x x))) (check-eqv? x y))
```

316k: Couldn't `(fl+ +nan.0 +nan.0)` result in a different `+nan.0` ?

feeley: In principle yes. But that would be unexpected. But even ```(fl/ 0.0 0.0)``` and a different ```(fl/ 0.0 0.0)``` could give different NaNs... some implementations store in some of the bits of the NaN the address of the faulty operation.  So maybe this is "safe":

```
(define (f x y) (fl/ x y))
(let ((x (f 0.0 0.0)) (y (f 0.0 0.0))) (check-eqv? x y))
```

316k: Ok, this is should be good enough.

feeley:
I have accepted the pull request but will make some minor improvements myself.

PR #259, opened 2017-05-12
Universal backend: fixes various negative zero issues with PHP

316k:
As I suspected, most of the failed unit tests were due to the way PHP handles -0.0 (which had an impact on cpxnum implementations of trigonometric functions, among other things).

With these changes, all number-related unit tests should pass with `make ut-php`.

### DIFF
--- a/gsc/_t-univ-4.scm
+++ b/gsc/_t-univ-4.scm
@@ -903,7 +903,12 @@
 (univ-define-prim "##fl-" #f
   (univ-fold-left
    #f ;; 0 arguments impossible
-   (lambda (ctx arg1)      (^- arg1))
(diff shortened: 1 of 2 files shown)


gambiteer:
Just two comments:

Vel Kahan wrote a famous paper

Branch Cuts for Complex Elementary Functions
or
Much Ado About Nothing's Sign Bit

available here:

https://people.freebsd.org/~das/kahan86branch.pdf

and which was used as a reference for some of the functions in lib/_num.scm.

It seems like these latest conversations could be an appendix to that paper.

Second, I'm glad that there are enough unit tests to instill a bit of 
confidence in the correctness of new back ends.

Brad

PR #260, opened 2017-05-18
Universal backend: more corrections

316k:


gsc/_t-univ-1.scm
```diff
@@ -930,7 +930,13 @@
      (^ expr1 " >>> " expr2))
 
     ((python ruby php)
-     (^ expr1 " >> " expr2 " & (1 << 32 - " expr2 ") - 1"))
+     (^bitand
+      (^>> expr1
+           expr2)
+      (^- (^parens
+           (^<< (^int 1)
+                (^- (^int univ-word-bits) expr2)))
```
abdelq: This might not actually work if `expr2 == 0`. Requires some unit tests.

feeley: It should be ok because the only place ```^>>>``` is used is with expr2 = 14 in ```g_bignum_from_u32```.

Perhaps instead of ```(^>>> m (^int 14))``` the expression ```(^bitand (^>> m (^int 14)) (^int (- (expt 2 (- 32 14)) 1)))``` could be used when the target is not JS or Java to avoid a run time cost.

PR #262, opened 2017-05-19
Universal backend: various improvements  and unit tests

316k:


gambiteer:
I don't really understand the relationship between `tests/unit-tests/...` and `gsc/tests/...`, but it appears that `check-=` in `tests/unit-tests/#.scm` does something similar to what you have with, e.g.,
```
(define (test a b)
  (let ((x (##flacosh a)))
   (println (and (##fl>= x (##fl* 0.999999999999999 b))
                 (##fl<= x (##fl* 1.000000000000001 b))))))
```

316k:
That's a good point. I based my tests on existing `gsc/tests/...` files, I'm not sure if there is reason behind that style.

feeley:
The style of unit tests in gsc/tests was chosen to rely on a very minimal run time support, basically just the existence of a basic ```println```.  This is to ease the incremental development of a backend where very little functionality works at the beginning.  So the unit tests in gsc/tests simply perform some output with ```println``` and then compare if the backend gives the same result as the C backend which is used as a reference.

gsc/tests/24-proc/apply.scm
```diff
@@ -0,0 +1,17 @@
...
+(println (##apply - '(1 2 3)))
+(println (##apply + '(1 2 3 4)))
+(println (##apply * '(1 2 3 4 5)))
+(println (##apply - '(1 2 3 4 5 6)))
+(println (##apply + '(1 2 3 4 5 6 7)))
+(println (##apply * '(1 2 3 4 5 6 7 8)))
+(println (##apply - '(1 2 3 4 5 6 7 8 9)))
+(println (##apply + '(1 2 3 4 5 6 7 8 9 10)))
+
+(println (##apply + '(1 2 3 4 5 6 7 8 9 10 11 12 13 14 15 16 17 18 19 20)))
```
feeley: These tests aren't valid because ```eq?```, ```+```, etc are not predefined procedures at this level of testing.  So instead of ```eq?``` there should be ```(lambda (x y) (##eqv? x y))```. Of course, ```##apply``` is currently unimplemented in the universal backend but it will have to work at some point (note that lib/_univlib.scm defines a hackish ```##apply``` that works for a small number of arguments).

gsc/tests/24-proc/first-argument.scm
```diff
@@ -0,0 +1,7 @@
+(declare (extended-bindings) (not constant-fold) (not safe))
+
+(println (##= (##first-argument 1 1) 1))
+(println (##= (##first-argument -1000 2000 3000) -1000))
+(println (##= (##first-argument -1000.9293221 #t #f 10 20 30 '(1 2 3 4 5) '(2 . 45)) -1000.9293221))
+(println (##eq? (##first-argument ##+ ##- ##* ##/) ##+))
+(println (##= (##car (##first-argument '(1 2 3 4 5))) 1))
```
feeley: These tests aren't valid because ```##=```, ```##+```, etc are not predefined procedures at this level of testing.

gsc/tests/24-proc/apply.scm
feeley: Actually ```##apply``` is implemented by the universal backend!

gambiteer:
> On May 19, 2017, at 3:31 AM, Marc Feeley <notifications@github.com> wrote:
> 
> These tests aren't valid because ##=, ##+, etc are not predefined procedures at this level of testing.
> 
Why?  They’re defined in _num.scm, which means they’re defined in _univlib.scm, no?  Is something more needed?

Brad

feeley:
This is a *compiler* unit test (in gsc/tests) so the unit test is not linked with the universal library. The purpose of the compiler unit tests is to verify that all the inlined primitives (such as ```##fx+```, ```##cons```, ```##car```) work properly.  These primitives are needed to *implement* the universal library.  For example, ```##+``` in the universal library is defined using ```##fixnum?```, ```fx+?```, ```##fl+```, etc.  So when implementing a backend the first thing to get working are the compiler unit tests, and once they pass (or most of them do), then work can proceed in getting the general unit tests working (in tests/unit-tests).

gambiteer:
Yeah, thanks, I figured that out after I hit “send”.

Brad

PR #269, opened 2017-06-06
Dynamically sized fdsets for poll

vyzo:
Implements dynamically sized fdsets for poll, so that we can scale to arbitrary numbers of file descriptors without having to manually increase the compile-time constant `MAX_POLLFDS`.
This solves the issue of crashing on an `___FD_SET` (or worse, silently corrupting) when the file descriptor is higher than `MAX_POLLFDS` by checking and resizing as needed.

Note that it's not intended to side-step the limit on the number of file descriptors that can be concurently polled (that's specified by `MAX_CONDVARS`), but to allow us to have arbitrary numbers of open file descriptors without fear of crashing in the next file we open and try to read.

### DIFF
--- a/include/gambit.h.in
+++ b/include/gambit.h.in
@@ -8381,6 +8381,15 @@ typedef struct ___sync_op_struct
     ___VOLATILE ___WORD arg[2];   /* arguments of the operation */
   } ___sync_op_struct;
 
+/* fdset state */
+#ifdef USE_POLL_FOR_SELECT
+typedef struct ___fdset_state_struct {
+  int size;
+  void *readfds;
+  void *writefds;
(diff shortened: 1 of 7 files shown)


vyzo:
Moved pollset thread state initialization to generic `___setup_thread_local_state` per comments in #270.

vyzo:
ah, i see a build failure without multi-thread vms; will fix.

vyzo:
also, the guard in os_io.c state needs to be `___SINGLE_THREADED_VMS`

vyzo:
Discussion/Review in #270 which supersedes.

vyzo:
Closing as #270 supersedes.

PR #270, opened 2017-06-06
[WIP] Implement raw devices and Dynamic Pollsets

vyzo:
Implements raw devices so that we can treat raw file descriptors (epoll fds, udp sockets, etc) as first class devices with their own io condvars.
This adds a `raw-device-port` port type, the primitives `##open-raw-device` and `##make-raw-device-port` and the kernel primitives required to support them through a `___device_raw` device.
The new port type is poll/selectable (for read or write) and closable, but doesn't implement any other port procedures because they are not relevant.
It also gets its own port kind, `___RAW_DEVICE_KIND`

This PR started on top of #269, so it also adds dynamic pollsets:
Implements dynamically sized fdsets for poll, so that we can scale to arbitrary numbers of file descriptors without having to manually increase the compile-time constant MAX_POLLFDS.
This solves the issue of crashing on an ___FD_SET (or worse, silently corrupting) when the file descriptor is higher than MAX_POLLFDS by checking and resizing as needed.

Note that it's not intended to side-step the limit on the number of file descriptors that can be concurently polled (that's specified by MAX_CONDVARS), but to allow us to have arbitrary numbers of open file descriptors without fear of crashing in the next file we open and try to read.

### DIFF
--- a/include/gambit.h.in
+++ b/include/gambit.h.in
@@ -8381,6 +8381,15 @@ typedef struct ___sync_op_struct
     ___VOLATILE ___WORD arg[2];   /* arguments of the operation */
   } ___sync_op_struct;
 
+/* fdset state */
+#ifdef USE_POLL_FOR_SELECT
+typedef struct ___fdset_state_struct {
+  int size;
+  void *readfds;
+  void *writefds;
+} ___fdset_state;
+#endif
+
 /* Processor structure */
 
 typedef struct ___processor_state_struct
@@ -8462,6 +8471,10 @@ typedef struct ___processor_state_struct
     ___MUTEX_DECL(sync_mut)
     ___CONDVAR_DECL(sync_cv)
 
+#endif
+
+#ifdef USE_POLL_FOR_SELECT
+    ___fdset_state fdset_state;
 #endif
   } ___processor_state_struct;
 
--- a/include/gambit.h.in
+++ b/include/gambit.h.in
@@ -8146,6 +8146,15 @@ typedef struct ___rc_header_struct
     ___SCMOBJ data; /* needed for C closures */
   } ___rc_header;
 
+/* fdset state */
+#ifdef USE_POLL_FOR_SELECT
+typedef struct ___fdset_state_struct {
+  int size;
+  void *readfds;
+  void *writefds;
+} ___fdset_state;
+#endif
+
 typedef struct ___pstate_os_struct {
 
   /*
@@ -8157,6 +8166,11 @@ typedef struct ___pstate_os_struct {
 #else
   ___half_duplex_pipe select_abort; /* POSIX self-pipe */
 #endif
+  
+#ifdef USE_POLL_FOR_SELECT
+    ___fdset_state fdset_state;
+#endif
+
 
 } ___pstate_os;
 
@@ -8381,15 +8395,6 @@ typedef struct ___sync_op_struct
     ___VOLATILE ___WORD arg[2];   /* arguments of the operation */
   } ___sync_op_struct;
 
-/* fdset state */
-#ifdef USE_POLL_FOR_SELECT
-typedef struct ___fdset_state_struct {
-  int size;
-  void *readfds;
-  void *writefds;
-} ___fdset_state;
-#endif
-
 /* Processor structure */
 
 typedef struct ___processor_state_struct
@@ -8471,10 +8476,6 @@ typedef struct ___processor_state_struct
     ___MUTEX_DECL(sync_mut)
     ___CONDVAR_DECL(sync_cv)
 
-#endif
-
-#ifdef USE_POLL_FOR_SELECT
-    ___fdset_state fdset_state;
 #endif
   } ___processor_state_struct;
 
--- a/lib/_io#.scm
+++ b/lib/_io#.scm
@@ -150,6 +150,7 @@
 (##define-macro (macro-tty-kind)         (+ 15 64))
 (##define-macro (macro-serial-kind)      (+ 15 128))
 (##define-macro (macro-tcp-client-kind)  (+ 15 256))
+(##define-macro (macro-raw-device-kind)  (+ 15 512))
 (##define-macro (macro-tcp-server-kind)  (+ 1 512))
 (##define-macro (macro-directory-kind)   (+ 1 1024))
 (##define-macro (macro-event-queue-kind) (+ 1 2048))
@@ -528,6 +529,17 @@
 
 (##define-macro (macro-tcp-client-port? obj)
   `(##port-of-kind? ,obj (macro-tcp-client-kind)))
+
+;;; - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - -
+
+;;; Representation of raw device ports.
+
+(define-check-type raw-device-port 'raw-device-port
+  macro-raw-device-port?)
+
+(##define-macro (macro-raw-device-port? obj)
+  `(##port-of-kind? ,obj (macro-raw-device-kind)))
+
 
 ;;; - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - -
 
--- a/lib/_io#.scm
+++ b/lib/_io#.scm
@@ -545,7 +545,7 @@
   unprintable:
 
   extender: define-type-of-raw-device-port
-  
+
   rdevice-condvar
   wdevice-condvar
   fd
--- a/lib/_io#.scm
+++ b/lib/_io#.scm
@@ -534,6 +534,23 @@
 
 ;;; Representation of raw device ports.
 
+(define-type-of-port raw-device-port
+  id: f55e3678-0414-63d0-3fda-68b9bc518bca
+  type-exhibitor: macro-type-raw-device-port
+  constructor: macro-make-raw-device-port
+  implementer: implement-type-raw-device-port
+  macros:
+  prefix: macro-
+  opaque:
+  unprintable:
+
+  extender: define-type-of-raw-device-port
+  
+  rdevice-condvar
+  wdevice-condvar
+  fd
+)
+
 (define-check-type raw-device-port 'raw-device-port
   macro-raw-device-port?)
 
--- a/lib/_io.scm
+++ b/lib/_io.scm
@@ -8254,6 +8254,33 @@
         (macro-dynamic-bind output-port port thunk)))))
 
 ;;;----------------------------------------------------------------------------
+(define-prim (##open-raw-device
+              direction
+              name
+              fd
+              settings)
+  (define (fail)
+    (##fail-check-settings 4 ##open-raw-device direction name fd settings))
+  
+  (##make-psettings
+   direction
+   '(direction:)
+   settings
+   fail
+   (lambda (psettings)
+     (let ((device
+            (##os-device-open-raw
+             fd
+             (##psettings->device-flags psettings))))
+       (if (##fixnum? device)
+         (##exit-with-err-code device)
+         (and device
+              (##make-device-port-from-single-device
+               name
+               device
+               psettings)))))))
+
+;;;----------------------------------------------------------------------------
 
 (define-prim (##open-predefined
               direction
--- a/lib/_io.scm
+++ b/lib/_io.scm
@@ -8257,15 +8257,14 @@
 (define-prim (##open-raw-device
               direction
               name
-              fd
-              settings)
+              fd)
   (define (fail)
-    (##fail-check-settings 4 ##open-raw-device direction name fd settings))
-  
+    (##fail-check-settings 1 ##open-raw-device direction name fd))
+
   (##make-psettings
    direction
-   '(direction:)
-   settings
+   '()
+   '()
    fail
    (lambda (psettings)
      (let ((device
@@ -8274,11 +8273,111 @@
              (##psettings->device-flags psettings))))
        (if (##fixnum? device)
          (##exit-with-err-code device)
-         (and device
-              (##make-device-port-from-single-device
-               name
-               device
-               psettings)))))))
+         (##make-raw-device-port device fd name direction))))))
+
+(define-prim (##make-raw-device-port device fd uname direction)
+  (let ((mutex
+         (macro-make-port-mutex))
+        (rkind
+         (if (or (##eq? direction (macro-direction-in))
+                 (##eq? direction (macro-direction-inout)))
+           (macro-raw-device-kind)
+           (macro-none-kind)))
+        (wkind
+         (if (or (##eq? direction (macro-direction-out))
+                 (##eq? direction (macro-direction-inout)))
+           (macro-raw-device-kind)
+           (macro-none-kind)))
+        (roptions
+         0)
+        (rtimeout
+         #t)
+        (rtimeout-thunk
+         #f)
+        (woptions
+         0)
+        (wtimeout
+         #t)
+        (wtimeout-thunk
+         #f)
+        (rdevice-condvar
+         (and (or (##eq? direction (macro-direction-in))
+                  (##eq? direction (macro-direction-inout)))
+              (##make-rdevice-condvar device)))
+        (wdevice-condvar
+         (and (or (##eq? direction (macro-direction-out))
+                  (##eq? direction (macro-direction-inout)))
+              (##make-wdevice-condvar device)))
+        (uname (or uname 'raw-device)))
+
+    (define (name port)
+
+      ;; It is assumed that the thread **does not** have exclusive
+      ;; access to the port.
+
+      (##declare (not interrupts-enabled))
+      (##list uname (macro-raw-device-port-fd port)))
+
+    (define read-datum #f)
+
+    (define write-datum #f)
+
+    (define newline #f)
+
+    (define force-output #f)
+
+    (define set-rtimeout #f)
+
+    (define set-wtimeout #f)
+
+    (define (close port prim arg1)
+
+      ;; It is assumed that the thread **does not** have exclusive
+      ;; access to the port.
+
+      (##declare (not interrupts-enabled))
+
+      (macro-port-mutex-lock! port) ;; get exclusive access to port
+
+      (let ((result
+             (##close-device
+              port
+              (macro-raw-device-port-rdevice-condvar port)
+              (macro-raw-device-port-wdevice-condvar port)
+              prim)))
+        (macro-port-mutex-unlock! port)
+        (if (##fixnum? result)
+            (##raise-os-io-exception port #f result prim arg1)
+            result)))
+
+    (let ((port
+           (macro-make-raw-device-port
+            mutex
+            rkind
+            wkind
+            name
+            read-datum
+            write-datum
+            newline
+            force-output
+            close
+            roptions
+            rtimeout
+            rtimeout-thunk
+            set-rtimeout
+            woptions
+            wtimeout
+            wtimeout-thunk
+            set-wtimeout
+            #f ;; io-exception-handler
+            rdevice-condvar
+            wdevice-condvar
+            fd)))
+      (if rdevice-condvar
+        (##io-condvar-port-set! rdevice-condvar port))
+      (if wdevice-condvar
+        (##io-condvar-port-set! wdevice-condvar port))
+      port)))
 
 ;;;----------------------------------------------------------------------------
 
--- a/lib/_kernel.scm
+++ b/lib/_kernel.scm
@@ -4279,6 +4279,12 @@ end-of-code
             scheme-object   ;; device
    "___os_device_stream_open_process"))
 
+(define-prim ##os-device-open-raw
+  (c-lambda (scheme-object  ;; fd
+        scheme-object) ;; flags
+       scheme-object   ;; device
+   "___os_device_raw_open"))
+
 (define-prim ##os-device-process-pid
   (c-lambda (scheme-object) ;; dev
             scheme-object   ;; pid (fixnum)
--- a/lib/os_io.c
+++ b/lib/os_io.c
@@ -109,6 +109,8 @@ static ___poll_fdset ___fdset_writefds (___processor_state ps)
   return ___fdset_state_writefds (ps);
 }
 
+#endif
+
 int ___fdset_resize_pstate
    ___P((___processor_state ___ps,
          int fd),
@@ -117,10 +119,13 @@ int ___fdset_resize_pstate
 ___processor_state ___ps;
 int fd;)
 {
+#ifdef USE_poll
   return ___fdset_realloc (___ps, fd);
+#endf
+  return 0;
 }
 
-#endif
+
 
 /*---------------------------------------------------------------------------*/
 
--- a/lib/os_io.c
+++ b/lib/os_io.c
@@ -84,27 +84,27 @@ static void ___fdset_realloc (int fd)
   int newsize = oldsize;
   while (newsize <= fd)
     {
-      newsize = newisze * 2;
+      newsize = newsize * 2;
     }
 
   ___fdset_state_readfds = realloc (___fdset_state_readfds, newsize/8);
   ___fdset_state_writefds = realloc (___fdset_state_writefds, newsize/8);
   ___fdset_state_size = newsize;
-  memset(___fdset_state_readfds + oldsize/8, 0, (newsize - oldsize)/8)
+  memset(___fdset_state_readfds + oldsize/8, 0, (newsize - oldsize)/8);
   memset(___fdset_state_writefds + oldsize/8, 0, (newsize - oldsize)/8);
 }
 
-static void ___fdset_size ()
+static int ___fdset_size ()
 {
   return ___fdset_state_size;
 }
 
-static void ___fdset_readfds ()
+static ___poll_fdset ___fdset_readfds ()
 {
   return ___fdset_state_readfds;
 }
 
-static void ___fdset_writefds ()
+static ___poll_fdset ___fdset_writefds ()
 {
   return ___fdset_state_writefds;
 }
@@ -774,9 +774,9 @@ int fd;
 ___BOOL for_writing;)
 {
   if (for_writing)
-    ___FD_SET(fd, &state->writefds);
+    ___FD_SET(fd, state->writefds);
   else
-    ___FD_SET(fd, &state->readfds);
+    ___FD_SET(fd, state->readfds);
 
   if (fd >= state->highest_fd_plus_1)
     state->highest_fd_plus_1 = fd+1;
@@ -1243,7 +1243,7 @@ ___time timeout;)
 
 #ifdef USE_ASYNC_DEVICE_SELECT_ABORT
 
-  if (___FD_ISSET(___PSTATE->os.select_abort.reading_fd, &state.readfds))
+  if (___FD_ISSET(___PSTATE->os.select_abort.reading_fd, state.readfds))
     {
       /* self-pipe has available data to read, discard all of it */
 
@@ -1422,7 +1422,7 @@ ___time timeout;)
 
 #ifdef USE_select_or_poll
 
-  if (___FD_ISSET(___PSTATE->os.select_abort.reading_fd, &state.readfds))
+  if (___FD_ISSET(___PSTATE->os.select_abort.reading_fd, state.readfds))
     {
       /* self-pipe has available data to read, discard all of it */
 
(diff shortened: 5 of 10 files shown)


vyzo:
this still needs debugging, i am seeing some crashes using the condvars; marking as [WIP]

vyzo:
The issue was incorrect implementation of select logic (the pass) in ___device_raw_select_virt.
Fixed in https://github.com/gambit/gambit/pull/270/commits/2d2c5193910208926d33dde0bb5512cd2a5aa9a2.

feeley:
I'll be reviewing the patch in more detail soon. A quick glance raises a few flags about the code's assumptions. The management of the "pollset state" is unclear.  At a minimum the initialization is at the wrong place (start_pthread_thread is part of the logic of the OS thread creation abstraction).  The best place is setup_io_pstate (which initializes any I/O related state in the "processor"). More comments to come.

feeley:
I'd like to backtrack to the initial motivation for this patch. What is the problem this is trying to solve? I worry that this extension is POSIX specific. Exposing OS specific features goes against Gambit's design philosophy.  So at a minimum it should be made clear by the naming scheme that this is only for POSIX (e.g. ```##open-posix-device```). But I'd like to be convinced that there isn't a higher level abstraction that could cover multiple OSes (at least POSIX and Windows). So it is important to define the core problem that is being addressed. My intuition is that you want an "opaque" port on which read/write operations are done at a lower level (C), but the "waiting for I/O" is done by the Gambit thread scheduler. If that's the case then there is clearly a generalization possible to other OSes.

vyzo:
> At a minimum the initialization is at the wrong place (start_pthread_thread is part of the logic of the OS thread creation abstraction). The best place is setup_io_pstate (which initializes any I/O related state in the "processor"). 

That probably belongs to #269, as it's the initialization of the pollsets, unless you want to review them together. I put it there because I didn't know where to put it -- i'll be happy to move it.

vyzo:
> I'd like to backtrack to the initial motivation for this patch. What is the problem this is trying to solve? I worry that this extension is POSIX specific. Exposing OS specific features goes against Gambit's design philosophy. So at a minimum it should be made clear by the naming scheme that this is only for POSIX (e.g. ##open-posix-device). But I'd like to be convinced that there isn't a higher level abstraction that could cover multiple OSes (at least POSIX and Windows). So it is important to define the core problem that is being addressed. My intuition is that you want an "opaque" port on which read/write operations are done at a lower level (C), but the "waiting for I/O" is done by the Gambit thread scheduler. If that's the case then there is clearly a generalization possible to other OSes.

Your intuition is very correct: I want an opaque port with userland read/write ops and I/O waiting integrated with the Gambit scheduler. That's the only way we can implement things like epoll and udp sockets (and other exotic file descriptors).

I agree that there is a case for generalization, and that's why I called it a raw device, but I don't know enough about windows to do this generalization! So the idea was "do it for posix and wait for someone with windows expertise who needs it to do the generalization"

feeley:
I tend to use another approach for portability: actually implementing a feature on at least 2 OSes before it is released tends to bring out issues which confirm or not that the abstraction is well thought out.  In this particular case I think that I/O abstraction should be easily supported by Windows and POSIX, so that doesn't worry me.  However, it is probably necessary to change the port inheritance hierarchy and add another level above "object port".  In other words, the highest class is a kind of port with no read/write ops defined.

vyzo:
Agreed, the opaque ports have no meaningful read/write ops except `close`.
Do you want me to change the hierarchy or do you prefer to work on top of the patch and extend it yourself?

PS: moving pollset initialization, update coming soon -- need to test it first.

vyzo:
Also, re: pollset initialization -- i think we can move `___fdset_alloc` in the initializer as well, will do.

vyzo:
Testing for multiple VMs with initializer only in `setup_io_pstate ` crashes the primordial thread -- looks like the initializer needs to be called in the primordial thread as well.

vyzo:
So the initializatoin can't be outside the thread because it uses ___thread state, it needs to be in the actual thread and can't be moved to `___setup_io_pstate`. 
I think we want a general mechanism to initialze state for thread local variables, something less ugly than calling `___fdset_state_init`.

vyzo:
Moved the initialization to `___setup_thread_local_state`, which is called from *within* the thread to initialize thread local variables.
So no more ugly direct pollset initialization and a generic procedure to initialize more TLS variables in the future.

feeley:
Still don't understand why this can't be done in ```setup_io_pstate``` which is the logical place for this. Please explain.

I will look into extending the class hierarchy to add a "generic port" layer above "object port".

vyzo:
`___device_select` has no access to the processor state.
Perhaps you would prefer to change that function interface and have all state in processor state instead of tls variables?

vyzo:
Note that this cascades up to `___os_condvar_select`, which is its caller and doesn't have access to processor state either -- and it's not clear where to get it either.

vyzo:
The definition of `##os-condvar-select!` will have to change to pass pstate:
```
(define-prim (##os-condvar-select! devices timeout)
  (##declare (not interrupts-enabled))
  (##c-code
   "___RESULT = ___os_condvar_select (___ARG1, ___ARG2);"
   devices
   timeout))
```

is the pstate in `___PSTATE` here?

feeley:
Yes the correct approach is to pass the processor state. But it is tricky to do that and I can look into it. For now, you can use the ```___PSTATE``` macro in ```___os_condvar_select```.

vyzo:
Oh, it's accessible in `___os_condvar_select`? 
That's handy, i can change the pollset state management to use pstate then.

vyzo:
Note that this will allow us to move `___fdset_alloc` to the initializer as well, so that it's out of line in `___device_select`.

feeley:
Yes ```___PSTATE``` will get the processor state structure of the current OS thread. It uses (in the multithreaded VM case) the OS thread local storage if it is available. So for portability, it is best to use the processor state to store "thread local" stuff and use ```___PSTATE```.

vyzo:
OK, will do.

feeley:
But for performance, it is best to write ```___condvar_select``` to receive the processor state in a parameter if ```___PSTATE``` is slow (some systems have a slow mechanism to access thread local storage).  In ```##os-condvar-select!``` the processor state is available in ```___ps```.

vyzo:
there is a complication: `___device_select_add_fd` also needs the state to realloc if the bounds check fails.
But that only happens very sparingly, so it's ok to just get it with `___PSTATE` there.

feeley:
It is ok for now but I will have to think if ```___device_select_add_fd``` needs an extra processor state parameter too.

I notice a few problems with the memory management of the fd sets.  First of all there is no heap overflow handling. The complexity of the error handling was the reason to use a fixed size (```___MAX_CONDVARS```) array. I think the fd sets should start at the size ```___MAX_CONDVARS``` and never shrink. Also ```malloc``` and ```realloc``` shouldn't be used directly.  Instead gambit.h has the ```___ALLOC_MEM(bytes)``` macro and a ```___free_mem``` function.

vyzo:
They do starts at the size `MAX_POLLFDS =___MAX_CONDVARS` in `___fdset_alloc`, but I think we can remove `MAX_POLLFDS` altogether and make it clear it starts at `MAX_CONDVARS`.
This will now happen at pstate initialization, so it won't be inline.

Thanks for pointing out about the memory management functions, will change to use them.

vyzo:
Also, they never shrink already! They only grow when they see a file descriptor out of bounds.

vyzo:
there is also `___device_cleanup` that uses `___device_select` and needs access to `___PSTATE`.

feeley:
I know... I'm just pointing out that by design they shouldn't shrink because this could cause issues. I think the placement of the resizing of the fd set should be reconsidered. Currently it is lazy resizing when adding a fd to a fd set. Perhaps it would be better to resize when the fd is allocated (after an "open").  This would be a better place to detect/report memory overflow.

vyzo:
I am making it explicit that they don't shrink any more, `___fdset_realloc` will refuse to sink now, not just the usage that guarantees it.

Dunno about doing it in open, as this doesn't consider external file descriptors from raw devices.
So you would want it in the `___setup_from_fd` functions, which is too much change for not adding much over lazy resizing.

vyzo:
Almost done with the changes, doing some make checks in the various configurations to make sure that nothing broke and will also rerun my many file descriptors test.

vyzo:
Moved all state to `___processor_state` in https://github.com/gambit/gambit/pull/270/commits/f57a7d756a7baebcb7ac0dc36029c9df8c43d2de and got rid of the thread local storage etc.

Note that I am getting segfaults in the compiler tests in make check, not clear whats causing it.

vyzo:
So i see your point of wanting to detect memory allocation failure out of line, with resizing in the open.
Do you want me to move the resizing in the `__setup_from_fd` functions? This will take into account all kinds of devices, which is better than doing it in `open` (we have sockets, predefined, and raw devices that don't open at all)

vyzo:
So for the crash now: it happens when running the compiled mix test with this:
```
./mix -:m4000 > test6.out
/bin/sh: line 1: 15467 Segmentation fault      ./mix -:m4000 > test6.out
makefile:140: recipe for target 'test6' failed
```
gdb says the following:
```
Program received signal SIGSEGV, Segmentation fault.
0x0000000000514f00 in ___trampoline (___ps=0xc9f980 <___gstate0>) at setup.c:2682
2682	      CALL_STEP;
(gdb) bt
#0  0x0000000000514f00 in ___trampoline (___ps=0xc9f980 <___gstate0>) at setup.c:2682
#1  0x0000000000515299 in ___call (nargs=0, proc=12061601, stack_marker=13298129)
    at setup.c:2803
#2  0x0000000000515427 in ___run (thunk=12061601) at setup.c:2837
#3  0x00000000005171a5 in ___setup (setup_params=0x7fffffffdc80) at setup.c:4478
#4  0x000000000053fc82 in ___main (linker=0x4a1336 <____20_mix__>) at main.c:756
#5  0x0000000000529f4d in ___main_char (argc=2, argv=0x7fffffffde78, 
    linker=0x4a1336 <____20_mix__>, script_line=0x0) at os_base.c:449
#6  0x00000000004a1370 in main (argc=2, argv=0x7fffffffde78) at mix_.c:10324
(gdb) p ___ps
$1 = (___processor_state) 0xc9f980 <___gstate0>
(gdb) p *___ps
$2 = {stack_trip = 0x7ffff7f55030, stack_limit = 0x7ffff7ee5040, fp = 0x7ffff7f54370, 
  stack_start = 0x7ffff7f55030, stack_break = 0x7ffff7f545b0, heap_limit = 0x7ffff72ac020, 
  hp = 0x7ffff729c8a8, r = {11026177, -18, -2, -10, 13286217}, pc = 0, temp1 = 11026273, 
  temp2 = 8, temp3 = 0, temp4 = 0, f64_temp1 = 0, f64_temp2 = 0, na = 2, intr_enabled = 252, 
  intr_mask = 0, intr_flag = {0, 0, 16, 0, 0, 0}, processor_scmobj = {43046, 13022241, 0, 
    13236833, 13236833, 13236833, 13236833, 140737339889137, 140737339889137, -2, 0, 13236833, 
    13236833, 140737339899609, 140737339899609, 140737339889137, -2, 140737339888913, 13236833, 
    13236833, 0, -10}, os = {select_abort = {reading_fd = 0, writing_fd = 0}}, mem = {
    tospace_offset_ = 0, stack_msection_ = 0x7ffff7ed5018, alloc_stack_start_ = 0x7ffff7f55030, 
    alloc_stack_ptr_ = 0x7ffff7f55010, alloc_stack_limit_ = 0x7ffff7ed5030, 
    heap_msection_ = 0x7ffff723c018, alloc_heap_start_ = 0x7ffff723c030, 
    alloc_heap_ptr_ = 0x7ffff723c030, alloc_heap_limit_ = 0x7ffff72bc030, 
    alloc_heap_chunk_start_ = 0x0, alloc_heap_chunk_limit_ = 0x0, heap_chunks_to_scan_ = 0, 
    heap_chunks_to_scan_head_ = 0x0, heap_chunks_to_scan_tail_ = 0x0, scan_ptr_ = 0x0, 
    still_objs_to_scan_ = 0, still_objs_ = 13315816, words_still_objs_ = 0, 
    words_still_objs_deferred_ = 200, bytes_allocated_minus_occupied_ = 0, rc_head_ = {
      prev = 0xc9fbb8 <___gstate0+568>, next = 0xc9fbb8 <___gstate0+568>, refcount = 1, 
      data = -2}, traverse_weak_refs_ = 0, nonexecutable_wills_ = 0, executable_wills_ = 1, 
    reached_gc_hash_tables_ = 0, words_prev_msections_ = 0, stack_fudge_used_ = 0, 
    heap_fudge_used_ = 0}, catcher = 0x7fffffffd940, fdset_state = {size = 8192, 
    readfds = 0xcae180, writefds = 0xcae590}}
```

not clear what is causing to SEGFAULT right now.

vyzo:
it also seems that it's better to have the state in `___pstate_os` instead of directly in `__processor_state`, will fix that as well. I noticed this structure in gdb :)

vyzo:
i will start working directly in this PR instead of merging the dynamic set I think, I don't want to make more of a mess with the merge commits.

vyzo:
So I don't think we can do resizing in `___setup_from_fd` -- that will only resize for the processor on which the device is allocated!

vyzo:
oh i know what's causing the segfault -- we know need to memset to zero in init!

vyzo:
no wait, that's totally not it, they are FD_ZEROed anyway.
will delete both this comment and the above.

vyzo:
Investigating the crash further, when I configure without poll it fails to compile entirely:
```
../gsc/gsc -:~~bin=../bin,~~lib=../lib,~~include=../include -f -warnings -c mix.scm
*** ERROR IN c#univ-prim-proc-add! -- Wrong number of arguments passed to procedure
(#<return #2>
 '("f64vector-ref" (2) #f 0 0 real gambit)
 'f64vector-ref
 '#<procedure #3>
 18
 #f
 772
 958
 '#((proc-obj) "f64vector-ref" #f #t #f (2) #<procedure #4> #f #<procedure #...
 'real
 'gambit)
makefile:134: recipe for target 'test5' failed
```
Any ideas what is happening here?

vyzo:
So now with poll I get a new error in `make check` gsc tests:
```
------------ TEST 6 (link and execute the code generated)
rm -f mix_.c mix.o mix_.o mix
../gsc/gsc -:~~bin=../bin,~~lib=../lib,~~include=../include -f -warnings -o mix -exe mix.c
./mix -:m4000 > test6.out
*** ERROR IN beta-subst -- Wrong number of arguments passed to procedure
(##kernel-handlers '('1 '1))
```

Am I seeing some weird caching effects with the compiler and generated code?

feeley:
The resizing can be done on all processors by calling ```on_all_processors``` which does a barrier sync operation over all processors.  You also need to add the operation in the switch statement of ```execute_sync_op```.

feeley:
The errors you are getting are very strange... make bootclean to recompile all.  Quite possibly a problem in different ```gambit.h``` files being used for different modules (I assume you have been changing it).

vyzo:
yeah, have been messing with gambit.h.in, will try bootclean.

feeley:
Just a general comment here... the ```select``` metaphor may not be the best for cases where there are many fds. Each time ```___condvar_select``` is called by a processor the set of file descriptors is reconstructed from the list of I/O condvars of the processor. A system which is based on notifications should be considerably more efficient (libev?). So perhaps your effort to simply remove the hard limit on the number of file descriptors will give disappointing performance until some other I/O waiting interface is devised.

vyzo:
Basically that's the intent behind these patches, to let you do highly scalable networked systems without having to rewrite the entire i/o sybsystem.

vyzo:
Btw my weird `make check` errors continue. I did a `bootclean` and even reverted the generated *.c files from .scm, and the weird error is still there:
```
------------ TEST 6 (link and execute the code generated)
rm -f mix_.c mix.o mix_.o mix
../gsc/gsc -:~~bin=../bin,~~lib=../lib,~~include=../include -f -warnings -o mix -exe mix.c
./mix -:m4000 > test6.out
*** ERROR IN beta-subst -- Wrong number of arguments passed to procedure
(##kernel-handlers '('1))
```

There is a difference however, this time the wrong number of parameters is 1 instead of 2 that was previously.
Definitely something weird happening with my build state, I can't imagine the compiler hardcoding `___processor_state`.
Btw, can you do a build with `--enable-poll` and verify that it `make check`s correctly on your system? -- just to make sure I am not chasing ghosts.

feeley:
"without having to rewrite the entire I/O subsystem"... do you mean the developer or the Gambit implementors? Ideally the Gambit runtime system should support large number of file descriptors efficiently without requiring userland heavy lifting.

vyzo:
The Gambit implementors -- the heavy lifting in userland can go to libraries, and it's written in Scheme and will be more understandable. I like the i/o subsystem as it is right now, it's very understandable and clean.

So don't worry about userland heavy lifting, that's what the Gerbil stdlib is for :)

vyzo:
Note that pushing it the userland allows us to specialize for the network protocol and not for general purpose computing; polling is fine for the general case.

vyzo:
My check problems continue: I did a hard reset and now I am back to SEGFAULTING:
```
------------ TEST 6 (link and execute the code generated)
rm -f mix_.c mix.o mix_.o mix
../gsc/gsc -:~~bin=../bin,~~lib=../lib,~~include=../include -f -warnings -o mix -exe mix.c
./mix -:m4000 > test6.out
/bin/sh: line 1:  9800 Segmentation fault      ./mix -:m4000 > test6.out
```

Can you try to reproduce this? 
configure is ` --enable-single-host --enable-poll --enable-debug`

feeley:
Where should I pull your latest code from?

vyzo:
https://github.com/vyzo/gambit/tree/raw-devices

vyzo:
the crash is still there with a fresh clone btw, and it's driving me crazy!

feeley:
Fresh clone of your repo or Gambit? A fresh clone of Gambit with those configure options works fine for me.

vyzo:
Fresh clone of my repo, with 
```
$ git clone git@github.com:vyzo/gambit.git gambit-vyzo
$ cd gambit-vyzo
$ git checkout raw-devices
```

vyzo:
I did a make gsc-boot on master before checking out though, maybe that has an affect who knows.

feeley:
No that is OK... gsc-boot isn't changing that much to have an impact.

vyzo:
So I am adding a new sync op called `___fdset_resize` which does a parallel resize, but first checks
to see if the fdset is big enough. This should make it fast enough to call on every `___setup_from_fd` function for every file descriptor it sees (and remove the lazy resizing and perhaps costly bounds check).

I have also changed `__fdset_realloc` to do overflow check and return 1 on error.

I have the couple of questions though:
-  are these barriers free to interrupt ___device_select? 
   If that's the case we would have a catastrophe with dangling pointers to the freed old fdsets. 
   It doesn't seem to be the case though.
- what happens if the ops have different results?
   For example, we could have an allocation failing in some processor -- it is perogative for correctness
   In this case we would want to signal a heap overflow error to the constructor of the device, so it's important to discriminate errors.

vyzo:
so there are quite a few places that need processor state to do the eager resizing, and I don't think it's prudent to lug around the processor state everywhere (there are changes in the scheme ikernel interface as well which I don't like).
I got started this way and it's a never ending hole, I'll fetch processor state from `___PSTATE` for now.

feeley:
Yes the processor barrier sync (```on_all_processors```) can interrupt ```device_select``` because this is required to implement sleeping processors and parallel GC.  But this is OK because the interrupt will cause the ```device_select``` to backout into Scheme code, and later the ```device_select``` will be called again after the interrupt is handled.  On POSIX, a self pipe is used for this purpose.

The resizing operation must not fail by raising an exception.  An error flag should be set when a heap overflow happens, and the caller of ```on_all_processors``` should check this flag.  The "resizing" should keep the same array if the allocation fails so that it isn't a fatal error, allowing the system to keep functioning.

vyzo:
So added the eager fdset resizing in https://github.com/gambit/gambit/pull/270/commits/37690c284f3b806943f3c236fba8ad1da01d990a

I think this should be ready, modulo my continuing `make check` woes.

vyzo:
Ah, thanks for the detail on the overflow check, i'll do the error flag.

vyzo:
Just to be clear, it doesn't raise an exception -- it returns 0 for success in realloc and 1 for failure, which the caller of `__fdset_resize` checks to see if there was an allocation failure.
But the implementation is faulty, because there is no value aggregation, so I'll do the overflow flag.

vyzo:
also notice that the overflow handler needs to use `___free_mem` instead of just `free`.

feeley:
I can't build your repo... I get an error about ```___PAIRMUTABLEP``` being undefined.  That appears to be a problem introduced in your repo by a wrong commit. I suggest you do a clone of the Gambit repo, then apply your modifications.  It will be a cleaner pull request in any case.

feeley:
The overflow error flag is not ideal.  I would prefer a combining of the results, similar to what is done for the parameters.  I'll add this to my TODO.

vyzo:
Thanks for reproducing my repo build problems!
I'll do some git surgery and do a force push, no need to open a new PR I think.

vyzo:
Added the heap overflow check with flag in https://github.com/gambit/gambit/pull/270/commits/b04f63ee2dcaa47aa32880dd9a328398f8061d1e

I will look into fixing my build issues next, but I think it's good for proper review.

vyzo:
Ok, I did a hard reset and repull on my master and I can compile there now.
I am making a new branch with all the commits from here, will open a new PR.

vyzo:
hrm, my new branch still barfs in the check and when I get back to master it ceases to compile again.
There might be an issue with some gitignored file that doesn't get cleaned?

I will dissect the commits to try find where the problem is introduced, this is very annoying.

vyzo:
So the crash is introduced somewhere after the first merge, and quite possibly where the new config.h.in is introduced.

vyzo:
Marc, there is a problem with the new gambit.h.in -- the make check failure appears the moment I changed processor state.
In my repaired repo, `make check` succeeds up to https://github.com/gambit/gambit/pull/270/commits/f31c1a5c62392ae906b41f41cc2ba070d774f98b.
The moment I introduce the new pstate, the crashes in `make check` test 6 start: https://github.com/gambit/gambit/pull/270/commits/f57a7d756a7baebcb7ac0dc36029c9df8c43d2de

Are there any hardcoded assumptions about `___processor_state` in the compiler, like its size?
I can't explain this in any other way.

vyzo:
Actually, there is another dissection point I skipped - it seems to be between introducing the new state in pstate and putting it in `___processor_state.os`.

I will try a build there and I have a hunch it might succeed.I suspect the compiler makes some assumptions about the `____processor_state` layout, and gaining new members in the middle of the struct breaks generated code.

vyzo:
the dissection point is https://github.com/gambit/gambit/pull/270/commits/7950361196a459e7fa3732695dddb07b02c8cf2c

vyzo:
nope, flat out fails there too, but it's with a SEGFAULT instead of the weird exception.
So yeah, after the `___processor_state` gains new members, the generated code breaks and `make check` fails at test 6.
Smoking gun for compiler breakage due to `___processor_state` layout and size assumptions.

vyzo:
I just confirmed that without -`-enable-poll` make check succeeds.

I have pushed a new branch here: https://github.com/vyzo/gambit/tree/os-io-work
This is cleaned up and will be the basis of a new clean PR once we have debugged this; let's keep this open in the meantime for discussion and debugging.

vyzo:
Actually I will up the PR soon, as there are some details I want to improve -- for instance `___setup_io_pstate` can signal failure from fdset allocation in `___fdset_init`.

vyzo:
Opened #271 for the continuation for this work. 
Marc, please use #271 for review comments, but let's keep this PR for debugging conversation.

vyzo:
So back to debugging.
Looking at `___pstate_mem`, if there is a layout bug, the first 3 fields would fall into the fdset state:
```
/* location of tospace in each msection */
___SIZE_TS tospace_offset_;

/* msection where continuation frames are currently being allocated */
___msection *stack_msection_;

/* start of allocation of continuation frames in stack_msection */
___WORD *alloc_stack_start_;
```

The last two especially look very relevant for what I see in the debugger and also the weird `##kernel-handlers` error.

vyzo:
That's the error in the head of `os-io-work` branch:
```
------------ TEST 6 (link and execute the code generated)
rm -f mix_.c mix.o mix_.o mix
../gsc/gsc -:~~bin=../bin,~~lib=../lib,~~include=../include -f -warnings -o mix -exe mix.c
./mix -:m4000 > test6.out
*** ERROR IN beta-subst -- Wrong number of arguments passed to procedure
```

vyzo:
Another possibility for the bug, outside the compiler and `kernel.scm` is that the `___processor_struct` is used or allocated somewhere unguarded from `USE_POLL_FROM_SELECT` which is what triggers the structure layout.

vyzo:
```
;; The procedure "##kernel-handlers" is only a container for the kernel
;; handlers.  It must never be called.  The function "___setup" in the
;; file "setup.c" is responsible for setting up the kernel handlers.
```

So how can this end up being called, with arguments no less?
A continuation must have been mucked somewhere.

feeley:
I think your comment about ```USE_POLL_FOR_SELECT``` is correct.  It can't be used in gambit.h.in because ```USE_POLL_FOR_SELECT``` is defined in include/config.h which is only used to compile the Gambit library.  It would need to be moved to gambit.h.in (similarly to ```#define @CONF_HAVE_CONDITION_VARIABLE@```). But another symbol than ```USE_POLL_FOR_SELECT``` should be used, because the feature here is "unlimited number of file descriptors" and it is not related to using poll vs. select as the I/O waiting system call.

vyzo:
What symbol should I use? And will it percolate all the way up to configure?
Note that the feature is currently only implemented for poll, but I can see it being extended to support windows as well (or other select mechanisms).

vyzo:
Looks like this can be done in the ifdef trickery at the top of gambit.h.in, so we shouldn't end up percolating to configure.

feeley:
Come to think of it, this feature is related to the implementation of the I/O so it is not a good idea to affect the layout of the processor state structure. In the declaration for ```___pstate_os``` I see that there is a field for ```select_abort``` which is only used when ```USE_ASYNC_DEVICE_SELECT_ABORT``` is defined (needed for multiple threaded VMs to wake processors that are sleeping when doing an ```on_all_processors```). So the same strategy should apply to the fd sets, possibly a void* pointer to the fd set (unless it is "easy" to define the fd set structure in gambit.h.in without having to include other header files).

vyzo:
The structure is an int and two pointers:
```
typedef struct ___fdset_state_struct {
  int size;
  void *readfds;
  void *writefds;
} ___fdset_state;
```

feeley:
```USE_UNLIMITED_SELECT``` perhaps...

feeley:
Yes that structure seems simple enough.  Why not simply ```___fdset_struct```?

Also it has to be defined in the else branch of an ```#ifdef ___OS_WIN32```.

vyzo:
Yes, I am putting there -- I think we can put it there unguarded and as a pointer, which is unused when `USE_POLL_FOR_SELECT` is undefined in os_io.

vyzo:
Actually, no need for the pointer -- unless you want to keep the state smaller.

vyzo:
And yes! The `make check` finally passes!
The saga comes near the end :)

vyzo:
```============ ALL TESTS SUCCESSFUL```

feeley:
I agree there's no need for the pointer.

By the way, on Windows, the ```MsgWaitForMultipleObjects``` function that is used as a ```select``` mechanism, is limited to ```MAXIMUM_WAIT_OBJECTS``` which is 64... not good. Apparently "I/O completion ports" can be used as a workaround.

vyzo:
Ouch! Hadn't realized there such a grave limitation!

The fix is in:
https://github.com/gambit/gambit/pull/271/commits/9f17cf5b5c934cf166e9e767a68ee60260ef3031

vyzo:
Doing some quick checks with `--enable-multiple-thread-vms` and then without `--enable-poll` to make sure everything works in all relevant configurations, but I think we are finally good for final review in #271.

vyzo:
Fixed brainfart in https://github.com/gambit/gambit/pull/271/commits/d0d9183ba65381d6628d3cb71bc306ccaea47b1b and generalized the raw device structs in https://github.com/gambit/gambit/pull/271/commits/393154c23f7c70f6834d5f9aeb8c82ac849f7322

vyzo:
I think we are done with debugging in this PR, so closing for review in #271.

PR #271, opened 2017-06-07
OS I/O: Raw Devices and Dynamically sized fdsets

vyzo:
Supersedes #269 and #270.

This improves the file descriptor scalability of the Gambit I/O subsystem in two ways:
- It makes fdsets used by poll dynamically sized so that there is no need for recompilation to exceed compile-time max file descriptor limites.
- It adds raw devices, currently implemented for POSIX only, so that special devices can be used to perform userland read/write operations while integrating with the Gambit scheduler.

The intent is to allow us to use raw file descriptors for things like `epoll`, `kqueue`, and `sockets` for highly scalable network i/o in userland. This makes Gambit better suited for the development of network daemons and servers, where scaling to large number of file descriptors and removing indirection from the i/o path is critical for performance.

### DIFF
--- a/include/gambit.h.in
+++ b/include/gambit.h.in
@@ -8381,6 +8381,15 @@ typedef struct ___sync_op_struct
     ___VOLATILE ___WORD arg[2];   /* arguments of the operation */
   } ___sync_op_struct;
 
+/* fdset state */
+#ifdef USE_POLL_FOR_SELECT
+typedef struct ___fdset_state_struct {
+  int size;
+  void *readfds;
+  void *writefds;
+} ___fdset_state;
+#endif
+
 /* Processor structure */
 
 typedef struct ___processor_state_struct
@@ -8462,6 +8471,10 @@ typedef struct ___processor_state_struct
     ___MUTEX_DECL(sync_mut)
     ___CONDVAR_DECL(sync_cv)
 
+#endif
+
+#ifdef USE_POLL_FOR_SELECT
+    ___fdset_state fdset_state;
 #endif
   } ___processor_state_struct;
 
--- a/include/gambit.h.in
+++ b/include/gambit.h.in
@@ -1344,7 +1344,6 @@
 #endif
 #endif
 
-
 /*---------------------------------------------------------------------------*/
 
 /* Dependencies on language (i.e. C++, ANSI-C, and K&R C). */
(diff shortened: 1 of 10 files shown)


vyzo:
I can also move towards the generalization of raw ports while at it.
The simple thing to do is to define the virt struct outside the `USE_POSIX` guard so that it is ready for implementation on windows.

vyzo:
So I also found a brainfart, was guarding with `___USE_poll` inside `___fdset_resize`, fixing.

vyzo:
there is an unecesary `___PSGET` in the implementation of `____fdset_resize`, fixing.

feeley:
I'd like this PR to be split in 2 separate features... unlimited number of descriptors and raw devices. I'll have to change the port hierarchy as previously explained before accepting a PR for the raw devices.

vyzo:
Ok, I'll fight with git to do the split.
I think we want the raw devices PR first because the fdset PR will change the definitions of the `setup_from_fd` functions.

vyzo:
So basically two PRs, one on top of the other:
- raw devices which will do the raw device definition
- dynamic fdsets which will sit on top of the raw devices PR and add they dynamic fdsets

vyzo:
Hrm, it's not a simple matter of making new branches and merging commits selectively, I'll have to cherry pick by hand in the two branches :(

vyzo:
ah, cherry pick seems painless enough.

vyzo:
Raw devices in #271.

vyzo:
Need to fix a couple things in #271, the new FD_SET signature, this will take a little while.

vyzo:
So #272 builds fine on travis on linux, we are good with it.
I'll make the other branch and prepare the fdsets PR on top.

vyzo:
spoke too soon - it seems that I need to guard the raw devices virt for posix better.

vyzo:
ok, linux build green in #272, will prepare the PR as well.

Marc, let me know in #272 when you are done with the port hierarchy so that I can update for merging.

vyzo:
Opened #273 for dynamically sized fdsets.

Marc, we'll need to merge that after #272 because it also updates the raw device `setup_from_fd` to resize fdsets.

vyzo:
PR has been into #272 and #273; Closing.

PR #272, opened 2017-06-08
OS I/O: Raw Devices

vyzo:
Cherry picked from #271.

Adds raw devices, currently implemented for POSIX only, so that special devices can be used to perform userland read/write operations while integrating with the Gambit scheduler.

### DIFF
--- a/lib/_io#.scm
+++ b/lib/_io#.scm
@@ -150,6 +150,7 @@
 (##define-macro (macro-tty-kind)         (+ 15 64))
 (##define-macro (macro-serial-kind)      (+ 15 128))
 (##define-macro (macro-tcp-client-kind)  (+ 15 256))
+(##define-macro (macro-raw-device-kind)  (+ 15 512))
 (##define-macro (macro-tcp-server-kind)  (+ 1 512))
 (##define-macro (macro-directory-kind)   (+ 1 1024))
 (##define-macro (macro-event-queue-kind) (+ 1 2048))
@@ -528,6 +529,17 @@
 
 (##define-macro (macro-tcp-client-port? obj)
   `(##port-of-kind? ,obj (macro-tcp-client-kind)))
+
+;;; - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - -
+
+;;; Representation of raw device ports.
+
+(define-check-type raw-device-port 'raw-device-port
+  macro-raw-device-port?)
+
+(##define-macro (macro-raw-device-port? obj)
+  `(##port-of-kind? ,obj (macro-raw-device-kind)))
+
 
 ;;; - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - -
 
--- a/lib/_io#.scm
+++ b/lib/_io#.scm
@@ -545,7 +545,7 @@
   unprintable:
 
   extender: define-type-of-raw-device-port
-  
+
   rdevice-condvar
   wdevice-condvar
   fd
--- a/lib/_io#.scm
+++ b/lib/_io#.scm
@@ -534,6 +534,23 @@
 
 ;;; Representation of raw device ports.
 
+(define-type-of-port raw-device-port
+  id: f55e3678-0414-63d0-3fda-68b9bc518bca
+  type-exhibitor: macro-type-raw-device-port
+  constructor: macro-make-raw-device-port
+  implementer: implement-type-raw-device-port
+  macros:
+  prefix: macro-
+  opaque:
+  unprintable:
+
+  extender: define-type-of-raw-device-port
+  
+  rdevice-condvar
+  wdevice-condvar
+  fd
+)
+
 (define-check-type raw-device-port 'raw-device-port
   macro-raw-device-port?)
 
--- a/lib/_io#.scm
+++ b/lib/_io#.scm
@@ -591,6 +591,7 @@
   rdevice-condvar
   wdevice-condvar
   device
+  id
 )
 
 (##define-macro (macro-raw-device-port? obj)
--- a/lib/_io#.scm
+++ b/lib/_io#.scm
@@ -166,7 +166,7 @@
 (##define-macro (macro-vector-kind)      (+ 3 16384))
 (##define-macro (macro-string-kind)      (+ 7 32768))
 (##define-macro (macro-u8vector-kind)    (+ 15 65536))
-(##define-macro (macro-raw-device-kind)   (+ 1 262144))
+(##define-macro (macro-raw-device-kind)  (+ 1 262144))
 
 ;;; - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - -
 
@@ -590,11 +590,8 @@
 
   rdevice-condvar
   wdevice-condvar
-  fd
+  device
 )
-
-(define-check-type raw-device-port 'raw-device-port
-  macro-raw-device-port?)
 
 (##define-macro (macro-raw-device-port? obj)
   `(##port-of-kind? ,obj (macro-raw-device-kind)))
--- a/lib/_io.scm
+++ b/lib/_io.scm
@@ -8413,8 +8413,15 @@
       ;; access to the port.
 
       (##declare (not interrupts-enabled))
+      (if (##fx= direction (macro-direction-in))
 
-      (##wait-device port direction))
+        (##wait-for-io!
+         (macro-raw-device-port-rdevice-condvar port)
+         (macro-port-rtimeout port))
+
+        (##wait-for-io!
+         (macro-raw-device-port-wdevice-condvar port)
+         (macro-port-wtimeout port))))
 
     (define (close port prim arg1)
 
--- a/lib/_io.scm
+++ b/lib/_io.scm
@@ -8344,15 +8344,15 @@
 
 ;;;----------------------------------------------------------------------------
 
-(define-prim (##make-raw-device-port raw-device device uname direction)
+(define-prim (##make-raw-device-port raw-device device id direction)
   (let ((mutex
          (macro-make-port-mutex))
         (rkind
-         (if (##eq? direction (macro-direction-out))
+         (if (##fx= direction (macro-direction-out))
            (macro-none-kind)
            (macro-raw-device-kind)))
         (wkind
-         (if (##eq? direction (macro-direction-in))
+         (if (##fx= direction (macro-direction-in))
            (macro-none-kind)
            (macro-raw-device-kind)))
         (roptions
@@ -8368,12 +8368,11 @@
         (wtimeout-thunk
          #f)
         (rdevice-condvar
-         (and (##not (##eq? direction (macro-direction-out)))
+         (and (##not (##fx= direction (macro-direction-out)))
               (##make-rdevice-condvar raw-device)))
         (wdevice-condvar
-         (and (##not (##eq? direction (macro-direction-in)))
-              (##make-wdevice-condvar raw-device)))
-        (uname (or uname 'raw-device)))
+         (and (##not (##fx= direction (macro-direction-in)))
+              (##make-wdevice-condvar raw-device))))
 
     (define (name port)
 
@@ -8381,7 +8380,7 @@
       ;; access to the port.
 
       (##declare (not interrupts-enabled))
-      (##list uname (macro-raw-device-port-device port)))
+      (##list (macro-raw-device-id port) (macro-raw-device-port-device port)))
 
     (define (wait port direction)
 
@@ -8472,7 +8471,8 @@
             #f ;; io-exception-handler
             rdevice-condvar
             wdevice-condvar
-            device)))
+            device
+            id)))
       (if rdevice-condvar
         (##io-condvar-port-set! rdevice-condvar port))
       (if wdevice-condvar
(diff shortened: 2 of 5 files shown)


vyzo:
@feeley any progress with the generic port hierarchy?
I can do them if you don't have the time!

vyzo:
I discovered a bug in the `___raw_device_select_virt` -- it fails to handle for_writing correctly; fixing.

feeley:
Halfway done...

feeley:
Waitable ports are now implemented by commit 05cec0edbd542fe4da76a78cced2969612199216.

vyzo:
Excellent! I will merge and update the raw-device-ports to be descendants of waitable-ports.

By the way, how do you generate the type-ids for the structs? Is it a random uuid or something structured?
I used a hash to generate it for raw-device-ports.

vyzo:
btw, `___PROCESS_DEVICE_KIND` disagrees with `(macro-process-kind)`.
I assume the two should be the same?
```
os_io.c:
#define ___PROCESS_DEVICE_KIND    (___PIPE_DEVICE_KIND+131072)

_io#.scm:
(##define-macro (macro-process-kind)     (+ 31 64))
```

vyzo:
unless the process is just a pipe in userland, in which case it's ok:
```
os_io.c:
#define ___PIPE_DEVICE_KIND       (___DEVICE_KIND+64)
```

vyzo:
Ok, merged! I'll do some casual testing and update #273 as well.

vyzo:
seems like my implementation of wait is not correct - `##wait-device` expects a device-port, so i'll have to call `##wait-for-io!` directly.

vyzo:
ok, verified that all port operations work as intended, this should be ready for merge!

vyzo:
Updates for review comments in #273.

The only sticking point left is the utility of `##open-raw-device` and its name.
For now it takes a device (not an fd :), opens a raw device and calls `##make-raw-device-port`.

We can drop the function altogether if you think the caller should be doing the actual opening of the device (ie if you think it's too POSIX specific)

vyzo:
Thinking more about the issue, is your concern how to obtain the actual device object that gets passed to `##open-raw-device` or the fact that the current implementation of `##os-device-open-raw` is only implemented for posix?
If the concern is the latter, then we should drop the function and keep just `##make-raw-device-port`. It's nicely convenient, but nothing that can't be done in a library.

vyzo:
On further thought, I will simply remove the function altogether.

vyzo:
@feeley any thoughts? I have addressed your review comments, both here and in #273.

I'd like to get this merged soon, as I already have libraries ready in a big Gerbil stdlib branch.
They provide the full POSIX socket API (nonblocking), epoll, and also high level network programming 
facilities that automate i/o multiplexing through epoll (and soon kqueue).

lib/os_io.c
```diff
@@ -8901,6 +8901,258 @@ ___SCMOBJ options;)
...
+
+  *dev = d;
+
+  ___device_add_to_group (dgroup, &d->base);
+
+  return ___FIX(___NO_ERR);
+}
+#endif
+
+___SCMOBJ ___os_device_raw_open
```
vyzo: `open_predefined` tries to check the file descriptor type with fstat, which fails to take into account non-file or socket file descriptors. It also results in a hard crash with epoll file descriptors.

lib/_kernel.scm
```diff
@@ -4279,6 +4279,12 @@ end-of-code
             scheme-object   ;; device
    "___os_device_stream_open_process"))
 
+(define-prim ##os-device-open-raw
```
vyzo: see comment above.

lib/os_io.c
```diff
@@ -8901,6 +8901,258 @@ ___SCMOBJ options;)
...
+    }
+  else if (is_not_closed & direction & ___DIRECTION_RD)
+    d->base.read_stage = ___STAGE_CLOSED;
+  else if (is_not_closed & direction & ___DIRECTION_WR)
+    d->base.write_stage = ___STAGE_CLOSED;
+
+  return ___FIX(___NO_ERR);
+}
+
+___HIDDEN ___SCMOBJ ___device_raw_select_virt
```
vyzo: You can't read from many of these devices -- eg `epoll` and `kqueue` file descriptors.

lib/os_io.c
feeley: I see, so it makes sense to have a function that does no device kind detection.  As far as I can see, this would mean that the methods should have the same implementation as for the "file device" kind.  So either, copy/paste or reuse the file device code (probably you need to copy/paste because the structure offsets are different).

lib/_io.scm
```diff
@@ -8344,6 +8344,143 @@
...
+              (##make-wdevice-condvar raw-device)))
+        (uname (or uname 'raw-device)))
+
+    (define (name port)
+
+      ;; It is assumed that the thread **does not** have exclusive
+      ;; access to the port.
+
+      (##declare (not interrupts-enabled))
+      (##list uname (macro-raw-device-port-device port)))
```
vyzo: I use it as a means for typing raw devices, and also having a nice display of the port.
I could call it name and make it non-optional, so that we are consistent and don't default in raw-device.
Finally, I can see the point about non-closures, I can simply add the name (better, type) to the `raw-device-port` type.

lib/_io#.scm
```diff
@@ -573,6 +574,31 @@
...
+  macros:
+  prefix: macro-
+  opaque:
+  unprintable:
+
+  extender: define-type-of-raw-device-port
+
+  rdevice-condvar
+  wdevice-condvar
+  device
```
vyzo: yes, will do.

lib/os_io.c
vyzo: Many methods are very similiar, in fact I based my implementation on the file device implementation.

lib/os_io.c
feeley: But since for some devices you can, it should be supported. If reading is not supported, then the "direction" should not contain the "in" direction, so all should be fine.

vyzo: To clarify: the direction is still in, as they are waitable for input and have an open direction, but they can't be predefined ports because you can't do actual reads.

lib/_io#.scm
vyzo: called it type, as it is more representative of its usage

lib/_io.scm
vyzo: Added a type argument, and made the name return the type consed with the device for nice debugging output. 
No need to get the type from the name any more though, it's available directly from the port.

lib/_io#.scm
feeley: What do you think of "id", which identifies the port?

vyzo: Sure, even better; name also implies strongly that it's a symbol

vyzo:
Updated per review comments.

vyzo:
smal brainfart in usage of id field, fixing.

lib/os_io.c
feeley: Please make a copy of ```___device_file_select_raw_virt``` including the WIN32 part, because then it will probably work as-is on Windows.

lib/os_io.c
```diff
@@ -8901,6 +8901,263 @@ ___SCMOBJ options;)
...
+  int fl;
+  int direction;
+
+  device_translate_flags (___INT(flags),
+                          &fl,
+                          &direction);
+
+  fd = ___INT(index);
+
+  if ((e = ___device_raw_setup_from_fd
```
feeley: Indent properly... open paren for parameters should be 2 spaces indented from the ```___device_...```.

lib/os_io.c
```diff
@@ -8901,6 +8901,263 @@ ___SCMOBJ options;)
...
+       (&dev,
+        ___global_device_group (),
+        fd,
+        direction))
+      != ___FIX(___NO_ERR))
+    return e;
+
+  if ((e = ___NONNULLPOINTER_to_SCMOBJ
```
feeley: Same indent style should be applied here.

lib/os_io.c
feeley: I keep thinking that it is wrong to expose this function in Scheme since is is POSIX specific. It is OK however to make it available to C code, so that a device can be constructed from a fd using the FFI. The "index" parameter should be "fd" because that's what it is.  The function should be called ```___os_device_raw_open_from_fd```.

lib/_kernel.scm
feeley: See comment below.  This functionality should not be exposed to the Scheme level.

lib/os_io.c
vyzo: I will rename.

The exposure is very minimal, there is just a very slim interface through `##os-device-open-raw`.
It's convenient to have it available to Scheme code, so that we can write simpler ffi code that constructs the fds and then calls the scheme function outside the C code. That way we can compose ffi constructor functions with a generic functions that does the `##open-raw-device`.

lib/_kernel.scm
vyzo: Do you prefer the ffi code to have to define the primitive on its own?
Ie instead of having the ffi interface in _kernel.scm to have it in user code?
I don't have any problem with removing the function from Scheme code in this case, but see comment above per convenience.

lib/os_io.c
vyzo: ok, will do.

lib/_kernel.scm
vyzo: I will remove the function if you feel strongly about it, and move it to the library.

feeley: I guess if it is purely a ```##``` function it can stay with a ```-from-fd``` suffix.

lib/os_io.c
vyzo: do you have an emacs mode for your indentation style?

feeley: Nope.

feeley:
Seems like ```___device_raw_select_virt``` should be called ```___device_raw_select_raw_virt``` to be consistent with the other device kinds.

vyzo:
ok, will rename.

vyzo:
ok, updated! Sorry for the indentation mismatch, i tend to just follow emacs in GNU style.

vyzo:
Seems like the git CR interface is hiding some of the commits in the update, but they are there in the commits page: https://github.com/gambit/gambit/pull/272/commits

vyzo:
actually there is an accident from the copy paste and it's a wrong cast in `___device_raw_select_raw_virt` -- fixing.

vyzo:
fixed in situ in the last commit and did a force push.

vyzo:
ok, now we should be good -- waiting for the build to finish to run some tests, but it looks good to me.

vyzo:
testing successful.

adam-7:
@vyzo , what's a raw device, what are their usecases? Can you provide a short example script that illustrates their utility here?

vyzo:
Things like sockets, epoll, inotify and kquque file descriptors.
With raw devices we can open these OS-specific devices and integrate them with the Gambit scheduler.

adam-7:
Ah!

1)
By "raw device", do you just mean that you have created a facility whereby the user can allocate his own OS file descriptor of any type that the OS can produce,

(I have not studied epoll/inotify/kqueue so it's not in my awareness that they produce their own FD:s, but I understand that's what you're saying, so they fit the picture too;)

and, what you have done here is to provide a facility where the user can create a Gambit port object that functions only to represent such a file descriptor, so that the user may do |##device-port-wait-for-input!| and |##device-port-wait-for-output!| on that port object?

2)
I guess such a raw port object is actually totally passive, as in, Gambit will not do any operations on it such as OS write, read, close operations, but they function only to make Gambit include that FD in Gambit's select/poll() call, during the ##device-port-wait-for-input/output! call?

3)
Are any other operations than ##device-port-wait-for-input/output! supported on those raw ports?

4)
Are the raw ports documented in the manual now or do you keep these out of the manual due to being too low level?

5)
Probably the raw ports will be used for pretty low level stuff, right?

6)
Except for this and making Gambit's epoll() support unlimited number of ports (whereas previously there was an ~8000 FD:s limit and exceeding that led to SIGSEGV), have you done any other major addition now?

Looks great! :) Thanks

feeley:
This is just a heads up that the API for ```##make-raw-device-port``` will be changing to accommodate other types of raw-devices.  In particular what used to be called the "id" is now called the "type", and the "name" of the port is an explicit parameter (so it is easy to choose a custom name). The last field is a "specific" field that can be used to store anything.

Vyzo, here is how the old API can be emulated:

```
(define (##vyzos-make-raw-device-port raw-device device id direction)
  (##make-raw-device-port direction raw-device id (##list id device) device))
```

To access the "device" from the port, use ```(macro-raw-device-port-specific port)```.

adam-7:
Like what types?

vyzo:
thanks marc!

feeley:
I'm implementing UDP ports and it is convenient to build them on top of raw-device ports.  By the way, the API has changed again to: ```(##make-raw-device-port direction device kind id specific)```.  The third parameter is a numeric "kind", for example ```(macro-raw-device-kind)```.  This simplifies type testing.

vyzo:
Can the kind be an arbitrary object (eg a symbol) and not just a numeric kind in the new API?
That would be much better for identifying user ports.

PR #279, opened 2017-06-28
Avoid using git clone/checkout for bootstrap

fare:
Use git worktree for extracting sources at given versions.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -12,7 +12,7 @@ compiler:
 git:
   depth: 9999999
 
-script: git fetch --tags && ./configure && make -j4 current-gsc-boot && make clean && ./configure --enable-debug --enable-multiple-threaded-vms && make -j4 from-scratch && make check && make clean && ./configure --enable-debug --enable-multiple-threaded-vms --enable-cplusplus && make -j4 && make check && make clean && ./configure --enable-ansi-c && make -j4 && (cd tests; make test1) && (cd tests; make test2) && (cd tests; make test3) && (cd tests; make test4) && (cd tests; make test5)
+script: git fetch --tags && ./configure --enable-debug --enable-multiple-threaded-vms && make clean && make -j4 current-gsc-boot && make -j4 from-scratch && make check && make clean && ./configure --enable-debug --enable-multiple-threaded-vms --enable-cplusplus && make -j4 && make check && make clean && ./configure --enable-ansi-c && make -j4 && (cd tests; make test1) && (cd tests; make test2) && (cd tests; make test3) && (cd tests; make test4) && (cd tests; make test5)
 
 notifications:
   webhooks:
--- a/makefile.in
+++ b/makefile.in
@@ -120,56 +120,42 @@ latest-release: fake_target
 	$(GIT) reset --hard
 
 boot-most-recent-release: fake_target
-	@if test -x "boot/most-recent-release/gsc-boot"; then \
-	  if test "`boot/most-recent-release/gsc-boot -v | sed -e 's/ .*//'`" = "$(PACKAGE_VERSION)"; then \
-	    echo "**** boot/most-recent-release/gsc-boot doesn't need to be rebuilt"; \
-	  else \
-	    $(MAKE) boot-most-recent-release-gsc-boot; \
-	  fi \
+	@if test -x "boot/gsc-boot" && \
+	    test "`boot/gsc-boot -v | sed -e 's/ .*//'`" = "$(PACKAGE_VERSION)"; then \
+	  echo "**** boot/gsc-boot doesn't need to be rebuilt"; \
 	else \
 	  $(MAKE) boot-most-recent-release-gsc-boot; \
 	fi
 
 boot-most-recent-release-gsc-boot: fake_target
-	@echo "**** building boot/most-recent-release/gsc-boot"
+	@echo "**** building boot/gsc-boot"
 	rm -rf boot
-	mkdir boot
-	$(GIT) clone --branch master . boot/most-recent-release
-	cd boot/most-recent-release && \
-	$(GIT) checkout -f tags/$(PACKAGE_VERSION) && \
-	$(CONFIGURE_COMMAND) && \
-	$(MAKE) bootstrap && \
-	cd ../..
+	mkdir -p boot
+	$(GIT) archive tags/$(PACKAGE_VERSION) | (cd boot && tar xf -)
+	cd boot && \
+	./configure CC='$(C_COMPILER)' && \
+	$(MAKE) bootstrap
 
 boot-wip-compiler: fake_target
-	@if test -x "boot/wip-compiler/gsc/gsc"; then \
-	  if test "`boot/wip-compiler/gsc/gsc -v | sed -e 's/ .*//'`" = "$(PACKAGE_VERSION)"; then \
-	    if test "`find gsc/makefile.in gsc/*.scm -exec diff -q \{\} boot/wip-compiler/\{\} \; 2>&1`" = ""; then \
-	      echo "**** boot/wip-compiler/gsc/gsc doesn't need to be rebuilt"; \
-	    else \
-	      $(MAKE) boot-wip-compiler-gsc-gsc; \
-	    fi \
-	  else \
-	    $(MAKE) boot-wip-compiler-gsc-gsc; \
-	  fi \
+	@if test -x "boot/gsc/gsc" && \
+	    test "`boot/gsc/gsc -v | sed -e 's/ .*//'`" = "$(PACKAGE_VERSION)" && \
+	    test "`find gsc/makefile.in gsc/*.scm -exec diff -q \{\} boot/\{\} \; 2>&1`" = ""; then \
+	  echo "**** boot/gsc/gsc doesn't need to be rebuilt"; \
 	else \
 	  $(MAKE) boot-wip-compiler-gsc-gsc; \
 	fi
 
 boot-wip-compiler-gsc-gsc: boot-most-recent-release
-	@echo "**** building boot/wip-compiler/gsc/gsc using boot/most-recent-release/gsc-boot"
-	rm -rf boot/wip-compiler
-	$(GIT) clone --branch master . boot/wip-compiler
-	cd boot/wip-compiler && \
-	$(GIT) checkout -f tags/$(PACKAGE_VERSION) && \
-	cp ../../gsc/makefile.in ../../gsc/*.scm gsc && \
-	cp ../most-recent-release/gsc-boot . && \
-	$(CONFIGURE_COMMAND) && \
-	$(MAKE) && \
-	cd ../..
+	@echo "**** building boot/gsc/gsc using boot/gsc-boot"
+	cd boot && \
+	rm -rf gsc && \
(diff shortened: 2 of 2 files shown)


fare:
Questions: what minimal subset of the sources can we copy in release/ and bootstrap/ and still have it all build properly?

feeley:
That's nice.  A few points...

1) I don't like the ```$(wildcard ...)``` that may not be supported by some make.  Perhaps this logic should go in gsc/makefile because that makefile knows about the gsc files.

2) the directory should not be called "bootstrap" because it may conflict with a "make bootstrap" (I think "boot" is adequate)

Concerning your "minimal source" question... hard to say precicely, but roughly

- configure (and probably configure.ac)
- include/gambit.h.in and include/config.h.in and include/stamp.h
- makefile.in in ., include, lib, gsi, gsc, bin
- all .c and .h in lib, gsi and gsc
- the .scm in lib, gsi and gsc are not needed, for the first build, but useful to bootstrap

One of Gambit's design features is that the .c (and .h) files plus a C compiler are sufficient to build gsi and gsc.  In other words, the "delivery" language is C, and this makes the system very portable.  It would be nice if there was a script to create a .tar file containing the .c and .h and configure and makefiles required to build a given application, so that it can be delivered in source form and be portable to any C development environment.  Haven't had the time to work on that...

fare:
I pushed a revised version.
Note that configure hardcodes the list of all .in files, so the "minimal" release directory still must contain a copy of all of them.

fare:
GNU make considers that the # in arm#.scm is a comment starter, when expanding boot/gsc-boot: @GSC_SCM@

feeley:
Can't it be escaped? In gsc/makefile there is

```
RCFILES = makefile.in fixnum.scm generic.scm igsc.scm $(MODULES_SCM) \
$(LIBRARIES_SCM) _envadt.scm _gvmadt.scm _ptreeadt.scm _sourceadt.scm \
_univadt.scm runtests.scm _t-x86.scm _arm.scm "_arm\#.scm"
```

fare:
It can probably be escaped. Not sure what the right formula is to do it from configure.ac, though. Probably some sed regex magic.

fare:
Here, I used sed to escape, and it works at least using GNU make.

feeley:
Do you know why the build fails on linux on travis?

feeley:
I've implemented your feature request in a way that matches my view of the system, and in the process simplified the steps to build Gambit from the github repo.  It is commit b9c69958111166ff17dceb051d03b60d81fb917d .  Could you please have a look?

fare:
Very nice! And indeed better than what I offered, modulo a few comments I added on github for commit b9c6995 : using && and better use of git.

fare:
I updated this PR with a proposed patch to commit b9c6995 : using && and better use of git.

If `git worktree` is too recent (July 2015), then an alternative is `git archive ... | (cd ... ; tar xf -)`

fare:
Problem. When I try to build with this patch, I get this error:
```
making all in lib
make[3]: Entering directory '/home/fare/src/scheme/gambit/boot/wip-compiler/lib'
LD_LIBRARY_PATH=..:/home/fare/local/lib ../gsc-boot -:~~bin=../bin,~~lib=../lib,~~include=../include -f -c  -check -prelude "(##include \"header.scm\")" _io.scm
../gsc-boot: symbol lookup error: ../gsc-boot: undefined symbol: ___S__23__23_type_2d_18_2d_2babe060_2d_9af6_2d_456f_2d_a26e_2d_40b592f690ec
makefile:167: recipe for target '_io.c' failed
make[3]: *** [_io.c] Error 127
make[3]: Leaving directory '/home/fare/src/scheme/gambit/boot/wip-compiler/lib'
makefile:435: recipe for target 'all-recursive' failed
make[2]: *** [all-recursive] Error 1
```

Note: my `$(CONFIGURE_COMMAND)` is:
```
./configure '--prefix=/home/fare/local/stow/gambit' '--enable-single-host' '--enable-c-opt=-O2' '--enable-gcc-opts' '--enable-shared' '--enable-absolute-shared-libs' '--enable-poll' '--enable-openssl'
```

fare:
It works better when `$(CONFIGURE_COMMAND)` is just `./configure --enable-single-host` or `./configure` which is what it was before. Shouldn't that be what the makefile should use?

fare:
OK, I modified the `makefile.in` to use `./configure --enable-single-host` when compiling the `boot/` compilers.

feeley:
The reason ```$(CONFIGURE_COMMAND)``` is used is so that the ```CC=...``` configure option is propagated to the sub-configurations.  It matters on MacOS where the default gcc is the llvm one, and it causes wrong code to be generated and it is much slower to compile.

feeley:
Please avoid ```git worktree``` is is too recent (for example I'm still using git 1.8 and even if I upgrade there may be other users in the same situation).

feeley:
And in the ```boot-wip-compiler-gsc-gsc``` target, don't ```$(MAKE) bootstrap```.  That will create boot/wip-compiler/gsc-boot but that is not correct, because that compiler will not necessarily be able to build the sources in boot/wip-compiler.  The makefiles don't currently need to build the sources in boot/wip-compiler a second time, but if ever someone modifies a .scm file in that directory and does a ```make``` it would be nice if the build completes without giving strange errors. In other words, boot/wip-compiler/gsc/gsc must be built with boot/most-recent-release/gsc-boot and there is no reason to want to use a "newer" compiler.

feeley:
It would be interesting to figure out why the build fails when using ```---enable-shared```.  I suspect that copying gsc-boot is insufficient because the shared lib for the Gambit library is also needed.  I'm not sure what is the proper practice for such things.  I always compile with static linking to avoid these issues...

fare:
OK, I can use `git archive | tar` instead of `git worktree`.

As to the configure command: my understanding is that following the instructions with the previous boot process, one would just use `./configure`, and that used to work. Same thing with the previous boot process using `make bootstrap` for the wip-compiler. Of course, you know better than anyone what things *should* be, but then, I require more of your guidance in determining what to do.

If `--enable-shared` is the problem, would `--disable-shared` before and/or after solve the issue? and/or a sed command to turn one into the other? (Yuck)

feeley:
It is still possible to configure again after the ```make current-gsc-boot```, i.e.

```
./configure <options1>
make current-gsc-boot
./configure <options2>
make from-scratch
```

So ```<options1>``` will apply to the ```gsc-boot``` that is used by the ```make from-scratch```, and ```<options2>``` will be for the final system.

But I like the single configure approach because it simplifies things for newbees.

I see that when ```--enable-shared``` is used the configure script is setting up ```SETDLPATHBOOT``` like this:

```SETDLPATHBOOT="LD_LIBRARY_PATH=\$(rootfromhere):\${LD_LIBRARY_PATH}"

This is the ```LD_LIBRARY_PATH``` used when invoking gsc-boot.  On a ```make current-gsc-boot``` this should be made to point to boot/most-recent-release, but there's no way for the configure script to know this.

I wonder if it wouldn't be more appropriate for ```gsc-boot``` to be a script that sets up the correct ```LD_LIBRARY_PATH``` so that when ```gsc-boot``` is copied to boot/current-compiler it would still work properly when the system was configured with ```--enable-shared```.

Another solution is for ```make current-gsc-boot``` to do sub-configure with no options (to get a static link) but to propagate the ```CC=``` option in case a specific C compiler was requested.  I kind of prefer this solution because there's no real reason for all the configure options of the final system to apply to the bootstrap compiler.  In fact, for robustness, it is best when fewer configure options are used.

**Comment by feeley** [comment]
This discussion makes me realize that ```make bootstrap``` simply doesn't work when the system is built with ```--enable-shared```.  That's because the gsc-boot executable that is created is dependent on shared libraries that aren't packaged with gsc-boot and can change with subsequent changes to the sources even if there is no subsequent ```make bootstrap```.  So a ```make bootstrap``` should take a snapshot of the shared libraries in addition to gsc/gsc, so gsc-boot probably has to be a script.

**Comment by fare** [comment]
What would be the correct way to capture and re-export `CC=`, if it's not done already?

I admit I too like that there should be a single ./configure with a single set of option, and that the bootstrap options should be either constant or deduced from the above.

**Comment by feeley** [comment]
The C compiler is in the ```C_COMPILER``` environment variable of the configure script, and ends up in the same variable in all the makefiles.  So that should be easy to get to.

**Comment by fare** [comment]
So something like that?
`./configure --enable-single-host C_COMPILER='$(C_COMPILER)'`

**Comment by feeley** [comment]
Rather,

```./configure --enable-single-host CC='$(C_COMPILER)'```

**Comment by feeley** [comment]
Actually, to support C++, probably better is:

```./configure CC='$(C_COMPILER)' CXX='$(C_COMPILER)'```

**Comment by feeley** [comment]
Scratch that... it should simply be:

```./configure CC='$(C_COMPILER)'```

**Comment by fare** [comment]
OK, PTAL.

**Comment by fare** [comment]
Also, could and should the wip-compiler tree be cp -r'ed from the most-recent-release tree including intermediate results and configuration, to save some build time?

**Comment by feeley** [comment]
Well copying it would save time, but determining what needs to be copied would break an abstraction barrier... I could say (with high confidence) that you simply need to copy all the .o files, but what if this changes in the future due to some new build strategy.  More robust would be to "mv boot/most-recent-release boot/current-gsc-boot" just before copying the gsc/*.scm files into boot/current-gsc-boot/gsc.  After all once the boot/most-recent-release/gsc-boot is built, there is no use for the rest of the boot/most-recent-release directory.

**Comment by fare** [comment]
I was thinking: copy everything from most-recent-release to wip-compiler, then replace the wip-compiler/gsc directory, then rebuild.

**Comment by fare** [comment]
One slight possible reason for copying rather than mv: for faster clean build when gsc changes.

**Comment by fare** [comment]
PTAL

**Comment by feeley** [comment]
This would not affect the speed of a clean build when gsc sources change.  That's because the gsc-boot that was built for the most recent release is being used.  So lets say we use a single directory, boot/wip-compiler, which is initially a clone (or archive) of the latest release, it is built to produce gsc-boot, then the gsc/*.scm files are copied to boot/wip-compiler/gsc and then after building there, boot/wip-compiler/gsc/gsc is used as the gsc-boot at the root.  If the gsc/*.scm files are updated the version of boot/wip-compiler/gsc-boot will still match the latest release, so the build process can be restarted at the "gsc/*.scm files are copied to boot/wip-compiler/gsc".

**Comment by fare** [comment]
Oh, now I see you copy boot/wip-compiler/gsc/gsc to not boot/wip-compiler/gsc-boot in the last step.

1- I suppose it could all happen in one boot directory indeed. Is that what you want?

2- Is it necessary to $(MAKE) in all subdirectories? It looks to me like a subset of them would suffice which would speed up the build slightly.

**Comment by fare** [comment]
PTAL

**Comment by feeley** [comment]
Yes I think one directory is good.  Trying to optimize the "make" is probably not worth it (will save little time, and will break abstractions).  But if you want to experiment...

```
cd lib; make; cd ..
cd gsi; make; cd ..
cd gsc; make; cd ..
```

fare:
I merged those patches and did as suggested. PTAL.

fare:
thanks!

PR #289, opened 2017-07-26
Revamp of `gambdoc` for ease of use in shell

mjsir911:
* Made gambdoc easily runnable in shell (still need to set `GAMBDOC_GAMBITDIR_DOC` variable)
* Made gambdoc  default to `$BROWSER` if no default browser set anywhere else: in _bash_ or in _gsi_
* `GAMBDOC_GAMBITDIR_BIN` is always set to `dirname $0`, the directory of running file
* Reworded variables in gambdoc for readability
  - GAMBDOC_ARG1: Same, now defaults to `help` if no `$1` provided
  - GAMBDOC_ARG2: seemed to be unused before, `$2` if specified, if not default to value of `$GAMBDOC_ARG4`
  - GAMBDOC_ARG3 => GAMBDOC_BROWSER
  - GAMBDOC_ARG4 => GAMBDOC_ARG2
* For compatibility all the old variables still work, it just reassigns them
  - **Note**: I don't understand how in _gsi_ the initial variables get set by calling `(help *keyword*)`. In the future if I look into it a bit more, I might try hardcoding the help procedure to output on the changed variables specified above, or even just `$1` and `$2`
* Reworked browser list to handle default browser with spaces
  - previously iterated through string, using spaces as delimiter. Broke if browser was path to executable with spaces in it (mac doing mac stuff)

This merge is mainly to help beginners in gambit/scheme (as I am) a bit better, as gambdoc was an executable in the `$PATH`, with cryptic input arguments passed through environmental variables. I tried to simplify and document through code how It's working a bit.


feeley:
Thank you for your effort to improve the Gambit documentation browser. I have a few comments.  Please make a new pull request based on my comments.

Could you explain why you rewrote ```find_in_path``` using ```which```?  Was the original implementation not working and if so in what case?  Using ```which``` creates a dependency on external functionality and this creates a portability problem (see https://unix.stackexchange.com/questions/85249/why-not-use-which-what-to-use-then).  The original implementation only depends on /bin/sh with very little variation possible.

The ```gambdoc``` script was not meant to be called interactively.  I like the idea of allowing the user to call it from outside of Gambit, for example from an interactive shell.  However this should be viewed as a secondary use (so the arguments to the script should not be renamed, but rather they should have defaults).  By the way, the arguments to the script (```GAMBDOC_ARG1```, ...) are generated programmatically in ```##gambdoc``` (defined in the file ```lib/_repl.scm```).

A command-line "usage" instructions should be output if ```GAMBDOC_ARG1``` is undefined and ```$1``` is not supplied.  Otherwise:

```GAMBDOC_ARG1``` should default to ```help```.

```GAMBDOC_ARG2``` should default to ```$1```.

```GAMBDOC_ARG3``` should default to nothing (so the browser is searched for in the PATH).

```GAMBDOC_ARG4``` should default to a link created from ```"Definition of "``` and ```$1``` with appropriate URL percent escapes.

The ```GAMBDOC_GAMBITDIR_BIN``` and ```GAMBDOC_GAMBITDIR_DOC``` variables could easily have defaults provided by the ```configure``` script which has the paths in ```$bindir``` and ```$docdir```.  In the configure script, these could be assigned to ```GAMBITDIR_BIN``` and ```GAMBITDIR_DOC```.  Then adding ```AC_SUBST(GAMBITDIR_BIN)``` and ```AC_SUBST(GAMBITDIR_DOC)``` will cause the patterns ```@GAMBITDIR_BIN@``` and ```@GAMBITDIR_DOC@``` to be substituted by the paths.  The ```gambdoc``` script can then use these patterns as defaults for the ```GAMBDOC_GAMBITDIR_BIN``` and ```GAMBDOC_GAMBITDIR_DOC``` variables.

Finally, it would be good for the Windows script to be updated to behave similarly.

mjsir911:
Thanks for your feedback

In retrospect, the redoing of `find_in_path` was unnecessary and does add additional dependencies

I can work on simple usage instructions, but I feel it is more intuitive to not have to pass an argument to `gambdoc`, and if no argument is passed just open up the table of contents, without a section to jump to; `GAMBDOC_ARG4 == ""`. How do you feel about that? The question there though is how do you make it print the usage instructions then.

Thanks for explaining what the arguments are a bit more, I'll revert them back and try explaining what they are in the code a bit.

How do you feel about the `GAMBDOC_ARG3` variable defaulting to the `$BROWSER` variable, as it is a standard, but not very [well defined](https://unix.stackexchange.com/questions/213367/where-do-editor-pager-browser-environment-variables-come-from)

The `GAMBDOC_GAMBITDIR_BIN` and `GAMBDOC_GAMBITDIR_DOC` suggestions are good, and I'll try to put those in place, as you've explained it simply enough.

My batch isn't the best, but I'll see what I can do with the `gambdoc.bat` file

Issue #243, opened 2017-04-25
Universal backend: Implement eq? tables


feeley:
Estimated time: 40 hours

The target languages have support for hash tables.  However it is not clear if ```eq?``` hash tables (where the keys are compared using object identity, i.e. ```eq?```) are possible.  One idea for JavaScript is to add a unique "serial number" field to objects dynamically when they are used as a key of a hash table.

316k:
Just a though, string=? hash tables could be implemented in a similar way in JavaScript, Python and PHP, without even requiring a serial number.
Plus, string=? tables could be easily integrated to scm2host and host2scm functions.

Issue #242, opened 2017-04-25
Universal backend: Implement wills and weak references


feeley:
Estimated time: 40 hours

The will objects and hash tables can contain weak references.  Is it possible to implement this in the target languages with the existing garbage collectors?

316k:
* JS : It would be possible to implement weak *keys* in ES6 using [the WeakMap object](https://developer.mozilla.org/en/docs/Web/JavaScript/Reference/Global_Objects/WeakMap), but not weak *values*.

* Python : Weak references are available with `import weakref` (see https://stackoverflow.com/questions/9908013/weak-references-in-python#9908216)

* In Ruby, there is [the WeakRef class](http://ruby-doc.org/stdlib-2.0.0/libdoc/weakref/rdoc/WeakRef.html)
The example in the documentation shows a simple way to create a hash table with weak keys and weak values, we could base our code on that.

* Java has a WeakReference class, https://docs.oracle.com/javase/7/docs/api/java/lang/ref/WeakReference.html

* In PHP, ~weak references are probably represented with strings, just like everything else~ there surprisingly is a Weakref class, although it's a PECL extension and it is not installed by default.

feeley:
Could you elaborate for each target to give the rough outline of how this would work?

316k:
This issue is closely related to #243.

The current representation for hash tables in the universal backend is an alist, but for some test functions (such as eq?), we can use the native hash tables available in the host languages.

If a good abstraction is possible, supporting weak keys/values in native hash-tables should be easy.


## Implementation ideas

### Ruby

From [the example in the documentation](http://ruby-doc.org/stdlib-2.0.0/libdoc/weakref/rdoc/WeakRef.html), I wrote three classes : https://gist.github.com/316k/61a898e3add48e1f4b55eb42d9aa0245

These should allow us to use a `Weak...Hash` the exact same way we would use a `Hash`, so weak keys/values would be really painless to implement.

I don't like the fact that most operations have to go through a `begin...rescue Exception`, but I haven't found any other way to check if a WeakRef is pointing to a dead object or not.


### Python

In Python, with `import weakref`, we can have access to `WeakKeyDictionary` and `WeakValueDictionary`, but there seems to be nothing built-in for a weak-keys and weak-values dictionary. Again, I wrote classes : https://gist.github.com/316k/90a1aa5edec878915df5dd5fcb0bf873 that would allow us to use the weak hash tables painlessly.


### Js

[...]

PR #292, opened 2017-08-04
Universal backend: implement native weak-key/values hashtables where available

316k:
See #243 and #242


gsc/_t-univ-4.scm
```diff
@@ -2902,7 +2902,7 @@
     (return arg1)))
  (make-translated-operand-generator
   (lambda (ctx return arg1)
-    (return (^not (^eq? (^cast*-scmobj arg1) (^obj #f)))))))
+    (return (^not (^parens (^eq? (^cast*-scmobj arg1) (^obj #f))))))))
```
316k: This is what caused the if(!obj === false) bug

PR #313, opened 2017-11-20
[WIP] TLS certificate validation

### LINKED ISSUE #200: TLS certificates are not verified
I was very happy to see the native OpenSSL support (issue #123), but when I tried it out, the client would not (and I could not make it) verify the server's certificate. This completely defeats the security as the client could just be creating an impeccably secure connection directly to an attacker's proxy.

Looking at lib/os_io.c, it seems to call SSL_CTX_set_verify only in server mode (if you set request-client-authentication). It should be called in client mode as well (and set SSL_VERIFY_PEER).

Try this slightly modified example from the documentation (section 17.7.3):

Server:

``` scheme
(define ctx (make-tls-context options: '(server-mode)
                              certificate: "server.pem"))

(define s (open-tcp-server (list server-address: "localhost"
                                 port-number: 1443
                                 tls-context: ctx)))
(define p (read s))
(display "Hello from the server!\n" p)
(force-output p)
```

Client:

``` scheme
(define ctx (make-tls-context))
(define c (open-tcp-client (list server-address: "localhost"
                                 port-number: 1443
                                 tls-context: ctx)))
(display "Hello from the client!" c)
(force-output c)
(display (read-line c))
(newline)
```

Generate a self-signed certificate which nobody has any reason to trust:

``` bash
openssl req -new -x509 -days 7 -nodes -out server.pem -keyout server.pem
```

...and run it. The client connects just fine to the server, even though 1) it has not been told in any way which certificate the server is supposed to use and 2) the server's certificate is not signed by anybody. An attacker could set up their own server with their own certificate and the client would obliviously connect to it.

I believe the make-tls-context API needs something along the lines of [what Python has](https://docs.python.org/3/library/ssl.html#ssl.SSLContext) which can be used like this (for connecting to a server with a known certificate):

``` python
ctx = ssl.SSLContext(ssl.PROTOCOL_SSLv23)
ctx.load_verify_locations(cadata=my_string_of_trusted_certificate_data)
ctx.verify_mode = ssl.CERT_REQUIRED
ctx.check_hostname = False
```

For using certificates signed by CAs we would also need equivalents of SSLContext.set_default_verify_paths and make sure that a hostname can be passed through open-tcp-client to work with SNI (with ctx.check_hostname = True).

### LINKED ISSUE #312: open-tcp-client doesn't validate the server certificate
As it stands, the TLS implementation doesn't seem to validate server certificates in client-side connections.

For example:
```
> (define ctx (make-tls-context))
> (define c (open-tcp-client `(server-address: "untrusted-root.badssl.com" port-number: 443 tls-context: ,ctx))
> (display "GET / HTTP/1.1\r\nHost: untrusted-root.badssl.com:443\r\n\r\n" c)
> (force-output c)
> (read-line c)
"HTTP/1.1 200 OK\r"
```

Try the same with curl and we get:
```
$ curl https://untrusted-root.badssl.com
curl: (60) SSL certificate problem: unable to get local issuer certificate
```

vyzo:
Closes #312 
Closes #200 

Adds clients-side TLS certificate validation.

This the first pass, which adds validation by default, using the default certificate chain. It is sufficient for the general usage of TLS clients.

Nonetheless, there are a couple of special cases we also want to support:
- Allow custom verify paths (necessary for custom certificate authorities and testing)
- Allow an insecure mode where we can disable certificate validation (testing, misconfigured servers)

Also of note: there is no special error reported by certificate validation failure, just a delayed SSL error when we force output in the port. This is far from ideal, we would like to have a nice loud and clear error when certificate validation fails.

### DIFF
--- a/lib/os_io.c
+++ b/lib/os_io.c
@@ -4460,6 +4460,10 @@ ___SCMOBJ client_ca_path;)
       SSL_CTX_set_mode (c->tls_ctx,
                         SSL_MODE_ENABLE_PARTIAL_WRITE |
                         SSL_MODE_ACCEPT_MOVING_WRITE_BUFFER);
+
+      /* always verify server certificate */
+      SSL_CTX_set_default_verify_paths (c->tls_ctx);
+      SSL_CTX_set_verify (c->tls_ctx, SSL_VERIFY_PEER, NULL);
   }
 
   if ((scm_e =___NONNULLPOINTER_to_SCMOBJ


vyzo:
Should I open issues for the notes for the WIP parts?

feeley:
Yes please do.

PR #316, opened 2017-12-14
thread-group-specific

vyzo:
Adds a specific field to thread-groups, similar to other structures. This allows us to easily associate shared state for all threads in a thread group.

The patch commandeers tgroup-unused1 to act as the specific field and adds `thread-group-specific` and `thread-group-specific-set!` primitives.


vyzo:
squashed to a single commit

vyzo:
@feeley ping

PR #318, opened 2017-12-25
Implement ,stg command in repl

vyzo:
Adds a new repl command, `.stg` and `,(stg X)` to recursively display thread group state.
This allows us to easily see _all_ threads in a thread group, including the ones in the nested thread groups.
In particular,  `,(stg #1)` will show us all running threads in the system.


vyzo:
I wonder whether we should also add a parameter to top to utilize the new recursive? option in `##display-thread-group-state`.

vyzo:
@feeley ping

feeley:
Will look into this.

lassik:
Would be useful to standardize a command to get information about threads in a SRFI. A few Schemes now have them. Of course, there can still be a comma command idiomatic to Gambit as well.

PR #321, opened 2017-12-27
thread-send-hook: A hook for debugging actor systems

vyzo:
Debugging actor systems is hard, and it's going to get harder with the smp scheduler.

This patch adds a hook to `##thread-send`, aptly called `##thread-send-hook` which is invoked before a message is queued for another thread. The hook is settable with `##thread-send-hook-set!`

This hook allows us to trace all messages in the system, which is invaluable for debugging actor systems.
It also empowers us to implement additional functionality, like validating messages or thread state, restricting the flow of messages between certain thread-groups for actor isolation, and so on.

cc @fare


vyzo:
squashed to a single commit.

fare:
```
(def (current-thread-group)
  (thread-thread-group (current-thread)))

(def (primordial-thread-group)
  (let loop ((tg (current-thread-group)))
    (if-let ((up (thread-group-parent tg)))
      (loop up)
      tg)))

(def (all-threads (group (primordial-thread-group)))
  (with-list-builder (c)
    (let loop ((g group))
      (for-each c (thread-group->thread-list g))
      (for-each loop (thread-group->thread-group-list g)))))

(def (thread-named name)
  (find (λ (t) (equal? (thread-name t) name)) (all-threads)))
```

Example use:
```
(set! ##thread-send-hook (let (th (thread-named "actor")) (lambda (dst msg) (let ((src (current-thread))) (when (or (equal? th src) (equal? th dst)) (write (list 'thread-send src dst msg)) (terpri))))))
```

vyzo:
you probably want to write to `##stdout-port` or `##console-port` in the example above, otherwise your writes will go to the sending thread's current output port!

feeley:
Hmmm... I agree a hook on ```thread-send``` would be useful for debugging.  Then again, there are many places where having a similar hook would be useful, so where to stop? Moreover in the specific case of ```thread-send``` why not simply redefine ```thread-send``` so that it includes the hook? Something like:

```
(set! thread-send
      (let ((old-thread-send thread-send))
        (lambda (thread msg)
          (hook thread msg)
          (old-thread-send thread msg))))
```

Or define your extended version of ```thread-send``` in a debug module, and use the module system to redirect calls to ```thread-send``` to your extended version.

Also your proposed hook can only observe the messages that are sent.  I think it is more general to allow modifying the message, similarly to the ```filter``` parameter of ```make-parameter```. This function can observe the message, but also transform it or wrap it within some other object before sending it (either for adding a "return address", a serial number, etc).

vyzo:
Using the module system and redefining `thread-send` was my first thoughts in implementing the hook.

Using the module system has the problem that it only captures code that uses the module system and not pure gambit code; also there are a couple more primitives in gerbil that do sends before calling `##thread-send` directly (and not `thread-send`) and I most certainly prefer doing it once instead of multiple times.

Setting `thread-send` to a hooked version is possible, but I am not sure I like the practice of redefining such primitives; also, does not capture uses of `##thread-send` directly, as above.

vyzo:
> Also your proposed hook can only observe the messages that are sent. I think it is more general to allow modifying the message, similarly to the filter parameter of make-parameter. This function can observe the message, but also transform it or wrap it within some other object before sending it

I can modify the patch to change the hook to have a filtering behaviour

fare:
Indeed, maybe a general purpose ADVICE facility would be more general than this thing
as long as indeed we can be sure that thread-send won't be inlined and we can override it.

Optional features of an ADVICE facility:
1- Can define `:around` `:after` `:before` methods, maybe more.
2- Can give a name or tag to each advice to selectively enable, modify or disable each of them.
3- It is possible to redefine the surrounding advice function (hence remember the original).
4- It is possible to redefine the original (hence remember the advices).
5- The "effective" function is dynamically recomputed when any of the original function or its advices are modified.

vyzo:
@fare I think `defadvice` is a bit out of scope for this PR :)

Thinking a bit more about this, one other good reason that I would like the hook in core (aside from having to change quite a bit of code and the unsolvable problem of leak through extern code), is that I think it's beneficial for all Gambit users (not just Gerbil). 

When you are writing actor based code, sooner or later you will have a bug to debug that requires tracing. So then you would have to reinvent the wheel with doing a `(set! thread-send ...)` and also shoot yourself in the foot because you forgot to save the original `thread-send` to turn your hook back off.

I am in agreement that since we are putting it there, we might as well generalize it to have other sensible uses. March's suggestion for a filter is the most general and useful one -- and auto-wrapping messages is a very nice trick.

Implementation forthcoming.

vyzo:
@feeley ping -- thoughts?

PR #324, opened 2018-01-03
define tgroup-specific field in SMP branch

vyzo:
the thread-group structure is missing the specific field in the SMP branch; mea cupla.


vyzo:
@feeley ping -- smp thread-group-specific is broken.

Issue #330, opened 2018-01-10
Race condition running tests 7 and 10 in parallel


abdelq:
Tests failed on a Hewlett Packard Precision Architecture (HP/PA) server due to a race condition between
parallel runs of tests 7 and 10. Technically, it could happen in other architectures as well:

    ------------ TEST 7 (memory management and C-interface)
    rm -f mem.c mem_.c mem.o mem_.o mem
    LD_LIBRARY_PATH=../lib:../gsi:../gsc: ../gsc/gsc -:~~bin=../bin,~~lib=../lib,~~include=../include -f -warnings -o mem -exe mem.scm
    56.590734000000005 secs elapsed cpu time
    heartbeat frequency = 217.70348481431606 Hz
    [...]
    ------------ TEST 10 (consistency of lib/gambit#.scm)
    rm -rf mem.o mem.o*
    LD_LIBRARY_PATH=../lib:../gsi:../gsc: ../gsc/gsc -:~~bin=../bin,~~lib=../lib,~~include=../include -f -i test10.scm
    gcc: error: mem.o: No such file or directory
    makefile:148: recipe for target 'test7' failed
    make[2]: *** [test7] Error 70
    [...]

[Full log](https://buildd.debian.org/status/fetch.php?pkg=gambc&arch=hppa&ver=4.8.8-2&stamp=1515173584&raw=0)

feeley:
I assume you got that with a ```make -j check```?  That's not the way to run the self checks...  I guess the makefile could be changed to force sequential execution when using a parallel make.

PR #331, opened 2018-01-11
Run tests serially

abdelq:
#330

### DIFF
--- a/tests/makefile.in
+++ b/tests/makefile.in
@@ -102,6 +102,8 @@ all-post:
 	  touch server.rm; \
 	fi
 
+.NOTPARALLEL:
+
 check: all test1 test2 test3 test4 test5 test6 test7 test8 test9 test10 test11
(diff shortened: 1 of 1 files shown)


feeley:
When was the .NOPARALLEL make option introduced?  Is it portable to all make implementations? It is probably possible to achieve the same thing with standard makefile rules by adding fake dependencies.

abdelq:
The syntax previously used for this PR is quite recent (post-2011). I changed it for an older one.
`.NOTPARALLEL` was introduced in GNU Make somewhere before 2006. I would have to dig years before to find the exact date.

> It is probably possible to achieve the same thing with standard makefile rules by adding fake dependencies.

It might be possible, but the end result would be quite convoluted, especially for a modification as minor as this.

feeley:
Indeed.  I think on old "make" implementations it will be ignored because it looks like a rule for a .NOTPARALLEL target.

PR #341, opened 2018-03-14
Fix the config script for windows to handle CONF_USE_SIGSET_T

ckkashyap:
I took the liberty of creating the PR even before discussing. This change will only affect build on windows that use uses the js file so it should be safe.
After applying this change, I can build gsc/gsi using 32bit CL.


feeley:
Thanks!

PR #348, opened 2018-04-06
use SNI for TLS connections

### LINKED ISSUE #347: Cannot open TLS connections to servers that require SNI
Any attempt to connect to a server that requires SNI results in handshake failure -- for instance it's impossible to connect to `https://www.clojure.org` because of this.

We need to set the SNI with `SSL_set_tlsext_host_name` when creating the SSL socket with `SSL_new`; however, this is created in  `___device_tcp_client_setup_from_sockaddr` and there doesn't seem to be an obvious way to get the actual name server there.

vyzo:
Fixes #347

### DIFF
--- a/lib/_io.scm
+++ b/lib/_io.scm
@@ -7043,7 +7043,8 @@
                          address)
                      port-number
                      (psettings->options psettings)
-                     (macro-psettings-tls-context psettings))))
+                     (macro-psettings-tls-context psettings)
+                     (and (##string? address-or-host) address-or-host))))
                (if (##fixnum? device)
                    (if raise-os-exception?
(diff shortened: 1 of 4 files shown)


vyzo:
should be ready for review now

feeley:
When is server_name_p released?

vyzo:
It is released in `___device_tcp_client_setup_from_sockaddr`.
Maybe I missed some paths though and this can leak.

vyzo:
It can leak indeed, if `___SCMOBJ_to_sockaddr` fails for instance; let me try and fortify.

vyzo:
It might be better to release it in `___os_device_tcp_client_open` so that the reference is only borrowed and we don't ahve to duplicate error path releases.
Fixing.

PR #360, opened 2018-07-01
Sheme character encoding/decoding, JS I/O developpement

deladurantayefrancis:
Adding Scheme versions of functions for character encoding and decoding.

Implemented some functions for basic I/O when compiling to JS.


feeley:
The commit should not include all the generated .c files.  They can be unselected with "make deselect-gen-for-commit".

Also, please make a pull request that doesn't cause the Travis CI build to fail!

PR #364, opened 2018-09-05
Scheme version of character encoding/decoding, extended I/O subsystem for JS

deladurantayefrancis:


feeley:
If you check the output of travis-ci you will see it ends with

```
------------ TEST 10 (consistency of lib/gambit#.scm)
rm -rf mem.o mem.o*
../gsc/gsc -:~~bin=../bin,~~lib=../lib,~~include=../include -f -i test10.scm
************ file lib/gambit#.scm needs to be edited ************
==== these names should be ADDED ====
err-code-from-char-encoding
make[1]: *** [test10] Error 1
make[1]: Leaving directory `/home/travis/build/gambit/gambit/tests'
make: *** [check] Error 2
```

Which means that in your changes you have added the top-level ```err-code-from-char-encoding``` procedure.  This name should be prefixed with ```##``` and it should be protected in a

```
(macro-case-target
 ((js)
  ...))
```

So that it does not interfere with other backends.  In particular, there is no reason to add new definitions when the C backend is used.

PR #365, opened 2018-09-09
Compatibility with OpenSSL version 1.1.

skilldown:
Version 1.1 of OpenSSL introduced several non-backwards compatible
changes that broke the building of Gambit with OpenSSL features.  This
change addresses those differences and allows Gambit to use the new
version.

Note that using SSL v2 as a minimum protocol version is no longer
supported in OpenSSL 1.1.  As a result using ssl-v2 as the minimum
version in a call to make-tls-context will raise a "Wrong TLS version"
exception.

### DIFF
--- a/lib/os_io.c
+++ b/lib/os_io.c
@@ -3998,7 +3998,7 @@ ___SCMOBJ client_ca_path;)
   if (openssl_initialized == 0)
     {
 #if OPENSSL_VERSION_NUMBER >= 0x10100000L
-        if (!OPENSSL_init_ssl(0, NULL))
+      if (!OPENSSL_init_ssl(0, NULL))
         return ___FIX(___TLS_LIBRARY_INIT_ERR);
 #else
       if (!SSL_library_init())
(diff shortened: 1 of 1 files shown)


lib/os_io.c
```diff
@@ -3996,11 +3997,17 @@ ___SCMOBJ client_ca_path;)
      https://github.com/lighttpd/lighttpd1.4/blob/master/src/network.c */
   if (openssl_initialized == 0)
     {
+#if OPENSSL_VERSION_NUMBER >= 0x10100000L
+        if (!OPENSSL_init_ssl(0, NULL))
+        return ___FIX(___TLS_LIBRARY_INIT_ERR);
```
feeley: please indent

skilldown:
I tested the change against OpenSSL 1.1.0g and 1.0.2g and didn't find any issues.

I missed the indentation issue!  A commit has been added to correct it.  Thanks for the quick review!

Issue #366, opened 2018-09-19
Gambit doesn't work with LibreSSL


zmyrgel:
Gambits TLS functionality doesn't work with LibreSSL.
Gambit compiles just with `--enable-openssl` flag enabled but when trying to use TLS it errors out. Simple example as follows:

```
 $ gsi
Gambit v4.9.0

> (make-tls-context)
*** ERROR IN (console)@1.1 -- 
(make-tls-context)
1> 
```
I'm testing this with OpenBSD-current which uses libressl 2.8.0. I asked around on IRC from void linux users and same thing seems to happen there as well. Void linux also uses LibreSSL instead of OpenSSL so I assume LibreSSL is root of this issue.

LibreSSL API should be subset of OpenSSL 1.0.1 API so I would have expected Gambit to have worked with it just fine. If any, I would have expected issues during compilation. As the code compiles I'm not sure what is wrong with the code and the generic error doesn't help pointing to the root of this.

zmyrgel:
Possibly found the issue, libressl doesn't support sslv2/v3 anymore so the min_tls_version check fails. I'll cook a patch for it. I have simple version to make tls 1.2, 1.1 working but setting min version to tls1.0 fails at the moment.

belmarca:
PR https://github.com/gambit/gambit/pull/367 is supposed to fix this issue and has already been merged. Can this one be closed?

juanfra684:
@belmarca yes, gambit has been building fine with libressl for a while on OpenBSD.

gambiteer:
Reported to work

PR #367, opened 2018-09-20
fix make-tls-context with LibreSSL

zmyrgel:
LibreSSL no longer supports SSL protocols and so has set old SSL_OP_NO_SSLv3 and
SSL_NO_SSLv2 defines to 0. Make the current TLS configuration setup skip using
them when using LibreSSL. Fixes the #366.

### DIFF
--- a/lib/os_io.c
+++ b/lib/os_io.c
@@ -4146,12 +4146,12 @@ ___SCMOBJ client_ca_path;)
                                 SSL_CTX_set_options (c->tls_ctx,
                                                      SSL_OP_NO_TLSv1)));
         case 0x301:
-#ifndef LIBRESSL_VERSION_NUMBER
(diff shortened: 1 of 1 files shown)


adam-7:
@zmyrgel , i see you condition SSL v2&3 use against the LIBRESSL define here.

Doesn't OpenSSL provide the option to _not_ compile in v2&3 support so there's some other more general define you could use instead?

At least just to double-check.

zmyrgel:
Found OPENSSL_NO_SSL3/2 defines which are present in LibreSSL and OpenSSL so it seems to be better define to check. Not sure which OpenSSL version introduced that though.
(make-tls-context) call at least works on OpenBSD using latest LibreSSL. Haven't tested the diff on other systems.

PR #395, opened 2019-06-03
Add ___ps argument to mem.c's ___glo_list_search_obj() & ___obj_to_global_var(), making --enable-multiple-vms build again.

adam-7:
* Before merging please validate that this is an existing problem indeed, in the related work I have had friction from the Windows WSL Linux build platform. *

"./configure --enable-multiple-vms ...; make" for Gambit master 2019-06-03 and at least back to 2018-11 has compile failure in mem.c because ___ps is missing in ___glo_list_search_obj, this commit leads ___ps all the way from the call chain's source in _kernel.scm's |##object->global-var| via mem.c's ___obj_to_global_var and then to there.

gcc ... mem.c ...
In file included from mem.c:7:0:
mem.c: In function ‘___glo_list_search_obj’:
../include/gambit.h:4426:62: error: ‘___ps’ undeclared (first use in this function)
 #define ___GLOCELL(x)___GLOCELL_IN_VM(___VMSTATE_FROM_PSTATE(___ps),x)

### DIFF
--- a/lib/_kernel.scm
+++ b/lib/_kernel.scm
@@ -3994,7 +3994,7 @@ end-of-code
 
 (define-prim (##object->global-var obj primitive?)
   (##c-code
-   "___RESULT = ___obj_to_global_var (___ARG1, !___FALSEP(___ARG2));"
+   "___RESULT = ___obj_to_global_var (___ps, ___ARG1, !___FALSEP(___ARG2));"
    obj
    primitive?))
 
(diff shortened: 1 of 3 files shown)


philsuero:
I've got the same error compiling with gcc-8 and g++-8 on mac.

adam-7:
This issue fixed by commit https://github.com/gambit/gambit/commit/a13236e1f062a860141404ee7415761be927a816 , dismissing this PR.

PR #412, opened 2019-10-17
Don't use sysctl and sysctlbyname on Linux

abdelq:
Has been deprecated and will probably be removed at some point or another.

### DIFF
--- a/lib/os.h
+++ b/lib/os.h
@@ -281,11 +281,15 @@
 #endif
 
 #ifdef HAVE_SYSCTL
+#ifndef __linux__
 #define USE_sysctl
+#endif
 #endif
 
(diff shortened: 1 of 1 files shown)


lib/os.h
```diff
@@ -280,11 +280,11 @@
...
 
-#ifdef HAVE_SYSCTL
+#if defined(HAVE_SYSCTL) && !defined(__linux__)
 #define USE_sysctl
 #endif
 
-#ifdef HAVE_SYSCTLBYNAME
+#if defined(HAVE_SYSCTLBYNAME) && !defined(__linux__)
 #define USE_sysctlbyname
 #endif
```
feeley: I prefer to write this with:

```
#ifndef __linux__
...
#endif
```

Issue #427, opened 2019-11-07
master branch parse cc-options wrong


jxy:
```
$ gsc -cc-options '-march=native -O3' test.scm
cc1: error: bad value ('native -O3') for '-march=' switch
```

Similar errors happen with ld-options.

By the way, GAMBCOMP_VERBOSE does not do anything for me.

feeley:
These options have changed recently (as recently as the latest commit).  This was motivated in part by the need to put any characters in the option, including doublequotes, backslashes, etc.  Now you need one `-cc-options` per C compiler option:

```
% gsc -cc-options '-march=native' -cc-options '-O3' -cc-options '-DMSG="hello world"' test.scm
```

Moreover, to get a verbose output you need to set `GAMBUILD_VERBOSE` (because the script that is invoked is now called `gambuild-C`) or pass the option `-verbose` to gsc.

feeley:
By the way, now you can put the cc-options, ld-options, etc as meta-information in the Scheme source file itself, i.e. in test.scm:

```
(##meta-info cc-options "-march-native" "-O3" "-DMSG=\"hello world\"")
...
```

lassik:
Is there an up-to-date list of all environment variables used by Gambit somewhere?

vyzo:
This is a breaking change. 
Why not introduce `-cc-option` for a single option, and maintain the old behaviour for `-cc-options`?

lassik:
Indeed the name `options` becomes slightly misleading.

lassik:
Then again having both `option` and `options` feels a bit cluttered and is potentially confusing.

Maybe add `option`, keep `options` with the old semantics for now and gradually phase out `options` (deprecation warning)?

feeley:
Unfortunately the old `-cc-options` did not have a clean semantics because it did not really pass the options as-is to the C compiler:

```
% gsc -cc-options '-DFOO="hello"' opts.scm
% gsc -cc-options '-DFOO="hello world"' opts.scm
gcc-8: error: world": No such file or directory
*** ERROR IN ##main -- target compilation or link failed while compiling "opts.scm"
```

This prevents using paths with spaces (which are hard to avoid on Windows due to the essential "C:\Program Files" directory).  For example, you can't do `-cc-options '-IC:\Program Files\foo\include'`.

So breaking compatibility with an unclean semantics is justified.

[...]

PR #428, opened 2019-11-07
Fix expression/option ellipses gsi/gsc -help messages

lassik:
* The -e flag can take more than one expression.

* The -cc/ld-options flags were just changed to take only one
  compiler/linker flag per Gambit option (#427).


lassik:
I'll close this PR.

> * The -e flag can take more than one expression.

This is a minor detail.

> * The -cc/ld-options flags were just changed to take only one
>   compiler/linker flag per Gambit option (#427).

They weren't changed after all; see discussion in that PR.

PR #433, opened 2019-11-11
Add --enable-gambitdir=...

feeley:
Add --enable-gambitdir=... to bind ~~ to a specific path, possibly relative to the user's $HOME

### DIFF
--- a/INSTALL.txt
+++ b/INSTALL.txt
@@ -159,7 +159,7 @@ subdirectories can be redirected to other directories using
 the following configure options:
 
   --bindir=DIR            executables (gsi, gsc, ...)
-  --libdir=DIR            libraries (libgambit.a, syntax-case.scm, ...)
+  --libdir=DIR            libraries (libgambit.a, _gambit#.scm, ...)
   --includedir=DIR        C header files (gambit.h, ...)
   --docdir=DIR            documentation (gambit.pdf, gambit.html, ...)
   --infodir=DIR           info documentation (gambit.info, ...)
@@ -224,7 +224,10 @@ The configure options which are specific to the Gambit system are:
                           module whitelist (default is github.com/gambit)
   --enable-ansi-c         link only with ANSI C libraries
   --enable-symlinks       use symbolic links for installed files not in the
-                          central installation directory
+                          Gambit central installation directory
+  --enable-gambitdir=...  set the default binding of ~~ (the central
+                          installation directory) to override the default
+                          which is the directory specified by --prefix
   --enable-interpreter-name=INTERP
                           choose name for Gambit interpreter (default = gsi)
   --enable-compiler-name=COMP
--- a/configure
+++ b/configure
@@ -981,6 +981,7 @@ enable_default_runtime_options
 enable_shared
 enable_ansi_c
 enable_symlinks
+enable_gambitdir
 enable_interpreter_name
 enable_compiler_name
 enable_multiple_versions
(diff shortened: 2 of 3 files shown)


lassik:
I'll rebase my cirrus branch on top of this.

lassik:
How do you give the `$HOME`-relative gambitdir?

feeley:
OK.  I was waiting for the PR to pass CI.  But it should work...

For the Cirrus build artifacts, you could do the following:

```
% ./configure --prefix=$HOME/.gambit_scheme/<hash> --enable-gambitdir=~/.gambit_scheme/<hash> --enable-single-host <etc..>
% make
% make install
% ln -s ~/.gambit_scheme/<hash> ~/.gambit_scheme/current
% tar cf gambit-<hash>.tar ~/.gambit_scheme/<hash>
```

That way, the `~/.gambit_scheme/current` will be a symbolic link pointing to the latest Gambit installed.  The user then simply has to have `~/.gambit_scheme/current/bin` in their shell's PATH to use Gambit, and they don't have to change anything if they "upgrade" to another "tar release".

lassik:
> OK. I was waiting for the PR to pass CI.

I thought Travis is broken at the moment since my last PR failed the Mac build. But maybe it was an intermittent failure.

> For the Cirrus build artifacts, you could do the following:
> 
> ```
> % ./configure --prefix=$HOME/.gambit_scheme/<hash> --enable-gambitdir=~/.gambit_scheme/<hash> --enable-single-host <etc..>
> ```

Is the `~` support the only practical difference between `--prefix` and `--enable-gambitdir`? In other words, would it be better to detect whether the prefix begins with `$HOME`, and if so, remove it for gambitdir? That way a separate configure flag would not be needed.

With the current approach, prefix and gambitdir can be set to completely unrelated values. Would that break something / is there a use case for that?

> ```
> % ln -s ~/.gambit_scheme/<hash> ~/.gambit_scheme/current
> % tar cf gambit-<hash>.tar ~/.gambit_scheme/<hash>
> ```
> 
> That way, the `~/.gambit_scheme/current` will be a symbolic link pointing to the latest Gambit installed. The user then simply has to have `~/.gambit_scheme/current/bin` in their shell's PATH to use Gambit, and they don't have to change anything if they "upgrade" to another "tar release".

That sounds very nice!

feeley:
> > OK. I was waiting for the PR to pass CI.
> 
> I thought Travis is broken at the moment since my last PR failed the Mac build. But maybe it was an intermittent failure.
> 
> > For the Cirrus build artifacts, you could do the following:
> > ```
> > % ./configure --prefix=$HOME/.gambit_scheme/<hash> --enable-gambitdir=~/.gambit_scheme/<hash> --enable-single-host <etc..>
> > ```
> 
> Is the `~` support the only practical difference between `--prefix` and `--enable-gambitdir`? In other words, would it be better to detect whether the prefix begins with `$HOME`, and if so, remove it for gambitdir? That way a separate configure flag would not be needed.
> 
> With the current approach, prefix and gambitdir can be set to completely unrelated values. Would that break something / is there a use case for that?

That's not a bug... its a feature!  The `--prefix` indicates where things are going to be installed by a `make install`, and `--enable-gambitdir=...` indicates the binding of `~~` at run time.  I could imagine wilder things to do with this flexibility, like binding `~~` to a symbolic link at an absolute location in the file system.  I think it is best to not try to be smart about the link between the two things.

lassik:
OK, works for me!

PR #437, opened 2019-11-15
Fix gambuild-C on unix

feeley:
Revert to previous ability to specify multiple options per cc/ld-options and support accessing shell variables

### DIFF
--- a/bin/gambuild-C.unix.in
+++ b/bin/gambuild-C.unix.in
@@ -160,7 +160,7 @@ get_meta_info_list()
 {
   GAMBUILD_META_INFO_LIST=""
   for info in $(get_meta_info "$1" "$2"); do
-    GAMBUILD_META_INFO_LIST="${GAMBUILD_META_INFO_LIST}$(shell_quote "$(scheme_string_decode "${info}")") "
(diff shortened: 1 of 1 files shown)


feeley:
On unix, options such as cc-options, ld-options and ld-options-prelude can now contain multiple options and access shell variables.  This means that care must be taken for the `cc-options` etc parameters of `compile-file` to escape any characters that are special for the shell (unless it is the intent for these characters to be handled specially by the shell, such as for accessing shell variables).  For example, something like this could be used:
```
(define (shell-escape str)
  (append-strings
   (map (lambda (c)
          (cond ((char=? c #\\) "\\\\")
                ((char=? c #\") "\\\"")
                ((char=? c #\') "\\'")
                ((char=? c #\`) "\\`")
                ((char=? c #\$) "\\$")
                ((char=? c #\|) "\\|")
                ((char=? c #\&) "\\&")
                ((char=? c #\;) "\\;")
                ((char=? c #\() "\\(")
                ((char=? c #\)) "\\)")
                ((char=? c #\<) "\\<")
                ((char=? c #\>) "\\>")
                ((char=? c #\ ) "\\ ")
                (else (string c))))
        (string->list str))))
```

PR #440, opened 2019-11-18
Feeley/wsl

feeley:


lassik:
How is WSL accessed by the way - is it already bundled with stock Windows 10 or do you need to run a developer beta?

feeley:
It seems to be bundled with Windows 10 (at leas the VirtualBox image for developers I got free from Microsoft).  Just start a normal Windows Command Prompt and enter `wsl`, and you get a unix shell.

PR #441, opened 2019-11-18
Feeley/fix srfi 132

feeley:


feeley:
The Windows Visual Studio C compiler (cl.exe) gave a compile time error for `___F64DIV(1.0,0.0)` which was generated because the constant folding logic did not constant fold `(fl/ 1.0 0.0)` to `+inf.0`.  I don't know why there was a constraint on the divisor to be non zero for `fl/`.  It is logical for `/` but not for `fl/` where division by zero is meaningful.

PR #443, opened 2019-11-19
Fix issues with delete-directory on Windows, and misbehaving hypot on…

feeley:
… MinGW

### DIFF
--- a/include/gambit.h.in
+++ b/include/gambit.h.in
@@ -1644,7 +1644,10 @@
 #define ___HAVE_GOOD_TANH
 #define ___HAVE_GOOD_POW
 #define ___HAVE_GOOD_ATAN2
+
+#ifndef __MINGW32__
 #define ___HAVE_GOOD_HYPOT
+#endif
 
 #endif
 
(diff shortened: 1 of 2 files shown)


lassik:
From the commit:

> On Windows, a call to RemoveDirectory doesn't immediately delete the directory. Instead the directory is marked for deletion, so it is necessary to busy-wait for the OS to actually delete it. Unfortunately, this introduces a race condition where some other process may be creating a file or directory with the same name. A better solution would be to move the directory to a safe place with a unique name and then call RemoveDirectory on it.

MSDN says:

> The RemoveDirectory function marks a directory for deletion on close. Therefore, the directory is not removed until the last handle to the directory is closed.

So the issue is whether or not some process has the directory open. I don't think it's possible to move a file/directory when there are open handles to it, so moving the directory to a safe place may not be possible either.

lassik:
To be clear, on Unix `rename()` works despite open handles, but on Windows it doesn't IIRC.

feeley:
This idea comes from a youtube video of a Windows developer saying this is the only correct implementation (moving then deleting the directory) and he has implemented it that way.  See https://www.youtube.com/watch?v=uhRWMGBjlO8 at 7:40.

lassik:
Thanks, very interesting. Apparently it's possible to open files using a special mode that doesn't forbid others from moving their directory entries from underneath them (i.e. mimic Unix). I didn't understand from the video whether that's the only situation due to which the presenter is doing that temp rename thing, but it's surely better to follow his advice than mine.

PR #461, opened 2019-12-29
Match keywords

alvatar:
Just a small tweak to support keywords in `match`


alvatar:
Thanks for the quick review!

PR #465, opened 2020-01-11
Fix renameat2 missing in glibc

feeley:


feeley:
Fixes issue with rename-file on linux introduced by commit #462

PR #482, opened 2020-02-08
Avoid busy waiting in process-status

feeley:


vyzo:
what are event condvars?

feeley:
"Event condvars" are added to the previous reading and writing condvars.  They indicate that some event is received on a device not related to reading or writing becoming possible.  They are currently only used for process ports to indicate that the child process has terminated (making it possible to avoid a busy wait of the process termination in the implementation of process-status).  In principle an event condvar could be used for other devices needing some events on a side channel.

PR #514, opened 2020-03-27
Universal backend: add flhypot and fix fltruncate, unsafe exact?/inexact?, cleanup of _univlib.scm

feeley:


tests/unit-tests/03-number/_num.scm
```diff
@@ -777,36 +777,36 @@
...
-(let ((arg 1.)) (check-eq? arg (+ arg 0)))
-(let ((arg 1.)) (check-eq? arg (/ arg 1)))
-(let ((arg 1.)) (check-eq? arg (- arg 0)))
-(let ((arg 1+i)) (check-eq? arg (* 1 arg)))
-(let ((arg 1+i)) (check-eq? arg (* arg 1)))
-(let ((arg 1+i)) (check-eq? arg (+ 0 arg)))
-(let ((arg 1+i)) (check-eq? arg (+ arg 0)))
-(let ((arg 1+i)) (check-eq? arg (/ arg 1)))
-(let ((arg 1+i)) (check-eq? arg (- arg 0)))
+(let ((arg 1)) (check-equal? arg (* 1 arg)))
```
gambiteer: The code in `_num.scm` is written to preserve object identity in these cases, that's why I used `check-eq?`.  Is there a reason I don't know why your changing them?

gsc/_t-univ-4.scm
```diff
@@ -3386,20 +3389,55 @@
...
+                 (^cast* 'bigdigit
+                         (^bitand
+                          (^parens
+                           (^bitand (^parens
+                                     (^bitnot
+                                      (^array-index (^bignum-digits arg1)
+                                                    (^fixnum-unbox arg2))))
+                                    (^array-index (^bignum-digits arg3)
+                                                  (^fixnum-unbox arg4))))
+                          (^int univ-mdigit-base-minus-1))))
```
gambiteer: Here you're using that the mdigits and adigits are the same in the universal backend, and using `univ-mdigit-base-minus-1` in an `adigit` routine.   `univ-adigit-base-minus-1` might be a more appropriate variable name.

tests/unit-tests/03-number/_num.scm
feeley: There's no reason to require of these operations to preserve object identity.  What is required is that the value and exactness of arg is preserved.  When gsc compiles this code the expansion of these operations would have to be overly complex to preserve object identity.

gsc/_t-univ-4.scm
```diff
@@ -1068,7 +1068,9 @@
      (return
       (^flonum-box
        (if arg2
-           (^float targ-inexact-+0) ;; TODO: implement 2 argument fllog
+           (univ-ieee/ ctx
```
gambiteer: `_num.scm` has special code for when both arguments are positive or negative powers of 2.  I don't know whether it's important that the universal backend share that behavior.

gsc/_t-univ-4.scm
feeley: The universal backend uses the same width for mdigits and adigits, so either is OK.

gambiteer:
> On Mar 27, 2020, at 5:15 PM, Marc Feeley <notifications@github.com> wrote:
> 
> The universal backend uses the same width for mdigits and adigits, so either is OK.
> 
> 

Yes, but it is a good idea to bake this into these routines?  I’m just asking.

Brad

gsc/_t-univ-4.scm
feeley: When the 2 argument `##fllog` is inlined by the C backend, it calls the following C macro

```
#define ___F64LOG2(x,y)(___CLIBEXT(log)(x)/___CLIBEXT(log)(y))
```

so the definition in _num.scm is not used (so the universal backend is doing the same here and just dividing the logs).

gsc/_t-univ-4.scm
feeley: You are probably right... however this is nitpicking when there's a mountain of other stuff to resolve...

gambiteer:
> On Mar 27, 2020, at 5:13 PM, Marc Feeley <notifications@github.com> wrote:
> 
> There's no reason to require of these operations to preserve object identity. What is required is that the value and exactness of arg is preserved. When gsc compiles this code the expansion of these operations would have to be overly complex to preserve object identity.
> 
> 

OK, but as of yesterday, all these tests passed, as these are generic operations that go through the _num.scm library.

I suggest we leave them so if suddenly they don’t work we can notice, figure out why, and maybe then fix the library code or change the tests to use equal.

Brad

feeley:
They did not pass when compiled... i.e. `./run-unit-tests.scm -C -gsc`

gambiteer:
> On Mar 27, 2020, at 5:56 PM, Marc Feeley <notifications@github.com> wrote:
> 
> They did not pass when compiled... i.e. ./run-unit-tests.scm -C -gsc
> 
> 
OK, so what are we trying to test here?

The library?

The compiler output?  (I thought there were separate tests for the compiler.)

I put in a lot of tests to get as much test coverage of _num.scm as I could, and to check that certain properties that I expected the library to satisfy are satisfied.

Where should such tests of the library (not the compiler) go?

Brad

feeley:
We want the compiler and interpreter to give the correct result and the same result (as much as possible).

A test for object identity is overly restrictive.  It assumes a specific implementation.  We want the unit test to determine if the property we care about (preservation of exactness and value) holds.  We don't care about object identity (in this case).  Preserving object identity is a valid implementation but it is not the only valid implementation.

Do you have a motivation for preserving object identity?

BTW, the _num.scm test should be split in many smaller tests so that it is easier to determine which procedure is causing a problem.  It also makes it faster to execute the tests (when compiled).

gambiteer:
On Mar 27, 2020, at 7:23 PM, Marc Feeley <notifications@github.com> wrote:
> 
> We want the compiler and interpreter to give the correct result and the same result (as much as possible).
> 
> 

OK, I’ll investigate whether there are a set of declarations that will give the same result as the interpreter for these tests.

Like I said (well, implied, actually), if this is a library test, it should call the library routines and not have the tests basically compiled out.

Brad

tests/unit-tests/03-number/_num.scm
gambiteer: It appears that if you add
```
(declare (not inline-primitives))
```
to the top of
```
tests/unit-tests/#.scm
```
then all the original tests in `_num.scm` using `check-eq?` pass with
```
firefly:~/programs/gambit/gambit/tests> ./run-unit-tests.scm -C -gsc
```
I really think of the tests in `unit-tests` as runtime library tests,  not compiler tests, and I think we should make this change to `#.scm` (and put the old tests back, too).

feeley: I have nothing against the idea of testing the library, and the `(declare (not inline-primitives))` is fine for that.  But even for the library the preservation of object identity should not be a requirement of the test.  I want the test to represent what we expect from the library procedure.  I want to be able to change the implementation of the library and not have to change the unit tests.  So only the properties that are required should be tested and nothing else, otherwise we would end up having tests that depend on a specific implementation that is overly constraining.

Independently, the exact same tests should be run with `(declare (inline-primitives))` (as is the case now) so that the compiler's inlining can be tested too.  This is true for all tests, so maybe the `run-unit-tests.scm` script should have an option to select which declaration should be used.

BTW, these double tests are a good motivation for testing what is expected of the procedures (so that they work whether they end up being inlined or an actual call to the library).

PR #517, opened 2020-04-13
Add ##readtable-foreign-write-handler-register!

feeley:


lib/_io#.scm
```diff
@@ -1145,6 +1145,7 @@
   (here-strings-allowed?          unprintable: read-write:)
   (dot-at-head-of-list-allowed?   unprintable: read-write:)
   (comment-handler                unprintable: read-write:)
+  (foreign-write-handler-table    unprintable: read-write:)
```
gambiteer, line 1148: Does this require new documentation?

feeley: It is not exported, so not for now...  It exists for customizing the writing of foreign objects, which usually are written `#<type #serial-num address>`.  For specific foreign objects it is nice to use a more informative format.  For example:

PR #519, opened 2020-04-14
Add registry of display-exception handlers

feeley:


lib/_repl.scm
```diff
@@ -3535,7 +3535,7 @@
           (##write arg-num port)
           (##write-string ") " port))))
 
-  (define-prim (display-exception exc)
+  (define-prim (display-known-exception exc)
```
vyzo, line 3538: so is `display-exception` now gone? that's a breaking change, can we keep it around?

PR #525, opened 2020-04-22
Explicitly state the default for truncate in bidirectional ports

abdelq:
Minor change to the documentation to specify bidirectional ports too for the `truncate:` option. I had to look in the source code to make sure it was the default I was expecting.


feeley:
Thanks!

PR #526, opened 2020-04-22
Including EACCES and permission-denied-exception

abdelq:
It's a fairly common exception and should be supported in my opinion.

While adding it, I took the liberty of fixing a few things. However, I'm slightly unsure if I did the right thing in `lib/_io.scm`.

~This PR is only missing documentation.~


feeley:
Thanks! As far as I can tell you changed all the places that needed to change... and the handling for Windows too!

PR #530, opened 2020-04-25
fix _http module

FredericHamel:


lib/_http/test.scm
```diff
@@ -185,31 +185,35 @@
   (define result-identity1 (http-get (string-append base-url "/foo bar") #f #t))
   (define result-identity2 (http-get (string-append base-url "/foo%20bar") #f #f))
 
-  (check-equal? (u8vector->string
+  (test-equal (u8vector->string
```
feeley, line 188: Note that the `test-equal` form takes the expected result as the first parameter.  The `check-equal?` form took the expected result as the second parameter.  Also, the test description is now an optional parameter that comes before the expected result.

PR #532, opened 2020-04-25
DRAFT: Resolve file and module paths directly in compilation APIs

daviwil:
This change updates Gambit's compiler API (`compile-file` and `compile-file-to-target`) to enable modules to be compiled in the same way as normal files.  This (hopefully) removes the need for special-case behavior for module compilation in the codebase.

After looking over the code for a while, it occurred to me that the only real difference between compiling a module and a file is in determining the right input file and target directory.  This change adds a new procedure, `##with-resolved-input-file`, which is inserted into both compiler procedures to
resolve the input file and target directory before continuing with existing behavior.

This procedure will attempt to parse the input file as a module reference and, if found to be a module, will resolve the module against the search path.  The module's full file path will be given as the input file and a temporary directory based on the generated module output directory will be created and
returned to the compilation function.  Once the compilation procedure is complete, the temporary directory will be renamed back to the real target directory and the corrected output file path will be returned.

Because these changes are being added into the compiler procedures, they enable the expected behavior when running the various `gsc` formulations:

> These examples are run with `keep-temp` to show the output files

#### `gsc -c`

```
λ gsc -c . test-lib/a test-lib/b

λ tree test-lib/
test-lib/
├── a.sld
├── a@gambit409003@C
│   └── a.c
├── b.sld
└── b@gambit409003@C
    └── b.c
```

#### `gsc -exe`

```
λ gsc -keep-temp -exe -o lib-test . test-lib/a test-lib/b lib-test.scm

λ tree .
.
├── lib-test
├── lib-test.c
├── lib-test.o
├── lib-test.scm
├── lib-test_.c
├── lib-test_.o
└── test-lib
    ├── a.sld
    ├── a@gambit409003@C
    │   ├── a.c
    │   └── a.o
    ├── b.sld
    └── b@gambit409003@C
        ├── b.c
        └── b.o
```

#### `gsc -obj`

```
λ gsc -keep-temp -obj . test-lib/a test-lib/b

λ tree test-lib/
test-lib/
├── a.c
├── a.sld
├── a@gambit409003@C
│   └── a.o
├── b.c
├── b.sld
└── b@gambit409003@C
    └── b.o
```

**NOTE**: The `.c` files were not placed in the module target path because `compile-file` is resolving the target file path differently than `compile-file-to-target`.  I think it will be easy to fix but I decided to wait until you've taken a look at the overall approach before I spend time on it.

#### `gsc -dyn`

```
λ gsc -keep-temp . test-lib/a test-lib/b
test-lib/a:
test-lib/b:

λ tree test-lib/
test-lib/
├── a.c
├── a.sld
├── a@gambit409003@C
│   └── a.o1
├── b.c
├── b.sld
└── b@gambit409003@C
    └── b.o1
```

**NOTE**: The `-dyn` parameter output is also affected by the `compile-file` issue mentioned above.

#### `gsc -link`

```
λ gsc -keep-temp -link . test-lib/a test-lib/b
test-lib/a:
test-lib/b:

λ tree test-lib/
test-lib/
├── a.sld
├── a@gambit409003@C
│   └── a.c
├── b.sld
└── b@gambit409003@C
    ├── b.c
    └── b_.c
```

I've added some comments in the code marked with **`REVIEW`** where I think some discussion might be needed.


gsc/_gsclib.scm
```diff
@@ -159,82 +200,83 @@
...
-                               pkg-cfg
-                               pkg-cfg-path)))))))
+      (##with-resolved-input-file
+       filename
+       (if (##eq? options (macro-absent-obj))
+           '()
+           (macro-force-vars (options)
+             options))
+       (lambda (filename target-dir opts)
+         (let* ((out
```
daviwil, line 210: The changes here are mostly indentation to account for wrapping everything in a lambda.

feeley:
I think the temp files should go in the same build directory.  So when compiling a module it goes in a build directory with a suffix giving the build context (e.g. `mod@gambit409003@C`), and when compiling a file it goes in the same directory as the source unless a directory was specified with the `-o` command line option.

With the `-c` option the generated artifact is just `mod.c` (or `mod.js`, `mod.py`, etc).

With the `-dyn` option the generated artifact is `mod.o1` (or the next available `.oN`) but the input is `mod.c` if it exists, otherwise `mod.c` is first generated as per the previous case.

With the `-obj` option the situation is identical to `-dyn` except the generated artifact is `mod.o` (or `mod.obj` on Windows).  Note that `mod.o1` can't be generated from `mod.obj` because the C compilation options are slightly different (presence of `-D___DYNAMIC`).

With the `-link` option the generated artifact is `mod_.c` from `mod.c` and all the dependent modules.  So this may require building those modules (with the equivalent of the `-c` option) to get their `.c` file.

With the `-exe` option there are multiple generated artifacts: `mod_.c` (as generated by a `-link`), `mod.o` and `mod_.o` (as generated by a `-obj`), and the `.o` of dependent modules.

feeley:
For the atomic creation of the artifacts, the only reliable way I know is to create a temporary directory, create the artifact in that directory, and then rename the file, and then delete the directory.

daviwil:
I am creating a temporary directory during the compilation phase here using the generated module output folder (`mod@gambit409003@C`) as a prefix:

https://github.com/gambit/gambit/pull/532/files#diff-e891b74a2c1e49cfba925975f24c9588R52

The compilation output of individual .scm files does not get placed into a temporary subfolder though, it retains the original behavior.  The main differences in this change are that:

- Module refs are now resolved from on the search path
- Module compilation output is now placed in subfolders (and temporary folders are used until the compilation process is complete so that partially-finished outputs aren't picked up by other running instances of `gsc`, as you mentioned)

feeley:
Could you explain with a list of the steps (which directories and files are created and in what order)?

You shouldn't use `mod@gambit409003@C` as a temporary directory prefix because that format is reserved for contextual information, i.e. `mod@<context>` and I wouldn't want the temporary directory suffix to be confused with contextual information.  Just using `mod` as a prefix should do.  Perhaps the temporary directory creation algorithm should choose a special naming scheme that is clear and doesn't clash with other conventions.

daviwil:
Sure!  Here's what's happening:

1. User runs `gsc -c src/ mod`
1. `compile-file-to-target` gets invoked on `mod`
1. `compile-file-to-target` calls `##with-resolved-input-file` which then uses `##parse-module-ref` to determine `mod` is a module reference
1. `##search-module` is used to resolve `mod` to the path `src/mod/mod.sld`
1. `##module-build-subdir-path` is used to generate the subfolder path `src/mod/mod@gambit409003@C` which is then passed into `##create-temporary-directory`, resulting in a temporary folder path of `src/mod/mod@gambit409003@C9204`
1. The full path to the module source file and generated target folder are passed along to the `callback` given to `##with-resolved-input-file` resulting in the rest of `compile-file-to-target`'s behavior being executed
1. The `mod.c` file is written out into the `src/mod/mod@gambit409003@C9204` subfolder and the output file path is then returned from the callback to `##with-resolved-input-file`
1. The temporary folder is renamed to `src/mod/mod@gambit409003@C` and the output file path for `mod.c` is updated to reflect the final folder path, then returned

(Note that the temporary folder creation only occurs when we determine that the input is a module reference; the expected behavior for compiling individual .scm files is retained)

Based on what you've written in previous comments, I think I understand now that I need to make the following changes:

- The temporary folder path created in step 5 should be named to something like `mod-compile-tmp-9204` so that it isn't confused with a real module output folder
- In step 8, I should check for an existing module output folder `mod@gambit409003@C`, and if it exists, compare the contents of any `.c` files to see if they are exactly the same before copying them to the final output folder so that timestamps are not affected.

Was there anything else I might have missed in what you explained above?

feeley:
Basically yes. I wonder if in addition when going from `mod.sld` to `mod@.../mod.c` there's a need to check if `mod.c` is more recent than the last modification to `mod.sld` and any dependent module's source file (which may have changes in the macro definitions) to avoid compiling `mod.sld` to `mod.c` again.  But this does not seem correct because Scheme macros may cause the resulting expansion to depend on many things (current time, content of a data file, output of call to a subprocess, user input, etc).  So compiling again is unavoidable in general.  So all of this makes me wonder if it is possible to limit the extent to which the dependent modules must be recompiled because the previous observation seems to indicate that all modules must be recompiled in general. That's bad.

daviwil:
I can certainly check to see if the `.sld` is newer than the `.c` but I'd probably also have to check any `include[-ci]` directives in the `define-library` plus any other loose `include` (I've seen an `include` used _inside_ of the `define-library`'s `begin` in another repo).  And definitely, checking for updated input files is much more complicated when macro expansion comes into the picture.

Provided that we can check the timestamps of files in a module's dependency graph plus any files that are `include`d (perhaps with some static analysis), one possibility is that we could add an extra argument to `gsc` to indicate which modules should always be recompiled regardless of timestamp and we could use that information to figure out which parts of the dependency tree to recompile.

That part of the work would probably come in a separate PR though because this PR only deals with ensuring that modules given as inputs to `gsc` get resolved correctly with respect to the search path.  Do you think that the timestamp checking of input files can be left for the whole-program compilation work?  I'll still check the timestamp and contents of target files before moving them to their final location.

feeley:
I agree that for now we should keep things simple (and in a way broken) until we can find a more robust solution (probably using some of the infrastructure of the whole-program compilation mode).

daviwil:
Planning to get an update to this sent out today (or latest tomorrow morning), here is my list of things to do:

- [ ] Use a different naming scheme for the temporary folder path
- [ ] Check for existing target output to see whether it needs to be updated after compilation
- [ ] Figure out why `##with-resolved-input-file` is causing the compiler to segfault when building `_repl`.  This happened in CI and I also saw it happening in my own local bootstrap builds

daviwil:
Based on what I've learned so far, it looks like this PR is blocked because the changes require procedures that won't be available until the next public Gambit release.  I've considered an alternate solution of putting the changes in `gsi/main.scm` but even that will still exhibit the same issue because it would change the `-c` code path to use `##parse-module-ref`.

I can see 2 options for moving forward:

- Wait to consider merging this PR until after the next Gambit release.  Not ideal because it'd be nice to have this behavior in the next major release.  In the meantime users will have to compile modules separately with `-dyn` and include the object files in their program builds (which means hardcoding module build output paths in their `gsc` call).

- Update Gambit CI to run an extra `make bootstrap` after the initial build so that changes requiring procedures in `master` can be used to build Gambit itself.  This will mimic what the build will do after the next public release.  The downside is that it requires changing CI, but I have some thoughts on how to make that easier.

What do you think @feeley?

PR #534, opened 2020-04-26
Add second parameter to reverse and reverse!, the tail of the result

feeley:


lassik:
SRFI 1 has `append-reverse` and `append-reverse!` that do the same job, if I understood correctly. https://srfi.schemers.org/srfi-1/srfi-1.html#append-reverse If so, might be worth using those names; SRFI 1 is almost ubiquitous.

feeley:
Indeed! Will rewrite.

feeley:
Adding a tail to reverse is redundant when the SRFI 1 append-reverse is available, so that is a better way to go.  Withdrawing this PR.

PR #537, opened 2020-04-28
improve build of modules and add list-modules

FredericHamel:


feeley:
Could you also add a `test-modules` target that runs each module's test?

PR #541, opened 2020-04-30
Add define-procedure, define-primitive, define-prim&proc, primitive, and declare-safe-define-procedure special forms to simplify writing procedures that correctly process their parameters (auto-forcing, type checking, tail calling when exceptions raised)

feeley:


gambiteer:
Will you give some examples about how to use these new forms?

feeley:
This commit adds a few special forms to simplify defining procedures that process their parameters in a consistent way: `define-procedure`,`define-primitive`,`define-prim&proc`,`primitive`, and `declare-safe-define-procedure`.  The type checks, auto-forcing and optional parameter processing are automatically inferred by the type and default values that are declared in the parameter list.  Here is a representative example of `define-procedure`:

```
(define-procedure (string->utf8
                   (str   string)
                   (start (index-range-incl 0 (string-length str))
                          0)
                   (end   (index-range-incl start (string-length str))
                          (string-length str)))
  (primitive (string->utf8 str start end))) ;; this calls ##string->utf8
```

The definition `(define-primitive (foo ...) ...)` is similar to `(define-procedure (foo ...) ...)` but it defines `##foo`.  In other words, "primitives" are always in the `##` namespace.  This means the `##` prefix can be completely avoided in the common case.

A few of the new special forms have been used at the end of `lib/gambit/list/list.scm` to define the list sorting procedures.  By default the body of the `define-XXX` forms assume that the calls are to primitives (i.e. `(car ...)` is really the call `(##car ...)`) and that these calls are unsafe (without type checking). The commented call `(declare-safe-define-procedure #t)` can be uncommented to enable these calls to include their type checking, which is useful when debugging.

The `define-prim&proc` special form generate a `define-primitive` and `define-procedure` from the same definition, which is useful when defining primitives that are inlined by the compiler and also exported.  Here are a few examples:

```
(define-prim&proc (cons (car object) (cdr object)))
(define-prim&proc (car (pair pair)))
(define-prim&proc (cdr (pair pair)))
(define-prim&proc (set-car! (pair pair) (obj object)))
(define-prim&proc (set-cdr! (pair pair) (obj object)))
```

feeley:
Here is a list of the available types specifiers for a parameter `x` and the associated constraint on the value of `x`.  Note that there is no direct way for specifying the type "list" or "list of elements of type T".  For lists, the procedure body must iterate on the pairs, going from cdr to cdr until a non-pair is found and then do a check for the empty list with `(macro-check-list lst <arg-id> (<procedure-name> <args>...) <body>)` to check that the parameter is a proper list (i.e. that it ends with the empty list).

```
BASIC TYPES (other than numbers):

boolean                    x is a boolean
char                       x is a character
pair                       x is a pair
procedure                  x is a procedure
string                     x is a string
symbol                     x is a symbol
vector                     x is a vector

NUMBERS:

number                     x is a number (possibly complex, rational, etc)

real                       x is a real number (any number except complex)

fixnum                     x is a fixnum and -2^(64-3) <= x <= 2^(64-3) - 1
(fixnum-range lo hi)       x is a fixnum and lo <= x < hi
(fixnum-range-incl lo hi)  x is a fixnum and lo <= x <= hi

index                      x is a fixnum and 0 <= x <= 2^(64-3) - 1
(index-range lo hi)        x is a fixnum and lo <= x < hi
(index-range-incl lo hi)   x is a fixnum and lo <= x <= hi

exact-signed-int8          x is an exact integer, -128 <= x <= 127
exact-signed-int16         x is an exact integer n, -32768 <= x <= 32767
exact-signed-int32         x is an exact integer n, -2^31 <= x <= 2^31 - 1
exact-signed-int64         x is an exact integer n, -2^63 <= x <= 2^63 - 1
exact-unsigned-int8        x is an exact integer n, 0 <= x <= 255
exact-unsigned-int16       x is an exact integer n, 0 <= x <= 65535
exact-unsigned-int32       x is an exact integer n, 0 <= x <= 2^32 - 1
exact-unsigned-int64       x is an exact integer n, 0 <= x <= 2^64 - 1

flonum                     x is a flonum (floating point number)
inexact-real               x is a flonum
inexact-real-list          x is a flonum, exception mentions LIST

TIME TYPES:

time                       x is a time object
absrel-time                x is a real or a time object
absrel-time-or-false       x is #f or a real or a time object

PORTS:

port                       x is a port (input, output, or input-output)
input-port                 x is an input port
output-port                x is an output port

object-input-port          x is an object input port
object-output-port         x is an object output port
vector-input-port          x is a vector input port
vector-output-port         x is a vector output port

character-input-port       x is a character input port
character-output-port      x is a character output port
string-input-port          x is a string input port
string-output-port         x is a string output port

byte-port                  x is a byte port (input, output, or input-output)
byte-input-port            x is a byte input port
byte-output-port           x is a byte output port
u8vector-input-port        x is a u8vector input port
u8vector-output-port       x is a u8vector output port

device-input-port          x is a device intput port
device-output-port         x is a device output port
process-port               x is a process port

tcp-client-port            x is a tcp-client port
tcp-server-port            x is a tcp-server port

udp-port                   x is a udp port
udp-input-port             x is a udp input port
udp-output-port            x is a udp output port

tty-port                   x is a tty port

LIST AND VECTOR VARIANTS OF ABOVE:

list                       x is '(), exception mentions LIST
char-list                  x is a character, exception mentions LIST
char-vector                x is a character, exception mentions VECTOR
pair-list                  x is a pair, exception mentions LIST
exact-signed-int8-list     x is an exact-signed-int8, exception mentions LIST
exact-signed-int16-list    x is an exact-signed-int16, exception mentions LIST
exact-signed-int32-list    x is an exact-signed-int32, exception mentions LIST
exact-signed-int64-list    x is an exact-signed-int64, exception mentions LIST
exact-unsigned-int8-list   x is an exact-unsigned-int8, exception mentions LIST
exact-unsigned-int16-list  x is an exact-unsigned-int16, exception mentions LIST
exact-unsigned-int32-list  x is an exact-unsigned-int32, exception mentions LIST
exact-unsigned-int64-list  x is an exact-unsigned-int64, exception mentions LIST

GAMBIT TYPES:

error-exception            x is an error-exception object
box                        x is a box
condvar                    x is a condition variable
f32vector                  x is a f32vector
f64vector                  x is a f64vector
foreign                    x is a foreign object
keyword                    x is a keyword
mutex                      x is a mutex
processor                  x is a processor object
s16vector                  x is a s16vector
s32vector                  x is a s32vector
s64vector                  x is a s64vector
s8vector                   x is a s8vector
table                      x is a table
tgroup                     x is a thread group
thread                     x is a thread
u16vector                  x is a u16vector
u32vector                  x is a u32vector
u64vector                  x is a u64vector
u8vector                   x is a u8vector
will                       x is a will
continuation               x is a continuation object
random-source              x is a random-source object
readtable                  x is a readtable
type                       x is a structure type descriptor
mutable                    x is a mutable object

OTHERS:

initialized-thread
not-initialized-thread
not-started-thread
not-started-thread-given-initialized
string-or-ip-address
string-or-nonnegative-fixnum
```

lassik:
Is `pair` a mutable pair?

lassik:
How should we deal with primitives like this:

```scheme
(define-prim (null? obj)
  (macro-force-vars (obj)
    (##null? obj)))
```

Would an `object` vs `lazy-object` distinction make sense after all?

lassik:
In general, list checks are complicated by the fact that many procedures take a proper list, but if we check a list against cycles before handing it over to the procedure body, the body will traverse the list a second time. This is unacceptably inefficient. Do you have an idiom in mind for this?

lassik:
I gave up on converting SRFI 1 for now since almost all primitives there run into special cases like the above. In a sense, it's good that we started with it so we need to think about this stuff early.

I guess something like `(gambit string)` would be more straightforward since they cannot contain nested objects and cannot be circular.

feeley:
It should be

```
(define-prim&proc (null? (obj strict-object)))
```

because we want any promises passed to `null?` to be forced automatically.  As I said, promises should be transparent (except for the `promise?` procedure).

Yes the `pair` type only checks for a pair.  There's `(macro-check-mutable x ...)` to verify that `x` is mutable.  So in fact my example with `set-car!` was not complete as it should in addition do a mutability test in the body.

Checking for list parameters has to be done in the logic of the procedure.  The list is traversed with a `probe`: while `(pair? probe)` move to `(cdr probe)` and when no longer a pair do a `(macro-check-list probe ...)` to verify that `probe` is the empty list (which detects improper lists). There are also `(macro-check-char-list c ...)` and friends to check if `c` is a character and otherwise raise an exception saying that the parameter was not a list of characters.

Will you try with `(gambit string)`?

PR #542, opened 2020-04-30
update ifdef in os_io.c

FredericHamel:


daviwil:
Interesting, what does this fix?

FredericHamel:
It avoid having the procedure `sigchld_signal_handler` to be use in the function `___setup_child_interrupt_handling` while it is not defined.

This happen when the macro `USE_POSIX` is not defined and and the macro`USE_SIGNALS` is defined.

PR #550, opened 2020-05-04
Add -:iL runtime option to use LC_CTYPE and LC_ALL to determine default char-encoding and remove "file" and "stdio" I/O settings

feeley:


lib/main.c
```diff
@@ -308,6 +335,16 @@ char *option;)
 }
 
 
+___HIDDEN ___UCS_2 lc_ctype_env_var[] =
+{'L', 'C', '_', 'C', 'T', 'Y', 'P', 'E', '\0'};
```
gambiteer, line 339: I don't really know what this should be, but do you want two instances of 'C' in this list?

feeley: yes... it is the string "LC_CTYPE"... https://www.gnu.org/software/libc/manual/html_node/Locale-Categories.html

PR #558, opened 2020-05-07
gitignore more things to prevent inclusion in commit history

jcmdln:
When making adjustments in my copy of the repo I noticed there were
some seemingly missing items from the gitignore.  I appreciate that
the newly created files were visible so I could remove them, though
found them to create impedance between stashing some commits.

This is highly subjective, but anything created after running
./configure or make should be ignored based on the attached commit.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -6,6 +6,14 @@
 bin/gambcomp*
 !bin/gambcomp*.in
 
+# gambuild
+bin/gambuild*
+!bin/gambuild*.in
+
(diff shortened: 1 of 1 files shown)


lassik:
LGTM.

> anything created after running ./configure or make should be ignored based on the attached commit.

Agreed.

daviwil:
This will definitely help make my `magit-status` panel less noisy!  @feeley what do you think?

PR #568, opened 2020-05-25
Add GitHub Actions workflow for CI runs

daviwil:
This change adds an initial GitHub Actions workflow called "Gambit - CI" to perform builds for CI on `master` and pull requests.  The aim here is to use a single CI provider to build for all 3 OS platforms as discussed in https://github.com/udem-dlteam/libs/issues/4.  I've tried to replicate what is already being done in `appveyor.yml` and `.travis.yml` with some caveats listed in the Limitations section below.

You can check out the latest run on my fork to see what it looks like: https://github.com/daviwil/gambit/runs/706099926

**Current Issues**
- The `make check` run on MingW64 is failing on one of the tests.  I'm not sure if this is a legitimate test failure or a result of the segfault issues on MingW64 that we've been discussing recently.  I'll see if I can diagnose the issue locally.
- `misc\vstudio.bat` fails to build (despite the runner thinking it succeeded).  @FredericHamel is there anything else I need to do to make sure Gambit can build with this script?

**Limitations**
- Gitter does not (yet?) support receiving build webhooks from GitHub Actions so we won't get nice activity feed entries in the channel when builds pass or fail.
- Cygwin is not baked into the Windows image.  I've tried installing Cygwin via Chocolatey but it will take some effort to get all of the right programs installed (gcc, make, etc) so that Gambit can be built.  We *could* keep around the old `appveyor.yml` just for Cygwin builds.
- I'm currently not trying to build against the older MingW as the AppVeyor configuration does but I could give it a shot if necessary.  Like the Cygwin build, it'll require some extra effort to get the environment set up correctly.  Could also use the existing `appveyor.yml` for that "legacy" build.

**Future Work**
- Produce usable artifacts for all platforms that can be downloaded for testing purposes
- Make it easy to download these artifacts in CI runs of Gambit module repos so that they can be used for verification against the state of `master`.

/cc @feeley @lassik


feeley:
That's a great feature that will help streamline Gambit development!

I'm surprised there is no build webhook for gitter.  Isn't it just a matter of adding the hook to the gitter configuration?

I'm also experiencing random segfaults when building locally with MingW64.  That will have to be solved because 32 bit environments are progressively less in use and MingW64 is probably the most popular "go to" environment on Windows these days.  I don't know how useful it is to test cygwin (probably good to test it but a low priority).

Although it doesn't have to be done for the first integration, it will be nice to use our own servers for running the builds.  We have quite a few more or less idle machines that could be used to have speedy turnarounds.  Do you know what has to be done to set this up?

daviwil:
I think that Gitter just hasn't added native support for the "check_run" set of webhook events that GitHub sends when Actions workflows run.  It does seem possible to add a "custom" integration where I can send webhooks manually, I'll experiment with that to see if I can get something wired up on a test channel.

I'd love a resolution to the segfaulting issues in MingW64, it's been painful for me lately.  I'm pretty sure it's something that GCC's `-O1` optimization pass is doing, @FredericHamel and I were discussing that a couple of days ago in Gitter.

You can definitely set up "self-hosted runners" on the machines you already have, the documentation for doing so looks pretty good: https://help.github.com/en/actions/hosting-your-own-runners/about-self-hosted-runners.

FredericHamel:
The script vstudio.bat works on Gambit `make dist` not on a fresh clone of Gambit.
The `make dist` can be create on any platform.

daviwil:
Ahhh, OK, I think I see how it was working before on AppVeyor, the msvc build would complete first and then `vstudio.bat` would be run with the already-built Gambit compiler.  Thanks @FredericHamel!

daviwil:
First green build on Actions!  https://github.com/daviwil/gambit/runs/707001571?check_suite_focus=true

Turns out that it was the MinGW32 build that had issues running tests, which appears to be known because I see it being disabled in the `appveyor.yml` too.  I've skipped the test run for that architecture.  Also fixed the run of `vstudio.bat`, that also seems to be working fine now.

Last step will be to try and enable a custom webhook for Gitter, might be able to get to that today or tomorrow.

feeley:
Great! And we should also change the pass/fail badges in the README.md file (perhaps have one badge per platform, or something like that)

daviwil:
I can certainly take care of that. Should I also remove the unneeded `.travis.yml` and the parts of `appveyor.yml` that have been migrated, or keep those around for now?

feeley:
You can remove those files... if we ever need them again (very unlikely) we know where to find them...

daviwil:
Sounds good!

feeley:
Is this ready for master?

daviwil:
Not yet, the Gitter job isn't working correctly. Should have it resolved in a bit, will let you know.

feeley:
OK.  By the way, another advantage of running all of this on our local servers is that we could easily run our standard benchmarks on known machines with results browsable on the Gambit forensics web app: http://udem-dlteam.github.io/gambit-forensics/ .  That too could be integrated to gitter so that any performance anomaly would show up in a noticeable way.

daviwil:
Wow, that's a cool site!  Yep, that sounds like a great use case for self-hosted runners.  I'll be happy to help however I can in getting that set up once this PR is merged.  Will try to get the Gitter integration wrapped up now!

lassik:
Thanks for all the hard work on this @daviwil!

lassik:
> * Produce usable artifacts for all platforms that can be downloaded for testing purposes

This would be very useful indeed. Gambit hardcodes the installation directory into its executable. It can be set manually with:

```
$ gsi -:help 2>&1 | grep -A 1 -- ~~
  ~~NAME=DIR         override Gambit installation directory ~~NAME where NAME
                     can be the special "bin" and "lib", or empty, or any id
```

but it would be nice to set a useful default in the CI artifacts so `-:~~` doesn't need to be set separately each time when running Gambit.

daviwil:
Thanks @lassik!  My plan to get this finished up today is to create a small test repo to find the right way to set up the Gitter task so that it runs on pass or fail of earlier jobs.  Waiting 40 minutes for all the OS builds to complete takes too much time.  Once I get that figured out, we should be able to call this PR done.

daviwil:
One other thing, I noticed in https://github.com/udem-dlteam/libs/issues/4#issuecomment-549327538 that it was expected that `--enable-single-host` couldn't be used on CI because of RAM consumption.  I am happy to report that it works fine on GitHub Actions, I've been producing builds with it for a while :)

lassik:
Great.

I think we were using the lowest-RAM version of Cirrus at the time and that's why the builds failed. They support more RAM by setting the number of gigabytes in the `.yml` file. Anyway, `--enable-single-host` is the preferred way to build Gambit so if that works in Actions by default, so much the better.

lassik:
Are you using GNU GCC or LLVM by the way? I think MacOS ships a `gcc` command that is actually a work-alike from LLVM.

daviwil:
I believe they have the machines configured to use the actual GCC they bake into the image. I ran into that "fake GCC" issue on my own Mac and the build output is different. When I run jobs on Actions, everything looks like the real GCC.

feeley:
Thank god for "the real GCC"!

lassik:
`gcc --version 2>/dev/null | head -n 1 | grep LLVM` can detect the LLVM gcc

daviwil:
You can see exactly what gets installed on each OS image by taking a look at the [actions/virtual-environments](https://github.com/actions/virtual-environments) repo.  For macOS, they mention GCC 8 and 9 explicitly: https://github.com/actions/virtual-environments/blob/master/images/macos/macos-10.15-Readme.md#language-and-runtime

lassik:
Very nice listing!

However, it says:

> Clang/LLVM 10.0.0
> gcc-8 (Homebrew GCC 8.4.0_1) 8.4.0
> gcc-9 (Homebrew GCC 9.3.0_1) 9.3.0

That probably means `gcc` is Clang/LLVM while only the versioned command names `gcc-8` and `gcc-9` are the genuine GNU GCC from as installed from Homebrew. That's how things are set up on ordinary Macs when you use MacOS and Homebrew in the obvious manner.

feeley:
My interpretation is that you would need to do

   ./configure CC=gcc-9

to get the real GCC.

daviwil:
OK, I added some extra steps to log job failures to Gitter and then one at the end to report success of the overall job.  It's not perfect, but it's good enough to be useful at the moment.  I'm working on a custom GitHub Action that should reduce duplication and make the Gitter steps look cleaner, will send a future PR when I have that working.

After this PR is merged, I'll send another PR to produce usable artifacts and upload them to the build so that they can be used, will include `~~execdir` and `--enable-single-host` in the build for those.

I'll also double-check the `gcc` invocation to make sure we're using the right one!  Thanks for calling that out.

lassik:
Seconded - `./configure CC=gcc-9` is what I use on Mac.

lassik:
> I'll also double-check the gcc invocation to make sure we're using the right one!

Gambit's `./configure` script shows prominent warnings when it detects a few situations that are likely to be problematic. We could perhaps add such a warning for LLVM gcc.

daviwil:
Confirmed that `clang` is being used in the Actions builds using the configure output that reports it:

https://github.com/daviwil/gambit/runs/709754169?check_suite_focus=true#step:4:413

Will switch to `gcc-9`, thanks again for correcting my mistake!

lassik:
Note, `gcc-9` will get phased out and some point and Homebrew will add a new GCC version. This may work to find the latest GCC executable:

`brew list gcc | grep -E '/bin/gcc-[0-9]+$'`

daviwil:
First build!  Let's see if it works:

https://github.com/gambit/gambit/actions/runs/115906674

lassik:
Never used Actions before, the UI is nice!

The subheadings make the build output easy to read:

* Set up job
* Run actions/checkout@v2
* Fetch Tags
* Build and Test
* ...

It would be nice if the **Build and Test** heading could be broken down into finer-grained steps. @feeley Is there a natural division to sub-steps, such as building `gsc` and building libraries? `make` and `make modules` are currently separate and so they can easily go into their own subheadings as well. Could `make` be broken down further?

daviwil:
I can definitely break those down, the first pass was merely just carrying across the old build/test invocations from the previous CI platforms.  It'd be good if we can decide which kinds of things we want to test so that I/we can break them apart further.

lassik:
Right. That sounds like an excellent plan.

feeley:
I think the build should be broken into the steps:

1) build latest release
2) using #1 build gsc-boot
3) using #2 generate .c files from .scm files and build HEAD
4) building modules

I'm not sure how to separate these steps because they are substeps of a single invocation of `make`.

After the tests, there's also a build using `./configure --enable-ansi-c` which builds a "bare bones" Gambit on top of ANSI C (no networking etc).  It is primarily to check that the source code doesn't contain some implicit dependencies on OS features that should be made explicit.

lassik:
> I'm not sure how to separate these steps because they are substeps of a single invocation of `make`.

* `make build-latest-release`
* `make build-gsc-boot-using-release`
* `make build-head`
* `make modules`

:)

It's best if the CI subheadings mirror the make targets somehow. Not sure how easy it is to break down the targets themselves like that, though.

feeley:
Yes that is possible, but then you are no longer using the prescribed build instructions to build Gambit (which is just `./configure;make`).  But I guess we can live with that.

lassik:
Would it be possible for `make` to be simply a sequencing of the above targets? Then the CI would be (almost) equivalent to `make`, just with the steps broken down.

However, running `make` several times with a different target each time is subtly different from running `make foo` and letting its dependency solver figure out the subtargets, as you imply.

PR #571, opened 2020-05-26
Fix configure script for MSYS2

feeley:


lassik:
The commit message talks about MSYS2 specifically, but the build script tests for `msys*`. If I'm not mistaken, only MSYS2 is based on the Cygwin DLL whereas the original MSYS 1 was a different architecture unrelated to Cygwin. Will these changes also work with MSYS 1?

feeley:
The changes don't really affect mingw.  What is added is the detection of "msys*" and handling it the same way as mingw.  The log https://ci.appveyor.com/project/feeley/gambit/builds/33120520/job/ravqhm2arb49u5tq seems to indicate that all is well with the mingw build.

Issue #569, opened 2020-05-26
Auto-detect Gambit installation directory based on executable filename?


lassik:
From PR #568:

> * Produce usable artifacts for all platforms that can be downloaded for testing purposes

This would be very useful indeed. Gambit hardcodes the installation directory into its executable. The directory can be overridden using a command line option:

```
$ gsi -:help 2>&1 | grep -A 1 -- ~~
  ~~NAME=DIR         override Gambit installation directory ~~NAME where NAME
                     can be the special "bin" and "lib", or empty, or any id
```

but it would be nice to set a useful default in the CI artifacts so `-:~~` doesn't need to be set separately each time when running Gambit.

We talked earlier about CI build artifacts with @feeley. The conversation starts at comment https://github.com/udem-dlteam/libs/issues/4#issuecomment-549296234 but we didn't come up with an ideal installation directory to use for the artifacts.

It only now occurred to me that it's possible for an executable to find its own filename on many OSes. Could Gambit optionally support finding its installation directory relative to the executable filename? That would probably be easiest for CI artifacts.

(Ping @feeley @daviwil @FredericHamel)

lassik:
For example, if `gsi` is run as `~/gambit/05060cf/bin/gsi` then `gsi` would deduce that the the installation directory is `~/gambit/05060cf`. Any other directory could be used as well. E.g. `/tmp/Gambit-master/` or whatever else. This would also mean that the directory can be completely different on the CI server and on people's personal computers. That extra flexibility would be nice.

lassik:
If runtime auto-detection is supported, it should still be possible to set `./configure --prefix=...` to a particular directory at build time and there's `--enable-multiple-versions`.

The auto-detection should probably be enabled via a new configure flag.

daviwil:
I would very much appreciate this improvement because I am currently using a CI-built Gambit on CI for other projects and I have to make sure to specify `~~bin`, etc everywhere I invoke `gsc` or `gsi`.  I'm actually dealing with a problem with this very issue today because it seems `~~bin` isn't being propagated when building imported modules that need compilation, but that's a separate issue :)

I agree with your suggestion that Gambit should probably try to resolve the path from the relative path to the executable.

lassik:
Gambit is meant to be extremely portable, and the OS APIs to get the executable filename can be brittle under unusual circumstances, so the default choice should probably remain to hardcode the installation directory.

But for situations like CI where we know it's OK to auto-detect it, the option would be nice.

[...]

PR #572, opened 2020-05-26
Add ~~execdir to refer to current executable's directory

feeley:


lassik:
Implements #569 

Thanks for the very quick turnaround!

The current PR refers to the executable directory. Is that the best alternative? If Gambit is installed using things like `--prefix=/usr/local/Gambit` or `--prefix=$HOME/.local` then the `gsc` and `gsi` go into `bin/`. This would make `~~execdir` point to `/usr/local/Gambit/bin` or `$HOME/.local/bin` instead of `/usr/local/Gambit` or `$HOME/.local`. Is the intent to use `gsi -:~~execdir/..` in this scenario?

feeley:
Yes that is the idea.  `~~execdir` is the directory containing the executable (`gsi` in this case) and in a typical installation it is in a `bin` directory that is in the same directory as the `lib`, `include`, etc directories.

lassik:
Does that mean `./configure --enable-gambitdir=~~execdir/..` would get us the flexibility we want for the CI builds?

Does the current implementation in this PR get the `realpath()` of the executable filename?

feeley:
No it doesn't get the `realpath()`.  That is a separate operation.

@daviwil I have merged the PR.  So let's start using this in the CI.

daviwil:
Yep, was just about to try it in my personal CI build.  After the initial CI PR is merged I'll send a separate PR for building and uploading usable artifacts using `~~execdir` and `--enable-shared-host`.

lassik:
We should add a note in the manual that people probably want `~~execdir/..` in the common case.

Are the tildes `~~` written in the `./configure --enable-gambitdir=` option by the way?

lassik:
What happens if gambitdir is relative instead of absolute?

feeley:
A relative gambitdir, as in `gsi -:=some/relative/path` is resolved using the CWD at the time when `gsi` started.

I don't think `-:=~~execdir/..` should be the default, because on some OSes the `lib`, `bin`, `include`, etc directories aren't installed in the same directory.

lassik:
> I don't think `-:=~~execdir/..` should be the default, because on some OSes the `lib`, `bin`, `include`, etc directories aren't installed in the same directory.

Agreed that it shouldn't be the default `./configure` option. But we should use it for the CI artifacts which are downloadable by people who follow pre-release Gambit snapshots, right?

feeley:
Agreed... because this will probably avoid grief in typical situations.

PR #577, opened 2020-05-26
Add some SRFI 13 procedures to (gambit string)

lassik:
To start off, here's a trivial patch adding `string-null?` and `string-every`.

`string-every` should also support a SRFI 14 `char-set` object for the `key` argument, but this implementation doesn't yet.

Anything obviously wrong about this?

### DIFF
--- a/lib/gambit/string/string.scm
+++ b/lib/gambit/string/string.scm
@@ -25,6 +25,30 @@
 
 ;;;----------------------------------------------------------------------------
 
+(define-procedure (string-null? (str string))
+  (fxzero? (string-length str)))
+
+(define-procedure (string-every
+                   key
+                   (str   string)
+                   (start (index-range-incl 0 (string-length str))
+                          0)
+                   (end   (index-range-incl start (string-length str))
+                          (string-length str)))
+  (cond ((char? key)
+         (let loop ((i start))
+           (or (fx= i end)
+               (and (char=? key (string-ref str i))
+                    (loop (fx+ i 1))))))
+        ((procedure? key)
+         (let loop ((true-value #t) (i start))
+           (if (fx= i end) true-value
+               (and true-value
+                    (loop (key (string-ref str i))
+                          (fx+ i 1))))))
+        (else
+         (error "type error"))))
+
 (define-prim-nary-bool (##string=? str1 str2)
   #t
   #t
--- a/lib/gambit/string/string.sld
+++ b/lib/gambit/string/string.sld
@@ -15,6 +15,8 @@
   (export
 
 string?
+string-null?
+string-every
 make-string
 string
 string-length


lassik:
Oh yeah, the `(error "type error")` should definitely be replaced with a proper error message.

feeley:
Given that SRFI 13 depends on SRFI 14 (char sets), perhaps we should work on SRFI 14 first. Also, for `define-procedure` I'm thinking of adding a way to easily define new type checkers, for example the type "char-or-char-set-or-procedure".

lassik:
> Given that SRFI 13 depends on SRFI 14 (char sets), perhaps we should work on SRFI 14 first.

We can probably work on them in parallel. I can add 14 to this same PR. Which Gambit library should it go into? The R7RS-large import is planned to be `(scheme charset)` so should we have `(gambit charset)`?

> Also, for `define-procedure` I'm thinking of adding a way to easily define new type checkers, for example the type "char-or-char-set-or-procedure".

For sum types, the procedure implementation often needs to dispatch on the type of the argument. For example,

```
(cond ((procedure? key) ...)
      ((char? key ...)
      (else ...))
```

This would duplicate the type check - one check by `define-procedure`, another check manually in the `cond`.

Common Lisp has `etypecase` for situations like this: http://www.ai.mit.edu/projects/iiip/doc/CommonLISP/HyperSpec/Body/mac_typecasec_cm_etypecase.html We could have something similar to use in the procedure body. It could raise errors similar to `define-procedure`'s native type checker.

lassik:
Or does it benefit `gsc`'s optimizer in a particular way if all type checks are done by `define-procedure`?

feeley:
I'd like the `define-procedure` to declare the type (in this case "char-or-char-set-or-procedure") even if `define-procedure` is not generating the type check, so that the `define-procedure` can play the role of documentation.  This could be done with an annotation on the type, like `(nonautomatic char-or-char-set-or-procedure)`, or possibly `(nonautomatic (or char char-set procedure))`.  The definition of that "type check exception" would specify the condition, the error message and a procedure would be implicitly defined to raise that exception.

I have a feeling `charset` should be defined in the `char` library because they share a lot of logic (Unicode tables and such).  And the `char` library is rather small currently.

lassik:
> I'd like the `define-procedure` to declare the type (in this case "char-or-char-set-or-procedure") even if `define-procedure` is not generating the type check, so that the `define-procedure` can play the role of documentation.

Makes sense.

> This could be done with an annotation on the type, like `(nonautomatic char-or-char-set-or-procedure)`, or possibly `(nonautomatic (or char char-set procedure))`. The definition of that "type check exception" would specify the condition, the error message and a procedure would be implicitly defined to raise that exception.

Sounds like a reasonable approach.

Or perhaps instead of an error, there should be an extra variable that receives a symbol saying what the actual type of the argument is for that procedure call:

```scheme
(define-procedure (string-every (key (one-of key-type char char-set procedure))
                                ...)
  (case key-type
    ((char)
     ...)
    ((char-set)
     ...)
    ((procedure)
     ...)))
```

> I have a feeling `charset` should be defined in the `char` library because they share a lot of logic (Unicode tables and such). And the `char` library is rather small currently.

feeley:
That is still a double dispatch.

lassik:
Yes, but the predicates need to be called once only. The second dispatch is just testing that symbols are `eq?`.

The only way to avoid double dispatch would be a `case-lambda` style solution (but a type case). But what if the procedure's behavior depends on the interaction of the types of two arguments? The combinations could get out of hand. Then again, maybe it's manageable in practice.

feeley:
For many predicates (`string?`, `char?`, `procedure?`, `fixnum?`, ...) the checking of the predicate is similar in cost to `eq?`.  However, the extra register pressure in adding a flag variable is going to cause some spills that will have an indirect cost on using a flag variable.

lassik:
Fair enough. Is the `case-lambda` style is too fancy?

feeley:
I guess a special form for "type case" would work in many cases.  It can get hairy if more than one parameter can have multiple types.  This is what happens for generic arithmetic where there's a dispatch on the representation of the numbers.

Here's a procedure with a single dispatch:

```
(define-prim (##integer? x)
  (macro-number-dispatch x #f
    #t ;; x = fixnum
    #t ;; x = bignum
    #f ;; x = ratnum
    (macro-flonum-int? x) ;; x = flonum
    (macro-cpxnum-int? x))) ;; x = cpxnum
```

and one with two dispatches:

```
(define-prim (##*2 x y)

  (##define-macro (type-error-on-x) `'(1))
  (##define-macro (type-error-on-y) `'(2))
  (##define-macro (fixnum-overflow) `#f)

  (macro-number-dispatch x (type-error-on-x)

    (macro-number-dispatch y (type-error-on-y) ;; x = fixnum
      (cond ((##fx= y 0)
             0)
            ((if (##fx= y -1)
                 (##fx-? x)
                 (##fx*? x y))
             => (lambda (result) result))
            (else
             (macro-if-bignum
              (##bignum.* (##fixnum->bignum x) (##fixnum->bignum y))
              (fixnum-overflow))))
      (cond ((##fxzero? x)
             0)
            ((##fx= x 1)
             y)
            ((##fx= x -1)
             (##negate y))
            (else
             (##bignum.* (##fixnum->bignum x) y)))
      (cond ((##fxzero? x)
             0)
            ((##fx= x 1)
             y)
            ((##fx= x -1)
             (##negate y))
            (else
             (##ratnum.* (macro-exact-int->ratnum x) y)))
      (cond ((and (macro-special-case-exact-zero?)
                  (##fxzero? x))
             0)
            ((##fx= x 1)
             y)
            (else
             (##fl* (##fixnum->flonum x) y)))
      (cond ((and (macro-special-case-exact-zero?)
                  (##fxzero? x))
             0)
            ((##fx= x 1)
             y)
            (else
             (##cpxnum.* (macro-noncpxnum->cpxnum x) y))))

    (macro-number-dispatch y (type-error-on-y) ;; x = bignum
      (cond ((##fx= y 0)
             0)
            ((##fx= y 1)
             x)
            ((##fx= y -1)
             (##negate x))
            (else
             (##bignum.* x (##fixnum->bignum y))))
      (##bignum.* x y)
      (##ratnum.* (macro-exact-int->ratnum x) y)
      (##fl* (##exact-int->flonum x) y)
      (##cpxnum.* (macro-noncpxnum->cpxnum x) y))

    (macro-number-dispatch y (type-error-on-y) ;; x = ratnum
      (cond ((##fxzero? y)
             0)
            ((##fx= y 1)
             x)
            ((##fx= y -1)
             (##negate x))
            (else
             (##ratnum.* x (macro-exact-int->ratnum y))))
      (##ratnum.* x (macro-exact-int->ratnum y))
      (##ratnum.* x y)
      (##fl* (##ratnum->flonum x) y)
      (##cpxnum.* (macro-noncpxnum->cpxnum x) y))

    (macro-number-dispatch y (type-error-on-y) ;; x = flonum
      (cond ((and (macro-special-case-exact-zero?) (##fxzero? y))
             0)
            ((##fx= y 1)
             x)
            (else
             (##fl* x (##fixnum->flonum y))))
      (##fl* x (##exact-int->flonum y))
      (##fl* x (##ratnum->flonum y))
      (##fl* x y)
      (##cpxnum.* (macro-noncpxnum->cpxnum x) y))

    (macro-number-dispatch y (type-error-on-y) ;; x = cpxnum
      (cond ((and (macro-special-case-exact-zero?) (##fxzero? y))
             0)
            ((##fx= y 1)
             x)
            (else
             (##cpxnum.* x (macro-noncpxnum->cpxnum y))))
      (##cpxnum.* x (macro-noncpxnum->cpxnum y))
      (##cpxnum.* x (macro-noncpxnum->cpxnum y))
      (##cpxnum.* x (macro-noncpxnum->cpxnum y))
      (##cpxnum.* x y))))
```

lassik:
And not all operations are commutative or associative, so swapping the order of the arguments to reduce the number of cases doesn't work as a general approach.

feeley:
The main character class vector is uncompressed:

```
> (u8vector-length ##unicode-class)
201546
> (write (subu8vector ##unicode-class 0 70))
#u8(10 10 10 10 10 10 10 10 10 11 11 11 11 11 10 10 10 10 10 10 10 10 10 10 10 10 10 10 10 10 10 10 11 10 10 10 10 10 10 10 10 10 10 10 10 10 10 10 0 1 2 3 4 5 6 7 8 9 10 10 10 10 10 10 10 15 15 15 15 15)
```

This vector covers all the allocated characters with one byte per character indicating the "class".  For example, the class for whitespace is 11, and for digit it is <= 9.  This gives a constant time classification (with a very small constant) and also allows in-order traversal (although that is inefficient given that most classes are typically a small fraction of all the characters.

lassik:
What is the compression like?

lassik:
I looked at integrating SRFI 14 into `lib/gambit/char/char#.scm` yesterday, and I think I know how to do it now.

However, currently this fails:

```
Gambit v4.9.3-1220-ga3a2f895

> (import (gambit char))
*** ERROR IN gambit/char# -- Operator is not a PROCEDURE
(#!unbound #!unbound '#<procedure #2>)
```

How should I recompile `char#.scm` while programming, and do I load the new macro definitions are loaded into gsi? The obvious things like `import` and `load` do not work currently. I tried setting `GAMBITDIR` too.

lassik:
John Cowan posted a very helpful addendum at https://srfi.schemers.org/srfi-14/contrib/unicode-2019/CharsetDefs.html explaining how to derive the SRFI 14 charsets for the current version of Unicode. We have to get at the Unicode character **properties** and **categories**, which is easy enough to do within the existing framework of `char#.scm`.

lassik:
We should probably use the bitmap representation for all charset objects. Since that representation keeps the character ranges in sorted order, it ought to be simple to implement cursors and set operations.

PR #578, opened 2020-05-27
Fix C compiler warnings (^ bitwise operator).

bheesham:
When configured with OpenSSL I saw the following warnings:

```
os_io.c:5249:15: warning: '&' within '^' [-Wbitwise-op-parentheses]
              & SSL_RECEIVED_SHUTDOWN
              ^~~~~~~~~~~~~~~~~~~~~~~
os_io.c:5249:15: note: place parentheses around the '&' expression to silence this warning
              & SSL_RECEIVED_SHUTDOWN
              ^~~~~~~~~~~~~~~~~~~~~~~
os_io.c:5381:15: warning: '&' within '^' [-Wbitwise-op-parentheses]
              & SSL_RECEIVED_SHUTDOWN
              ^~~~~~~~~~~~~~~~~~~~~~~
os_io.c:5381:15: note: place parentheses around the '&' expression to silence this warning
              & SSL_RECEIVED_SHUTDOWN
              ^~~~~~~~~~~~~~~~~~~~~~~
```

---
This change is [Reviewable](https://reviewable.io/reviews/gambit/gambit/578)

### DIFF
--- a/lib/os_io.c
+++ b/lib/os_io.c
@@ -5245,9 +5245,8 @@ ___stream_index *len_done;)
       else if (n == 0)
         {
           /* Check if the other part initiated a shutdown and we didn't reply */
-          if (SSL_get_shutdown (d->tls)
-              & SSL_RECEIVED_SHUTDOWN
-              ^ SSL_SENT_SHUTDOWN)
+          int shut = SSL_get_shutdown (d->tls);
+          if ((shut & SSL_RECEIVED_SHUTDOWN) && !(shut & SSL_SENT_SHUTDOWN))
             {
               clear_tls_error_queue();
               SSL_shutdown (d->tls);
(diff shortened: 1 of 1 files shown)


lassik:
Why is xor used here?

lib/os_io.c
```diff
@@ -5246,8 +5246,7 @@ ___stream_index *len_done;)
         {
           /* Check if the other part initiated a shutdown and we didn't reply */
```
feeley, line 5247: The test in the `if` is wrong (and was wrong before).  The code should be:
```
int shut = SSL_get_shutdown (d->tls);
if ((shut & SSL_RECEIVED_SHUTDOWN) && !(shut & SSL_SENT_SHUTDOWN) )
```
In other words "if we received a shutdown but not yet sent a shutdown then ...".

bheesham: Ah! Good catch, thanks. I updated the PR to use your code.

PR #579, opened 2020-05-27
Fix invocation of gsc when building imported "_must-build_" modules on Windows

daviwil:
This change fixes a couple of issues I encountered when building a project which imports a module that uses a `._must-build_` file to force compilation via `gsc` before it gets imported.

The first issue is that the `.bat` scripts generated by `bin\makefile.in` were not invoking `gsc` from the folder in which they live.  This results in the following error when importing the `_must-build_` module in `gsi`:

```
'gsc' is not recognized as an internal or external command,
operable program or batch file.
```

The fix is to use `~dp0` in the path to invoke `gsc`/`gsi` so that the batch file's execution path is used to invoke it's peer exe.  I've fixed all instances of this pattern in `bin\makefile.in`.

The second issue is that `gsc-script.bat` was being launched as `gsc-script.exe` in `##build-module-subprocess`, resulting in this error:

```
*** ERROR IN ##build-module-subprocess -- No such file or directory
(##run-subprocess
 "D:\\a\\crash-the-stack\\crash-the-stack\\gambit\\bin\\gsc-script.exe"
 '("-:=~~execdir\\..,~~lib=./gambit/lib,~~userlib=./lib" "-e" "(##begin (##m...
 #f
 #f
 #f)
```

The fix is to use `##os-bat-extension-string-saved` instead of `##os-exe-extension-string-saved` when building the name of `gsc-script`.  I've tested this on Linux too just to make sure that a `.bat` isn't added to the name on non-Windows platforms.

### DIFF
--- a/bin/makefile.in
+++ b/bin/makefile.in
@@ -232,7 +232,7 @@ six@exe@: makefile
 	if test "@bat@" = ""; then \
 	  $(LN_S) $(rootfromhere)/gsi/$(GSI_EXECUTABLE) six@exe@; \
 	else \
-	  echo "@gsi %*" > six@bat@; \
+	  echo "@%~dp0gsi %*" > six@bat@; \
 	fi
 
 gsi-script@bat@: makefile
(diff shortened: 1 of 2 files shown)


daviwil:
@feeley Any concerns with this one?  It's currently blocking my ability to build another project.  Happy to answer any questions or test anything you think might be affected by it!

feeley:
Sorry... I've been having issues with my email since 2 days and the notification was overlooked.  Thanks for the fix!

daviwil:
No worries, thanks a lot!

PR #583, opened 2020-05-27
Produce usable Gambit builds for each platform, upload as artifacts

daviwil:
This change adds usable Gambit artifacts to CI runs both for pull requests and pushes to `master`.  I'm building Gambit for these artifacts with the following configure params:

```
./configure --enable-single-host --enable-gambitdir=~~execdir/.. --prefix=$(pwd)/dist 
```

After finishing up the build, I package up the `dist/` folder in a platform-specific way (`.zip` for Windows, `.tar.gz` for Linux and macOS) and upload it as an artifact to the workflow run.

You can see a test build with artifacts here: https://github.com/daviwil/gambit/actions/runs/116950482.  That run partially failed due to how I was running the build in MSYS2 but I should have that fixed now.

Note that for the Linux and MacOS builds the `.tar.gz` is wrapped in an outer `.zip` file.  This is a quirk of how artifacts are downloaded from GitHub Actions.  When I write the CI automation to download these builds from workflows in other CI jobs, I'll take care of unwrapping everything appropriately.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -36,7 +36,7 @@ jobs:
 
     - name: Build
       shell: cmd
-      run: msys2do ./configure --enable-debug --enable-multiple-threaded-vms;make -j4; make modules; make install
+      run: msys2do ./configure --enable-debug --enable-multiple-threaded-vms;make -j4; make modules
 
     # Only run tests for MinGW64 for now
     - name: Test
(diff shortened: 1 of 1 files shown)


daviwil:
Here are the artifacts generated by the build for this PR: https://github.com/gambit/gambit/actions/runs/116998778

PR #584, opened 2020-05-31
Add 'headers' parameter to http-get for specifying request headers

daviwil:
This change adds a keyword parameter called `headers` to `http-get` so that an alist of request headers can be specified for the request.  I've also converted the two existing optional parameters to keyword parameters to make it a little easier to call.

Since I'm not as familiar with the coding patterns of Gambit core libraries, could you help me figure out the right way to replace the `macro-check-*` calls that I removed from `http-get`?  I'm not sure how that pattern changes when switching to keyword arguments.

/cc @FredericHamel

### DIFF
--- a/lib/_http/_http.scm
+++ b/lib/_http/_http.scm
@@ -4,6 +4,7 @@
 
 ;;; Copyright (c) 2019-2020 by Frédéric Hamel, All Rights Reserved.
 ;;; Copyright (c) 2020 by Marc Feeley, All Rights Reserved.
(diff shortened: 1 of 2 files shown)


lassik:
@feeley Just added a convenient `define-procedure` primitive but I don't think it supports keyword arguments yet.

daviwil:
Thanks!  I might take a look to see if I can figure out how to add that if it's not there already.

lassik:
https://github.com/gambit/gambit/commit/050bdfaa7d36f834a4537cb563cb65a6b3ed9da4

PR #588, opened 2020-06-05
Improve workflow and job name display in CI

daviwil:
This is an attempt to improve how job names are displayed in the GitHub Actions check status list.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -38,14 +38,14 @@ jobs:
 
     - name: Build
       shell: cmd
-      run: msys2do ./configure --enable-debug --enable-multiple-threaded-vms;make -j4; make modules
+      run: msys2 ./configure --enable-debug --enable-multiple-threaded-vms;make -j4; make modules
 
     # Only run tests for MinGW64 for now
     - name: Test
       if: ${{ matrix.msystem == 'MINGW64' }}
       shell: cmd
       run: |
-        msys2do make check
+        msys2 make check
 
     - name: Build Gambit for Use
       shell: msys2 {0}
@@ -88,13 +88,13 @@ jobs:
       run: |
         call "C:\Program Files (x86)\Microsoft Visual Studio\2019\Enterprise\VC\Auxiliary\Build\vcvarsall.bat" x64 -vcvars_ver=14.16
         set PATH=%RUNNER_TEMP%\msys\msys64\usr\bin;%PATH%
-        msys2do ./configure --enable-c-opt=-Od --enable-debug CC="cl"; make -j4; make modules
+        msys2 ./configure --enable-c-opt=-Od --enable-debug CC="cl"; make -j4; make modules
 
     - name: Test
       shell: cmd
       run: |
         call "C:\Program Files (x86)\Microsoft Visual Studio\2019\Enterprise\VC\Auxiliary\Build\vcvarsall.bat" x64 -vcvars_ver=14.16
-        msys2do make check
+        msys2 make check
 
     - name: Build vstudio.bat
       shell: cmd
@@ -110,7 +110,7 @@ jobs:
       run: |
         call "C:\Program Files (x86)\Microsoft Visual Studio\2019\Enterprise\VC\Auxiliary\Build\vcvarsall.bat" x64 -vcvars_ver=14.16
         mkdir dist
-        msys2do ./configure --enable-single-host --enable-gambitdir=~~execdir/.. --prefix=$(pwd)/dist CC="cl";make -j4; make modules; make install
+        msys2 ./configure --enable-single-host --enable-gambitdir=~~execdir/.. --prefix=$(pwd)/dist CC="cl";make -j4; make modules; make install
 
     - name: Upload Build Artifact
       uses: actions/upload-artifact@v1
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -1,4 +1,4 @@
-name: Gambit - CI
+name: Gambit
 
 on:
   pull_request:
@@ -11,6 +11,7 @@ on:
 
 jobs:
   Windows-mingw:
+    name: "Windows - ${{ matrix.msystem }}"
     runs-on: windows-latest
     strategy:
       matrix:
@@ -63,6 +64,7 @@ jobs:
       if: failure() && (github.event_name == 'push' || github.event.pull_request.head.repo.full_name == github.repository)
 
   Windows-msvc:
+    name: "Windows - MSVC"
     runs-on: windows-latest
 
     steps:
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -28,9 +28,10 @@ jobs:
       run: git fetch --prune --unshallow
 
     - name: Install MSYS2
-      uses: eine/setup-msys2@v0
+      uses: eine/setup-msys2@v1
       with:
         update: true
+        cache: true
         install: "autoconf git make mingw-w64-${{ matrix.arch }}-gcc mingw-w64-${{ matrix.arch }}-make mingw-w64-${{ matrix.arch }}-libwinpthread-git"
         path-type: inherit
         msystem: ${{ matrix.msystem }}
@@ -74,9 +75,10 @@ jobs:
       run: git fetch --prune --unshallow
 
     - name: Install MSYS2
-      uses: eine/setup-msys2@v0
+      uses: eine/setup-msys2@v1
       with:
         update: true
+        cache: true
         install: "autoconf make patch"
         path-type: inherit
         msystem: MSYS


feeley:
Now that the CI is producing artifacts, it would be interesting on Windows and macOS to build the standard installers using the scripts in the prebuilt subdirectory.  The Windows installer is created with NSIS and the macOS installer is created with PackageMaker.  Is there an easy way to map the release of Gambit (or commit hash) to the URL that contains the artifacts?

daviwil:
Yep, there is a consistent URL structure for GitHub Release downloads, for example with MSYS2's installer.  The `2020-06-02` in the URL is the tag name for the release:

- Release page: https://github.com/msys2/msys2-installer/releases/tag/2020-06-02
- Installer download: https://github.com/msys2/msys2-installer/releases/download/2020-06-02/msys2-x86_64-20200602.exe

I'll see if I can get the installers building in CI.  Maybe we should have a separate release workflow that only gets triggered when a tag is pushed?  We can experiment with it on my fork.

daviwil:
I realize now that I might have answered a different question than what you asked: were you referring to CI artifacts or assets that are attached to a GitHub release? For the former, it's nontrivial to get the download link of CI artifacts, you can see what I had to do for that here (basically multiple API calls) :

https://github.com/daviwil/download-gambit/blob/5f8b116fa7b7df789b46acc322a41c85bef3c21b/index.js#L56

I'd definitely recommend using GitHub Releases for distributing Gambit releases though, it's very easy to automate.

daviwil:
Back on the topic of the installer scripts: It looks like they're designed run builds on VMs with a very different configuration than what is on CI.  I'm hesitant to change what's already there because I don't want to disrupt your existing release process.  Should I create a separate `prebuilt-ci` task (and potentially `release-ci`) that has its own scripts that are tailored to the CI environment?

feeley:
I think the scripts could evolve to work through the CI (for example they could install the required NSIS and PackageMaker or newer/better equivalent tools).  I like that there is an MSYS2 GUI-based installer.

One can wonder if a GUI-based installer is required for Gambit.  After all developers typically use package managers to install Gambit, e.g. `brew install gambit-scheme`.  However, for some developers (students, novice developers, etc) it is convenient to have a GUI-based installer.  Moreover one has to think about how "end-user" programs built with Gambit are installed by end-users.  Here there's a need for an installer that conforms to the practices for a given platform (and there can be more than one way to install, such as "app stores", "drag to Applications folder" and GUI-based installers).

One peculiarity of the installer for Windows is that it executes at install time the binpatch.exe program to patch the gsi.exe and gsc.exe programs (and libgambit.lib, libgambit.dll, etc) so that the Gambit installation path in those files will refer to the directory where the user has chosen to install Gambit.  This is done by building Gambit with

    configure --prefix=C:/Gambit/././././././././././././././././././././././././././././././././././././././././././././././././././././././././././././././././././././././././././.

which will put that long path, an easily recognizable pattern, in gsi.exe and other files.  The binpatch.exe program replaces that string with the actual installation directory (without actually changing the number of bytes, i.e. a NUL character will mark the end of the new string which overwrites the old one).  Note that the trick of using `--enable-runtime-options==~~execdir/..` will not work because a Scheme program compiled with Gambit to P.exe and installed to `C:/other/install/dir/bin` will incorrectly think that Gambit's `~~`, `~~lib`, etc can be found relative to `C:/other/install/dir`.  Gambit does not have a strategy for installing P.exe (which should install the "Gambit RTS" where P.exe expects it, either in a placed shared with other programs or local to the P.exe installation).

For macOS this trick is not used because Gambit is always installed in `/Library/Gambit` .  Similarly no binary patching is required for other OSes (linux, etc) where we know at Gambit build time where Gambit will be installed.

However, "how to install an end-user program on a machine where Gambit is not yet installed" is still an open question, particularly when the "installer" should not expose to the end-user that it was built with Gambit, and should not clobber any installation of Gambit previously installed for another program also built with Gambit.

Should Gambit come with scripts to build GUI-based installers?  Is this reasonable?  What should be installed (e.g. should gsi and gsc be installed too, after all the `compile-file` procedure could be seen as part of the Gambit RTS)?  Could "the intelligent linker" be part of the solution (to determine which parts of the Gambit RTS are required)?

lassik:
> However, "how to install an end-user program on a machine where Gambit is not yet installed" is still an open question, particularly when the "installer" should not expose to the end-user that it was built with Gambit, and should not clobber any installation of Gambit previously installed for another program also built with Gambit.

WIndows .exe files support "resources" (strings, icons, etc.) that are built into the .exe file itself. Perhaps the prefix directory could be stored as a resource string on Windows. It might be possible to store entire files (`.scm` source code, documentation, etc.) as resources with acceptable performance. I believe they are not loaded into the process image by default, and must be loaded manually on demand, so it should be possible to ship even large .exe files without incurring undue runtime overhead.

feeley:
I guess we need a Windows wizard to look into this.

But the general question remains... what are the options for installing an "end-user" program built with Gambit and which option is the best (and probably should be supported by the Gambit linker)?  This question applies to all platforms, and on some of them options are limited (for example on iOS the app must contain all of the dependencies, i.e. the Gambit RTS and supporting files that are needed).

daviwil:
Regarding installing a Gambit-compiled program on a machine without Gambit, I've done it without issue so far.  You can try running one of the build artifacts of a game I'm working on here:

https://github.com/substratic/crash-the-stack/actions/runs/118405452

It even works fine on Windows.  I did have to make sure certain dependency libraries were copied into the same (check out [this dark magic](https://github.com/substratic/build/blob/master/build.scm#L174-L260)) folder as the executable, but nothing Gambit-specific.  Maybe I missed a detail there that might be relevant for other Gambit users?

I'm happy to help look into general "How do distribute Gambit apps" questions since it does matter to me as a developer of Gambit apps :)

lassik:
I dabbled with .exe resources at some point long ago, and it didn't seem like wizard-level stuff - the relevant part of the Windows API was pretty basic and there was even a third-party program called Resource Hacker (I think) to modify the resources in any .exe file without having to recompile using a resource compiler.

This looks like the right API: https://docs.microsoft.com/en-us/windows/win32/menurc/using-resources

daviwil:
Yep, it's definitely possible to tweak resources in an `exe` without too much pain, that might be a good solution for the problem.

lassik:
Linux and the BSDs use ELF executables which can also store sections that are not loaded into the runtime image by the OS. But unlike Windows resources, there is no standard convention for which format those sections should be in (I think DWARF debugging information is stored in these sections, but there's a provision for storing arbitrary data).

lassik:
@daviwil Do you know whether there's a limit on the number or size of resources that can be stored in one .exe? Are string resources suitable for storing arbitrary binary data (possibly including null bytes and the like)?

lassik:
MacOS (and probably iOS) use an executable format called Mach-O (which probably originates from the Mach microkernel). Not sure how to store resource-like things there. But MacOS also uses those `.app` directories to store desktop programs that go in the `Applications` folder.

feeley:
> Regarding installing a Gambit-compiled program on a machine without Gambit, I've done it without issue so far. You can try running one of the build artifacts of a game I'm working on here:
> 
> https://github.com/substratic/crash-the-stack/actions/runs/118405452
> 
> It even works fine on Windows. I did have to make sure certain dependency libraries were copied into the same (check out [this dark magic](https://github.com/substratic/build/blob/master/build.scm#L174-L260)) folder as the executable, but nothing Gambit-specific. Maybe I missed a detail there that might be relevant for other Gambit users?
> 
> I'm happy to help look into general "How to distribute Gambit apps" questions since it does matter to me as a developer of Gambit apps :)

What I'm worried about is support for non-primitive features. Specifically, in the Gambit app can you:

1) `(help 'help)`  this requires a web browser and installed documentation
2) `(##load-module '_digest)` ?  in other words load a builtin Gambit module
3) `(eval '(let () (import (srfi 28)) (format "hi ~s" 'you)))` ?
4) `(##install-module (##parse-module-ref 'github.com/gambit/hello))` ?
5) `(compile-file "foo.scm")` ?  this one even requires an installed C compiler

So the question is what are the features of Gambit that can and cannot be used in the installed Gambit app?  In principle all of them could be supported, but I presume you are drawing the line somewhere (somewhat arbitrarily) so that not all of Gambit has to be installed with the app.

daviwil:
I suppose it depends on the type of application one is trying to make.  In my case, I'm not exposing any of Gambit's functionality directly to the user or using any kind of dynamic runtime loading of modules that I haven't already compiled into the exe.  For example, I use the `_hamt` module in the project I linked you; just compiling it into the exe with `gsc` along with my own modules was sufficient to have it work on another machine.

If someone wants to make an application where they can evaluate Scheme forms with the added benefit of Gambit's modules, we'll have to think again about this.  We might be able to provide behavior in `gsc` to create a more complete host application for Gambit.  Beyond that, I'd say one should just distribute Gambit with their application.

Issue #590, opened 2020-06-06
Possible enhancements to ci.yml


eine:
@daviwil, you might want to consider some of the following enhancements to `.github/workflows/ci.yml`:

- Instead of `git fetch --prune --unshallow`, use actions/checkout@v2's option `fetch-depth: 0`. You might need to use `git config --global core.autocrlf input`, as commented in https://github.com/eine/setup-msys2#usage. See actions/checkout#250 and actions/checkout#226.

- Use `defaults: run: shell` as explained in https://github.com/eine/setup-msys2#usage, to avoid writting `run: msys2 whatever` multiple times.

- Do you really need `path-type: inherit`? That might produce conflicts if `git`, `make` or some other tools are available in the PATH.

daviwil:
@eine Thanks for the tips!  I do want to switch over to using `msys2` as the shell.  I didn't know about the `default` setting, that's helpful!

For `fetch-depth: 0`, Gambit's `makefile` relies on having Git tags available for part of the build so I had to go a different route to ensure that the tags were getting fetched correctly.  In the past I wasn't able to use `fetch-depth: 0` because it didn't pull tags, but I see now that they added it thanks to the issue you linked: https://github.com/actions/checkout/issues/250#issuecomment-634799150.

Regarding `path-type: inherit`, I think I ran into issues if I didn't use that back when I was using `numworks/setup-msys2` but I'll try removing it now just in case it isn't needed any longer.

eine:
> @eine Thanks for the tips! I do want to switch over to using `msys2` as the shell. I didn't know about the `default` setting, that's helpful!

In fact, I found about it just a few days ago 😆 

> For `fetch-depth: 0`, Gambit's `makefile` relies on having Git tags available for part of the build so I had to go a different route to ensure that the tags were getting fetched correctly. In the past I wasn't able to use `fetch-depth: 0` because it didn't pull tags, but I see now that they added it thanks to the issue you linked: [actions/checkout#250 (comment)](https://github.com/actions/checkout/issues/250#issuecomment-634799150).

I was using `unshallow` for the same reason and I changed it precisely when it was fixed in the issue I linked.

> Regarding `path-type: inherit`, I think I ran into issues if I didn't use that back when I was using `numworks/setup-msys2` but I'll try removing it now just in case it isn't needed any longer.

[...]

PR #591, opened 2020-06-06
Simplify CI patterns for code checkout and MSYS/MinGW invocation

daviwil:
@eine left some helpful tips in issue #590, this PR implements them to simplify our CI jobs further.  I'm also adding `texinfo` as a dependency to be installed for Windows builds so that we can build `info` files in `make doc`.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -34,7 +34,7 @@ jobs:
       with:
         update: true
         cache: true
-        install: "autoconf git make texinfo mingw-w64-${{ matrix.arch }}-gcc mingw-w64-${{ matrix.arch }}-make mingw-w64-${{ matrix.arch }}-libwinpthread-git"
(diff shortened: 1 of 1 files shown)


eine:
@daviwil, it seems that `tar` might be installed on the default virtual environment, but not in MSYS2. Hence, when removing the path inheritance, it is not found. This is strange to me, as I'd expect `tar` to be part of the base MSYS2. Nevertheless, it should be fixable by adding `tar` to `install:`.

daviwil:
@eine Thanks!  That's probably the issue I had run into before.  I'll bet `inherit` caused the `tar` from the Git for Windows distribution to get picked up in `PATH`.

eine:
JFYTK, the default `tar` in the virtual environment is kind of broken: actions/virtual-environments#282. That's why 7z is used in setup-msys2: numworks/setup-msys2#26.

daviwil:
Good to know, thanks!

PR #595, opened 2020-06-09
Improve module setup speed

feeley:
The module meta information for the procedure name has changed to avoid a run time constructions of the symbols.  This saves about 1 millisecond CPU time.  This doesn't seem like much but it is about 25% of the startup CPU time.  Here is old vs new:

    % zsh -c 'TIMEFMT="%uE %uU %uS"; time gsi -e "(exit)"'
    7557us 4127us 2048us
    % zsh -c 'TIMEFMT="%uE %uU %uS"; time gsi/gsi -e "(exit)"'
    6922us 3294us 2144us

So it takes a total of 7 milliseconds to startup and exit `gsi` or about 140 executions per second for an empty program.


daviwil:
Any improvement to module setup speed will be great!  When I load up an application into `gsi` with a lot of modules it takes 7-10 seconds before everything is loaded and running.  I suppose I could be doing something wrong, might have to check with you on that sometime.

feeley:
For a compiled application 7-10 seconds is too much...  The gsc compiler is about 150KLOC and the startup time is 0.018 second.

daviwil:
This is actually when loading up the modules into `gsi` to run it interactively; when I compile the application with `gsc` it starts up instantly.

feeley:
I just noticed you said you were loading into gsi.  It could be that you haven't compiled the builtin modules with `make modules`.  One of those modules is `_define-library` which implements `define-library` and is much faster when compiled.

daviwil:
I did use `make modules` to compile the built-in modules, but your comment gave me a clue to check on something I did which might have caused this.  I'll let you know what I find!

PR #598, opened 2020-06-10
Universal backend: fix empty else when using Python

feeley:


daviwil:
Seems like this could be merged!

Issue #594, opened 2020-06-09
Build Gambit installers on CI pipeline


daviwil:
This issue is meant to track the work to get Gambit installers building on a GitHub Actions release pipeline.  This might be used to automate releases when new tags are pushed to the repo.

daviwil:
@feeley - I've started setting up a release pipeline but I'm hitting a `*** FATAL ERROR - Heap overflow` when using the bootstrapped `gsc-boot.exe` to build the latest `gsc`.  I decided to bump up the heap size when running `gsc-boot` (`-:h200000`) and I'm still getting [errors when building `_gsc.scm`](https://github.com/daviwil/gambit/runs/754717370?check_suite_focus=true#step:4:1953):

```
../gsc-boot -:h2000000,~~bin=../bin,~~lib=../lib,~~include=../include -f  -prelude "(##namespace(\"c#\"))(##include\"../lib/header.scm\")" -warnings -c _gsc.scm
/bin/sh: ../gsc-boot: Resource temporarily unavailable
```

Is it possible that we're running out of memory on the build machine?  Is there any other parameter I could tweak to help alleviate this issue?

daviwil:
I'm trying to put together the optimized build based on `make dist` but I find myself a little uncertain about the process.  Does this sound right?

**CI build on `master`**

1. Bootstrap Gambit from scratch
2. `make dist` in the _root_ of the repo to gather the source of the bootstrapped `gsc`
3. Store the resulting `tgz` as an artifact for that OS/arch.

**PR build**

1. Clone the repo without tags so that the dummy `gsc-boot` will be used
2. Download the source `tgz` from the last successful build on `master`
3. Unpack `.c` files from the `tgz` into `gsc/` and `lib/` into the repo folder
4. Run `make`, etc to build everything as usual

Should I be packing up `make dist` from `boot/` instead and then unpack it to `boot/` in the PR build, build it, `make bootstrap` and then run `make` on the repo folder?

daviwil:
btw @feeley, something about one of the latest commits to `master` (https://github.com/gambit/gambit/commit/5bd523790d4d548b6a8cef4be53fa80a51d18bf6) has broken tests on MinGW builds: 

https://github.com/gambit/gambit/runs/758606784#step:5:170

feeley:
> I'm trying to put together the optimized build based on `make dist` but I find myself a little uncertain about the process. Does this sound right?
> 
> **CI build on `master`**
> 
[...]

Issue #320, opened 2017-12-26
MingW 64bit-compiled binaries crash at close time on Windows 10 (but not mingw32, and not visualc32/64)


adam-7:
Reproduction:
 * On Windows 10 64bit, do:
 * Download a Gambit prebuilt executable, e.g. http://www.iro.umontreal.ca/~gambit/download/gambit/v4.8/prebuilt/gambit-v4_8_8-windows-mingw64.exe .
 * Either install Gambit by running the file, or extract the gsi/gsc file 
 * Open gsc or gsi
 * Do ",q"
 * A dialog saying "gsc/gsi.exe has stopped working" appears.

I did the above on Windows 10 64bit with all the latest updates only, for MingW64-compiled Gambit 4.8.8, 4.8.0, 4.7.9 and 4.7.0, and it produced the described error behavior on all of them.

I also tried the above on Windows 10 64bit with all the latest updates only, for MingW32-compiled Gambit 4.8.8 and 4.7.0, as well as VisualC64- and -32-compiled 4.8.8, VisualC64-compiled 4.7.9, and MingW32-compiled 4.8.8 and 4.7.0, and I did not experience the described error on any of them.

This leads me to understand that MingW64-compiled Gambit has a problem on Windows 10, and that probably MingW64-compiled Gambit has a problem on older Windows versions also, however this would need to be tested.

Next debugging steps would be:
 * Compile Gambit on MingW64 with debugging symbols enabled, to get a stack trace / detail debugging info of the error
 * Try MingW64-compiled Gambit on previous Windows versions.

So this issue needs followup to resolve the problem.

feeley:
I believe that's an old bug (repaired since the latest release).  Will check if the latest sources also have it.

budden:
Hi! It looks like it is a general problem. Similar issue for ClozureCL : 
https://github.com/Clozure/ccl/issues/24

Even Excel seem to suffer from similar issues for a long time: 
https://www.kapilarya.com/microsoft-excel-2016-has-stopped-working-on-windows-10

kflu:
I had the same problem. I reported it here https://github.com/gambit/gambit/issues/278

junjiemars:
![image](https://user-images.githubusercontent.com/1978181/37702511-be0c837c-2d2d-11e8-819b-ecbc53c9cc6d.png)

adam-7:
@junjiemars , exactly what debugging tool are you using there?

feeley:
[...]

PR #601, opened 2020-06-11
Re-enable `path-type: inherit` in MSYS config for MinGW builds

daviwil:
I noticed some errors about `findstr` not being found in MinGW builds, turns out this is because `gambuild-C` on Windows uses `findstr` to extract meta info from modules.  I'm turning `path-type: inherit` back on in MinGW builds so that Windows commands will be accessible in those builds.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -35,6 +35,7 @@ jobs:
         update: true
         cache: true
         install: "autoconf git make tar texinfo mingw-w64-${{ matrix.arch }}-gcc mingw-w64-${{ matrix.arch }}-make mingw-w64-${{ matrix.arch }}-libwinpthread-git"
+        path-type: inherit
         msystem: ${{ matrix.msystem }}
 
     - name: Build
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -160,12 +160,14 @@ jobs:
     - name: Build and Test
       run: |
         export CC=gcc-9
+        export CXX=g++-9
         ./configure --enable-debug --enable-multiple-threaded-vms && make clean && make -j4 && make check && make clean && ./configure --enable-debug --enable-multiple-threaded-vms --enable-cplusplus && make -j4 && make check && make clean && ./configure --enable-ansi-c && make -j4 && (cd tests; make test1) && (cd tests; make test2) && (cd tests; make test3) && (cd tests; make test4) && (cd tests; make test5)
 
     - name: Build Gambit for Use
       run: |
         mkdir dist
         export CC=gcc-9
+        export CXX=g++-9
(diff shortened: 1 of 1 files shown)


daviwil:
@feeley This change does seem to improve the issue I mentioned here: https://github.com/gambit/gambit/issues/594#issuecomment-642679965

This is a successful test run on `master` of my fork: https://github.com/daviwil/gambit/runs/762575593?check_suite_focus=true

daviwil:
Just noticed that `clang` was being used for the C++ compiler on macOS builds, added the `CXX` environment variable to rectify that.

daviwil:
Ouch, looks like that `make ut` issue is still present here.  It seems very sporadic.

feeley:
For the sporadic executions, could you try to execute a `(read-line (open-process "dir"))` in a loop to see if that is the problem?  Possibly it is only a problem on the CI machine.

daviwil:
I ran this script through `gsi` via a local build in `mingw64` on my machine and it worked fine:

```scheme
(let loop ()
  (println "Testing...")
  (println (read-line (open-process "dir")))
  (loop))
```

daviwil:
Actually, I'm not synced up to latest `master` in my local branch, let me rebuild and try that again.

daviwil:
Ran it again after rebuilding from `master` and it seems to work fine, though I did notice this error a couple times (2 out of 7 runs I'd say) when I `C-c` out of the script:

```
Testing...
dir: write error
```

I suppose that could just be from killing the script while the port is still active.

feeley:
@daviwil The merge failed... can you update your PR?

daviwil:
Done!

feeley:
The comment "Avoid a segfault caused by GCC optimizations" is worrisome...  what version of gcc is causing problems? Can we upgrade to a newer version?

daviwil:
The latest version of GCC on MSYS2/MinGW64 10.1.0, that's what we're using in these builds.  @FredericHamel and I tried all of the optimization flags individually and couldn't figure out which one in particular was causing the issue.  Specifying all the known optimization flags also doesn't cause an issue, it only occurs when you specifiy `-O1` or higher.  Very strange issue.

daviwil:
The segmentation fault issue I ran into was very similar to the one reported in issue #320.  It seems that some optimization is affecting stack frames in a way that crashes `gsc`/`gsi` sporadically.  This is the error that I would see when it happened:

https://github.com/gambit/gambit/issues/320#issuecomment-381115963

daviwil:
Removed the `-O0` flag and rebased on your latest commit, let's see what happens in CI.

daviwil:
With the `setjmp` fix and removal of `-O0` I think this PR is ready to go!

PR #604, opened 2020-06-12
Store output of 'make dist-boot' as artifact, build with processor count

daviwil:
This change is the first step toward faster PR builds.  It uploads the output of `make dist` as an artifact called `gambit-dist.tgz` that can be downloaded from the latest successful CI build on `master` and applied to `boot/` in PR runs.  I'm almost done updating [`daviwil/download-gambit`](https://github.com/daviwil/download-gambit) to support downloading this artifact.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -155,7 +155,7 @@ jobs:
         cd ..
 
     - name: Package Source
-      run: make dist; find ./ -name 'gambit-v*.tgz' -exec cp {} gambit.tgz \;
+      run: make dist && find ./ -name 'gambit-v*.tgz' -exec cp {} gambit.tgz \;
 
     - name: Upload Build Artifact
       uses: actions/upload-artifact@v1
@@ -166,7 +166,7 @@ jobs:
     - name: Upload Source
       uses: actions/upload-artifact@v1
       with:
-        name: gambit-linux-x86_64-dist
+        name: gambit-dist
         path: gambit.tgz
 
     - name: Report Failure to Gitter
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -94,7 +94,7 @@ jobs:
       run: |
         call "C:\Program Files (x86)\Microsoft Visual Studio\2019\Enterprise\VC\Auxiliary\Build\vcvarsall.bat" x64 -vcvars_ver=14.16
         set PATH=%RUNNER_TEMP%\msys\msys64\usr\bin;%PATH%
-        msys2 ./configure --enable-c-opt=-Od --enable-debug CC="cl" && make -j`nproc` && make modules
+        msys2 ./configure --enable-c-opt=-Od --enable-debug CC="cl" && make -j%NUMBER_OF_PROCESSORS% && make modules
 
     - name: Test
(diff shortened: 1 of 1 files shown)


feeley:
With commit 551edec8cb3be3fc993e83579df8e80495d2ff63 you can now do `make dist-boot` to create `boot.tgz` which can be unpacked to create the `boot` directory.  That will simplify the bootstrap procedure and is a logical name for that artifact (as suggested by @lassik).

daviwil:
Now that I've got a successful build with the new `boot.tgz`, I'll test a PR workflow that uses it to speed up the build.  Let's hold off merging this until I've verified that it works.

lassik:
> With commit [551edec](https://github.com/gambit/gambit/commit/551edec8cb3be3fc993e83579df8e80495d2ff63) you can now do `make dist-boot` to create `boot.tgz` which can be unpacked to create the `boot` directory.

Thank you very much for making this!

daviwil:
Whoa, I dunno what magit did, but it closed this PR on my behalf.  Can you reopen it @feeley?

daviwil:
Phew, OK, I was able to reopen it once I fixed the branch.  I dunno what keybinding I hit to make it delete my branch and push it immediately!

daviwil:
@feeley So somehow the culprit is having `PATH` inheritance turned on in msys2.  When I turn that off, I see a lot of messages related to `findstr` any time `gambuild-C.bat` is executed:

```
------------ TEST 6 (link and execute the code generated)
rm -f mix_.c mix.o mix_.o mix.exe
../gsc/gsc.exe -:~~bin=../bin,~~lib=../lib,~~include=../include -f -warnings -c mix.scm
../gsc/gsc.exe -:~~bin=../bin,~~lib=../lib,~~include=../include -f -warnings -o mix.exe -exe mix.c
'findstr' is not recognized as an internal or external command, operable program or batch file.
'findstr' is not recognized as an internal or external command, operable program or batch file.
'findstr' is not recognized as an internal or external command, operable program or batch file.
'findstr' is not recognized as an internal or external command, operable program or batch file.
'findstr' is not recognized as an internal or external command, operable program or batch file.
```

However, it seems that _not_ finding `findstr` prevents the failures from occuring in `make ut`.  I'll keep investigating why this is happening but I think for this PR is safe to merge, provided that you think the `#undef HAVE_CLOCK_GETTIME` approach is OK for removing the `libwinpthread-1.dll` dependency.

feeley:
I have merged the PR so we can move forward.  However the `#undef HAVE_CLOCK_GETTIME` is a bit hackish, so we should find a better place to disable it (`configure` is the right place but what should be detected to disable it?).  Also, I would expect that for a typical developer on Windows the PATH is inherited, so we should eventually execute the CI with inheritance when the issue is fully understood and resolved.

daviwil:
We could probably just `#undef` it after we see `$ENABLE_THREAD_SYSTEM` equal to `win32` in `configure` since that should be fair evidence that the user doesn't want to link to `libwinpthread`.  I'll experiment with a fix.

Regarding the path inheritance issue, I suspect it's some other kind of environmental issue that's causing unexpected behavior, so I'll keep poking at that problem here and there when I have time.

PR #606, opened 2020-06-13
Use gcc __builtin_setjmp when available

feeley:


feeley:
This seems to have fixed the testing issues with MINGW.  I hope it is not a fluke that there were no issues... time will tell!

daviwil:
It very well might have.  One thing I realized yesterday is that I wasn't doing the `-O0` hack on the initial MinGW build that gets used for the test run, so it's possible the same issue was showing up there too (even more likely because the issue seemed intermittent).

daviwil:
My local MinGW64 build seems to be working perfectly after this change!  I would always see a segfault when running `make modules` after `make` and this time it hasn't happened.

feeley:
I just hope using `__builtin_setjmp` won't cause problems further down the line... The documentation says it is not a drop in replacement for `setjmp` and is used internally by the C lib.  But Gambit only uses `setjmp` for non-local exit, and in only one place (the macro `___BEGIN_TRY` which emulates a `try-catch` form), so I'm confident it won't have ill effects.  The only case which worries me is linking to C libraries that use the "real" `setjmp`.  But if this is ever a problem it is possible to add `-D___DONT_USE_BUILTIN_SETJMP` to `CFLAGS` to use the real `setjmp`.

daviwil:
Hmmm, I just saw this when running `make check`:

```
make[1]: Entering directory '/c/dev/gambit/tests'
------------ TEST 1 (debugging support)
../gsi/gsi.exe -:~~bin=../bin,~~lib=../lib,~~include=../include -f debug.scm > test1.out
/bin/sh: line 1:  6302 Segmentation fault      ../gsi/gsi.exe -:~~bin=../bin,~~lib=../lib,~~include=../include -f debug.scm > test1.out
make[1]: *** [makefile:118: test1] Error 139
make[1]: Leaving directory '/c/dev/gambit/tests'
make: *** [makefile:208: check] Error 2
```

Here's what gdb says, seems to be a different stack trace than the original issue:

```
(gdb) run
Starting program: C:\dev\gambit\gsi\gsi.exe "-:~~bin=../bin,~~lib=../lib,~~include=../include" -f debug.scm
[New Thread 4316.0x1214]
[New Thread 4316.0x1d78]
[New Thread 4316.0x12a0]

Thread 1 received signal SIGSEGV, Segmentation fault.
0x00000000005b74d3 in ___H___io ()
(gdb) bt
#0  0x00000000005b74d3 in ___H___io ()
#1  0x000000000040afa0 in ___trampoline ()
#2  0x000000000040b106 in ___call ()
#3  0x000000000040b390 in ___run ()
#4  0x000000000040cb18 in ___setup ()
#5  0x0000000000434f5c in ___main ()
#6  0x000000000041eb2b in ___winmain ()
#7  0x00000000004073cc in WinMain ()
#8  0x00000000004013c4 in __tmainCRTStartup ()
    at D:/mingwbuild/mingw-w64-crt-git/src/mingw-w64/mingw-w64-crt/crt/crtexe.c:335
#9  0x0000000000401506 in mainCRTStartup ()
    at D:/mingwbuild/mingw-w64-crt-git/src/mingw-w64/mingw-w64-crt/crt/crtexe.c:219
(gdb)
```

feeley:
With `-O0`?

daviwil:
That's without `-O0`.  I'll try it again with it to see if I get a different result.  I wouldn't worry too much about this new segfault now though, it could just be a problem with my local build.

Looks like CI on master failed in unit tests again, so maybe this wasn't the silver bullet for that particular issue :(  However, the compiler definitely seems a lot more stable now without `-O0`.

daviwil:
The `___H___io` crash happens with `-O0` also so it seems to be a separate issue entirely.  I probably need to clean my repo and build completely from scratch to ward off any other potential causes.  Let's call it a non-issue for now!

feeley:
It is baffling that on the PR that uses `__builtin_setjmp` the unit tests succeeded on MINGW, but on the merge with master they fail again (albeit at the third test rather than the usual first).  Moreover it seems to happen only on the CI machine.  How can this be debugged?

@daviwil can you try to reproduce failing `make ut` on another Windows machine?  I suspect there is a race condition when exiting the program (perhaps the auxiliary threads are not terminating cleanly).  One avenue to explore it to avoid linking with pthreads and only use Win32 threads (I don't remember which one is linked on the CI machines).

daviwil:
Yep, I'll see if I can produce a build on CI that fails but still produces an artifact so that I can experiment with it locally. I'll look into how I can force Win32 threads instead of pthread, or if it's possible in MinGW

daviwil:
OK, I figured out I can use `--enable-thread-system=win32` to force Win32 threads in MinGW.  `make ut` fails at the same place consistently (but not that same as on CI!) both with `win32` threads and `posix` threads:

```
[161|  0]  64% ##########...... 19.0s 05-serdes\serdes.scm
*** FAILED 05-serdes\serdes.scm WITH EXIT CODE HI=5 LO=0
#!eof
[161|  1]  64% ##########...... 19.1s 06-thread\current_processor.scm
```

I'm not able to reproduce the crash we see on CI, but we'll see how it goes with the `--enable-thread-system=win32` parameter I just added.

Side note, it seems that we still end up linking to `libwinpthread-1.dll` because GCC causes it to be included by default.  More info here: https://sourceforge.net/p/msys2/mailman/message/35020436/

daviwil:
OK, scratch that `make ut` error, after I re-bootstrapped my local repo, both `make check` and `make ut` work just fine.

daviwil:
I downloaded an artifact from [this CI run](https://github.com/gambit/gambit/runs/768780984?check_suite_focus=true) and tried running tests with it on my local machine, it worked fine within a `mingw64` terminal across multiple runs but I hit issues where it expected `libwinpthread-1.dll` when running from `cmd.exe` (which is strange because it was built with `-static`).

The conclusion is that this does only seem to be occurring on CI for some reason.  The next thing I'll try is to try running the tests in `gdb`'s batch mode on CI so that we can see where exactly it fails.  Probably won't get to do that until tomorrow.

feeley:
Ok.  Can't wait to have this mystery resolved.

daviwil:
OK, ran the unit tests through `gdb` in CI, got this output:

```
mingw32-make[2]: Entering directory 'D:/a/gambit/gambit/tests'
gdb --batch --ex=run --ex=bt --args ../gsi/gsi.exe -:tl,~~bin=../bin,~~lib=../lib,~~include=../include -f ./run-unit-tests.scm
[New Thread 6208.0x6b0]
[New Thread 6208.0x210]
[New Thread 6208.0x1744]
[New Thread 6208.0x1b00]
[New Thread 6208.0x1918]
[  0|  0]   0% ................  .0s 00-check\check_eq.scm[Thread 6208.0x1918 exited with code 3]
[Thread 6208.0x1b00 exited with code 3]
[Thread 6208.0x1744 exited with code 3]
[Inferior 1 (process 6208) exited with code 03]
No stack.
```

Do you know why the unit test would return an exit code of `3`?

feeley:
I wonder if this is related to the linking with `libwinpthread-1.dll`... If it is linked due to the use of `clock_gettime`, you might want to force that it is not used (by adding `#undef HAVE_CLOCK_GETTIME` at the end of `include/config.h` after `./configure ...`).

Note that 3 threads are terminating with code 3 (I don't know what this code means on Windows).  But these are probably the "pump" threads for stdin/stdout/stderr (Gambit uses these threads to implement nonblocking I/O on Windows).  Look for `___device_stream_pump_setup` in `lib/os_io.c`.  I think this may be a termination of these threads when the program terminates (the Gambit RTS attempts to terminate the threads cleanly with a `___KILL_PUMP` message, but the C lib probably kills the threads forcibly if they don't have enough time to terminate cleanly).  Also note that the pump threads are Win32 threads created with `CreateThread`... could this cause problems if combined with the pthread lib?  The correct operation of the MSVC build (which only uses Win32 threads) should tell us something...

daviwil:
Turns out we see the same exit code when using pthread too, but that's probably just because it's a wrapper for Win32 threads.  MSDN docs are pretty useless on the topic, no indication of what that error code means.  I'm hoping this isn't a red herring because that appeared because I'm trying to run it through GDB; it stops on the first test rather than on the third as it normally does, so I'm skeptical.

I'll keep poking around and try to find another way to debug the issue.  I'll also try the `#undef HAVE_CLOCK_GETTIME` approach you mentioned.

daviwil:
It seems that `#undef HAVE_CLOCK_GETTIME` approach does cause the `libwinpthread-1.dll` dependency to go away:

```
$ ldd bin/gsi.exe
        ntdll.dll => /c/Windows/SYSTEM32/ntdll.dll (0x7fff64160000)
        KERNEL32.DLL => /c/Windows/System32/KERNEL32.DLL (0x7fff62c50000)
        KERNELBASE.dll => /c/Windows/System32/KERNELBASE.dll (0x7fff61360000)
        msvcrt.dll => /c/Windows/System32/msvcrt.dll (0x7fff63960000)
        USER32.dll => /c/Windows/System32/USER32.dll (0x7fff62e60000)
        win32u.dll => /c/Windows/System32/win32u.dll (0x7fff61330000)
        GDI32.dll => /c/Windows/System32/GDI32.dll (0x7fff640f0000)
        gdi32full.dll => /c/Windows/System32/gdi32full.dll (0x7fff61f20000)
        msvcp_win.dll => /c/Windows/System32/msvcp_win.dll (0x7fff61e80000)
        ucrtbase.dll => /c/Windows/System32/ucrtbase.dll (0x7fff61230000)
        WS2_32.dll => /c/Windows/System32/WS2_32.dll (0x7fff62600000)
        RPCRT4.dll => /c/Windows/System32/RPCRT4.dll (0x7fff62290000)
```

Would there be any negative impact to using that in Gambit builds?  If the Win32 thread support is sufficient I think it'd be great if we didn't need to reference `libwinpthread-1.dll` at all.

daviwil:
Quick update: I'm still investigating this, now trying to go back to earlier commits to see if there was a particular change that invoked the problem.  I'm still not able to reproduce the issue on my local machine so barring a discovery from my current CI run experimentation, I'll have to narrow down a minimal repro on CI so that we can instrument it somehow.

feeley:
Perhaps you should look at the `PATH` of the CI to see if it is mapping some of the DLLs differently than you expect.  If `ldd` is sensitive to the `PATH`, you should try on the CI and on your machine to see what is the diff.

daviwil:
Good idea, I'll give that a try too.

feeley:
While you are at it can you also print the processor model in case that has an effect.  There has to be something that is different between the machines where `make ut` fails and doesn't.  We should also execute the `make ut` 20 times in a row to see if consistently, for a particular CI run, all the `make ut` either fail or don't.

daviwil:
Because we only started to see regular failures after #597 was merged, I created a couple branches from the commits right before that (c62cc97 - [CI run](https://github.com/daviwil/gambit/runs/785121866?check_suite_focus=true), 2dd5f6e - [CI run](https://github.com/daviwil/gambit/runs/785477172) to see if I could invoke the failure.  I ran them both 3 times each and didn't see the error.

I'm now running a build from a branch created from the merge commit for #597 with the CI configuration updated to give us the information you asked about.  It failed with the error on the first try:

https://github.com/daviwil/gambit/runs/788265272?check_suite_focus=true

Does your intuition tell you anything about how PR #597 could have caused this error?  Is there anything I can tweak from the changes in that PR to see if it improves the situation?

Output from `cat /proc/cpuinfo`:

```
processor	: 0
vendor_id	: GenuineIntel
cpu family	: 6
model		: 63
model name	: Intel(R) Xeon(R) CPU E5-2673 v3 @ 2.40GHz
stepping	: 2
microcode	: 0x0
cpu MHz		: 2400.000
cache size	: 30720 KB
physical id	: 0
siblings	: 2
core id		: 0
cpu cores	: 2
apicid		: 0
initial apicid	: 0
fpu		: yes
fpu_exception	: yes
cpuid level	: 14
wp		: yes
flags		: fpu vme de pse tsc msr pae mce cx8 apic sep mtrr pge mca cmov pat pse36 clflush mmx fxsr sse sse2 ss ht pni pclmuldq ssse3 fma cx16 pcid sse4_1 sse4_2 movbe popcnt aes xsave avx f16c rdrand hypervisor lahf_lm fsgsbase bmi1 avx2 smep bmi2 erms invpcid xsaveopt md_clear
bogomips	: 4800.00
clflush size	: 64
cache_alignment	: 64
address sizes	: 44 bits physical, 48 bits virtual
power management:

processor	: 1
vendor_id	: GenuineIntel
cpu family	: 6
model		: 63
model name	: Intel(R) Xeon(R) CPU E5-2673 v3 @ 2.40GHz
stepping	: 2
microcode	: 0x0
cpu MHz		: 2400.000
cache size	: 30720 KB
physical id	: 0
siblings	: 2
core id		: 1
cpu cores	: 2
apicid		: 1
initial apicid	: 1
fpu		: yes
fpu_exception	: yes
cpuid level	: 14
wp		: yes
flags		: fpu vme de pse tsc msr pae mce cx8 apic sep mtrr pge mca cmov pat pse36 clflush mmx fxsr sse sse2 ss ht pni pclmuldq ssse3 fma cx16 pcid sse4_1 sse4_2 movbe popcnt aes xsave avx f16c rdrand hypervisor lahf_lm fsgsbase bmi1 avx2 smep bmi2 erms invpcid xsaveopt md_clear
bogomips	: 4800.00
clflush size	: 64
cache_alignment	: 64
address sizes	: 44 bits physical, 48 bits virtual
power management:
```

Output of `$PATH` and `ldd`:

```
PATH: /mingw64/bin:/usr/local/bin:/usr/bin:/bin:/usr/bin/site_perl:/usr/bin/vendor_perl:/usr/bin/core_perl

ldd for gsc.exe
	ntdll.dll => /c/windows/SYSTEM32/ntdll.dll (0x7ffcc6770000)
	KERNEL32.DLL => /c/windows/System32/KERNEL32.DLL (0x7ffcc5ab0000)
	KERNELBASE.dll => /c/windows/System32/KERNELBASE.dll (0x7ffcc2900000)
	msvcrt.dll => /c/windows/System32/msvcrt.dll (0x7ffcc6110000)
	USER32.dll => /c/windows/System32/USER32.dll (0x7ffcc6530000)
	win32u.dll => /c/windows/System32/win32u.dll (0x7ffcc2830000)
	GDI32.dll => /c/windows/System32/GDI32.dll (0x7ffcc60e0000)
	libwinpthread-1.dll => /mingw64/bin/libwinpthread-1.dll (0x64940000)
	gdi32full.dll => /c/windows/System32/gdi32full.dll (0x7ffcc2bd0000)
	msvcp_win.dll => /c/windows/System32/msvcp_win.dll (0x7ffcc3790000)
	ucrtbase.dll => /c/windows/System32/ucrtbase.dll (0x7ffcc3830000)
	WS2_32.dll => /c/windows/System32/WS2_32.dll (0x7ffcc66d0000)
	RPCRT4.dll => /c/windows/System32/RPCRT4.dll (0x7ffcc6320000)

ldd for gsi.exe
	ntdll.dll => /c/windows/SYSTEM32/ntdll.dll (0x7ffcc6770000)
	KERNEL32.DLL => /c/windows/System32/KERNEL32.DLL (0x7ffcc5ab0000)
	KERNELBASE.dll => /c/windows/System32/KERNELBASE.dll (0x7ffcc2900000)
	msvcrt.dll => /c/windows/System32/msvcrt.dll (0x7ffcc6110000)
	USER32.dll => /c/windows/System32/USER32.dll (0x7ffcc6530000)
	win32u.dll => /c/windows/System32/win32u.dll (0x7ffcc2830000)
	GDI32.dll => /c/windows/System32/GDI32.dll (0x7ffcc60e0000)
	libwinpthread-1.dll => /mingw64/bin/libwinpthread-1.dll (0x64940000)
	gdi32full.dll => /c/windows/System32/gdi32full.dll (0x7ffcc2bd0000)
	msvcp_win.dll => /c/windows/System32/msvcp_win.dll (0x7ffcc3790000)
	ucrtbase.dll => /c/windows/System32/ucrtbase.dll (0x7ffcc3830000)
	WS2_32.dll => /c/windows/System32/WS2_32.dll (0x7ffcc66d0000)
	RPCRT4.dll => /c/windows/System32/RPCRT4.dll (0x7ffcc6320000)
```

feeley:
> Because we only started to see regular failures after #597 was merged, I created a couple branches from the commits right before that ([c62cc97](https://github.com/gambit/gambit/commit/c62cc970ded5b1047f76ba34b2ed05fc90bfcee8) - [CI run](https://github.com/daviwil/gambit/runs/785121866?check_suite_focus=true), [2dd5f6e](https://github.com/gambit/gambit/commit/2dd5f6ed72a7c4274c74499a8151b6da7cd83e24) - [CI run](https://github.com/daviwil/gambit/runs/785477172) to see if I could invoke the failure. I ran them both 3 times each and didn't see the error.
> 
> I'm now running a build from a branch created from the merge commit for #597 with the CI configuration updated to give us the information you asked about. It failed with the error on the first try:
> 
> https://github.com/daviwil/gambit/runs/788265272?check_suite_focus=true
> 

Could you run the `make ut` 20 times while ignoring any failure?  That way we can see if there is some randomness to the failures.

> Does your intuition tell you anything about how PR #597 could have caused this error? Is there anything I can tweak from the changes in that PR to see if it improves the situation?

Nothing in that commit changed anything fundamental.  It is just Scheme code.  The only thing I can think of is memory alignment issues.  The code and possibly some of the static data will be at a different address from one build to the next due to the fact that the code changes.

Could you take one of the working commits and add a random innocuous change to see if it triggers a failure?  For example, to `lib/_kernel.scm` just before the line `(define-prim (##dynamic-env-bind denv thunk)` add

    (define (##mingw) "issue 606")

Another line of thought... how come only the `make ut` fails?  For the other tests of the `make check` the interpreter seems to work fine.  One difference is that `make ut` starts the interpreter to run `tests/run-unit-tests.scm` and it then creates one subprocess executing `gsi` for each unit test.  So I suspect it may be a problem in the implementation of `open-process`.  I also suspect that there is some randomness in this too (because it is not always the first unit test that fails).  Could the libwinpthread DLL be the source of the problem (it is still being linked)?  Running the `make ut` 20 times ignoring failures will help determine the degree to which failures are random.

> 
> Output from `cat /proc/cpuinfo`:

All of the processor information looks OK, but we should keep it around to compare with CI runs where there is no failure.

daviwil:
Whoops, forgot to add the multiple runs of `make ut`.  I'll add the suggested change to a working commit too just to see what happens.  Regarding `libwinpthread`, I was able to successfully remove that dependency in PR #604 but still saw the failure, so I don't think it's the culprit (for now).

feeley:
`libwinpthread` may not be the culprit, but it shouldn't be there because it requires the end-user to have mingw64 installed.  All the other dependencies appear to be "standard" Windows DLLs (at least I hope that's the case).

PR #608, opened 2020-06-15
Breakout _std.scm into modules

feeley:


feeley:
Yes that will help make the RTS a bit more modular... but there are other parts of the RTS that will also benefit from being modularized.

PR #614, opened 2020-06-23
Universal backend: add g_host2foreign(val) and g_foreign2host(obj)

feeley:


feeley:
@daviwil I was surprised to see the Windows CI fail quickly and I assume it is a bug in the Windows CI caused by the recent change.

daviwil:
Argh, looks like something broke with the MSYS2 action we use.  I'll look into it.

daviwil:
Filed this issue on `msys2/setup-msys2`: https://github.com/msys2/setup-msys2/issues/33

feeley:
I dislike *enormously* to be dependent on other things (libs, etc)... can we roll our own?

daviwil:
Technically yes since GitHub Actions recently started including MSYS2 on their base images, but I've found it to be much slower to install packages on and update the built-in MSYS2 (that's why MSYS2 setup used to take around 10 minutes).  I can look into maintaining our own solution for speeding it up but it's pretty complicated to do that fast and reliably.

The maintainer of `msys2/setup-msys2` is very responsive so I don't think this issue will linger for long.  However, if we continue to get broken by it over the next month we can try to come up with a better solution for our purposes.  Does that sound OK?

feeley:
Can we select a version of `msys2/setup-msys2` that "works" and just stay with that version?  Perhaps upgrade to a new version on a yearly basis rather than being dependent on its HEAD...

daviwil:
Based on what I see in the error message, I believe the problem is more likely caused by an update to a package in MSYS2, not the `setup-msys2` action itself.  Nothing substantive has changed in the action's code in the last couple of days.

If at some point we are being broken by changes they make, the simplest thing to do would be fork their repo and maintain our own release tag that we only update at some longer cadence, like you suggested.

daviwil:
Turns out it was an issue with MSYS2:
- https://github.com/msys2/setup-msys2/issues/33#issuecomment-649297277
- https://github.com/msys2/MSYS2-packages/issues/2021

Looks like it's working correctly in Gambit's latest PR #615.

PR #617, opened 2020-06-27
Fix crash when module is empty

feeley:
Fixes bug introduced in 98373cf53ca8d77b90e6559afe0da8032121ba07


daviwil:
Awesome, thanks! I'll give this a try today.

PR #621, opened 2020-07-08
Adapt unit tests to tolerate --enable-char-size=1 or 2

feeley:


gambiteer:
Perhaps you'll need to take out the über again in `_num.scm`.

feeley:
No that should be fine as `(char->integer #\ü)` = 252, which fits even in 8 bits!

PR #628, opened 2020-07-22
Add functions to get UDP socket receive and send buffer sizes

gcartier:


lib/os_io.c
```diff
@@ -11214,6 +11214,58 @@ ___SCMOBJ port_num;)
...
+#else
+
+  ___device_udp *d =
+    ___CAST(___device_udp*,___FIELD(dev,___FOREIGN_PTR));
+
+  int optVal;
+  SOCKET_LEN_TYPE optLen = sizeof(optVal);
+
+  optLen = sizeof(optVal);
+  getsockopt(d->s, SOL_SOCKET, SO_RCVBUF, (char*) &optVal, &optLen);
```
feeley, line 11235: Does this work on Windows?  I'm worried about the portability of `SO_RCVBUF`.  Also, it is best to use `___CAST(char*,...)` rather than `(char*)...`.

lib/os_io.c
```diff
@@ -11214,6 +11214,58 @@ ___SCMOBJ port_num;)
...
+#else
+
+  ___device_udp *d =
+    ___CAST(___device_udp*,___FIELD(dev,___FOREIGN_PTR));
+
+  int optVal;
+  SOCKET_LEN_TYPE optLen = sizeof(optVal);
+
+  optLen = sizeof(optVal);
+  getsockopt(d->s, SOL_SOCKET, SO_SNDBUF, (char*) &optVal, &optLen);
```
feeley, line 11261: Same comment as above.

lib/os_io.c
gcartier: It does work on Windows

PR #630, opened 2020-07-29
CI: Update to msys2/setup-msys2@v2

lazka:
Caching is now on by default and running inline scripts requires using '-c'


feeley:
Thanks for the PR!

daviwil:
Thanks @lazka!

Issue #512, opened 2020-03-25
assuming every macOS user uses brew for openssl is not OK


jxy:
I guess `configure` wants to be helpful and give sensible default, but practically I just see a lot of warnings from `gsc`,

```
ld: warning: directory not found for option '-L/usr/local/opt/openssl@1.1/lib'
```

which I found out was hardcoded here,

https://github.com/gambit/gambit/blob/c584bbcf6d9c66530dd3f9b981726f22a5973fde/configure#L9833-L9835

The simple thing would be to check if the directory exists before including in `LIBS`.

For me I'm just using,
```
if test "$ENABLE_OPENSSL" = yes; then
  LIBS="$LIBS -lssl -lcrypto"
fi
```
because I pass `CFLAGS=-I/opt/local/include LIBS=-L/opt/local/lib` to configure anyway.

lassik:
This is complicated by the fact that some OpenSSL versions shipped by operating systems are outdated. Another complication is that the `openssl` shell command may belong to a different installation of OpenSSL/LibreSSL than the `libssl` library, even on the same computer. Hence it's not reliable to run the shell command to infer things about the library.

Please see my notes at https://misc.lassi.io/2019/chicken-openssl/chicken-openssl-pkg-config.text

lassik:
For MacOS in particular, last I checked it doesn't ship pkg-config `.pc` files for the system OpenSSL/LibreSSL. Neither do some other operating systems. That's a pity.

Given the lack of `.pc` files, and the possibly outdated versions, it would be easier to rely on Homebrew OpenSSL/LibreSSL on Mac. MacOS versions of open source libraries tend to lag behind the upstream ones enough that programming without Homebrew is a bit difficult in general. Is there a particular reason you'd like to avoid using Homebrew packages?

gambiteer:
On 3/25/20 3:29 AM, Lassi Kortela wrote:
> For MacOS in particular, last I checked it doesn't ship pkg-config |.pc| 
> files for the system OpenSSL/LibreSSL. Neither do some other operating 
> systems. That's a pity.
> 
> Given the lack of |.pc| files, and the possibly outdated versions, it 
> would be easier to rely on Homebrew OpenSSL/LibreSSL on Mac. MacOS 
> versions of open source libraries tend to lag behind the upstream ones 
> enough that programming without Homebrew is a bit difficult in general. 
> Is there a particular reason you'd like to avoid using Homebrew packages?

[...]

PR #638, opened 2020-09-26
--enable-openssl on Darwin, more suitable

junjiemars:
#512

1. Check `OPENSSL_INC_DIR` and `OPENSSL_LIB_DIR` first, if it's empty then pick `pkg-config`;
2. If `pkg-config` no found or no works for `openssl`, keep `OPENSSL_INC_DIR` or `OPENSSL_LIB_DIR` empty;

### DIFF
--- a/configure
+++ b/configure
@@ -9876,13 +9876,14 @@ if test "$ENABLE_OPENSSL" = yes; then
   case "$target_os" in
     darwin*) # macOS ships an old version of the OpenSSL library, so it should
              # be installed with "brew install openssl"
-             OPENSSL_DIR="/usr/local/opt/openssl@1.1"
+             OPENSSL_INC_DIR="${OPENSSL_INC_DIR:-`pkg-config --cflags-only-I openssl`}"
+             OPENSSL_LIB_DIR="${OPENSSL_LIB_DIR:-`pkg-config --libs-only-L openssl`}"
              if test "$ENABLE_OPENSSL_STATIC_LINK" = yes; then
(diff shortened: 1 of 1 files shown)


lassik:
On mac, if pkg-config doesn't find the library, the configure script could recommend the user to try the following:

```
export PKG_CONFIG_PATH="$(brew --prefix openssl)/lib/pkgconfig"
export PKG_CONFIG_PATH="$(brew --prefix libressl)/lib/pkgconfig"
```

pkg-config `.pc` definition files installed by Homebrew packages are not on the default `PKG_CONFIG_PATH`.

junjiemars:
1. If there are no `brew` exists, what can we do. So, just focus on the system common environment not the  special cases, here the `pkg-config` is the common case, and on the other size, `brew` or `port` are special cases. 
2. If `openssl` can not be found, the C compiler should mention it.

lassik:
The common environment for SSL (using only libraries that come with the OS, and commands that work out-of-the-box on most operating systems) is quite poor. We need to cover the special cases for a useful result. Here's the easiest solution for each OS: https://misc.lassi.io/2019/chicken-openssl/chicken-openssl-pkg-config.text You'll notice that it's full of special cases and warnings.

Gambit probably doesn't need to use the `openssl` shell command at all. It should be enough to use the C library only. Note that the `openssl` command can belong to a different version of OpenSSL than the openssl library; it's possible to have more than one version of OpenSSL installed at once. On macOS, for example, `openssl` is usually `/usr/bin/openssl` which belongs to the (often out-of-date) version of OpenSSL (or LibreSSL) that ships with macOS.

PR #651, opened 2021-01-04
Srfi 60 : Integers as Bits

antoine-doucet:


gambiteer:
Thank you for this pull request.
Perhaps I don't quite understand all the code, but it appears that most routines defined in this code are already defined in `gambit/lib/_num.scm`.  I went through the list of routines in this SRFI and put a * next to every routine I found in `_num.scm`:
```
(##namespace ("srfi/60#"
    bitwise-and             ;; *
    bitwise-ior             ;; *
    bitwise-xor             ;; *
    bitwise-not             ;; *
    bitwise-merge           ;; *
    any-bits-set?           ;; *
    bit-count               ;; *
    integer-length          ;; *
    first-set-bit
    bit-set?                ;; *
    copy-bit                ;; *
    bit-field
    copy-bit-field          ;; *
    rotate-bit-field
    reverse-bit-field
    integer->list
    list->integer
    booleans->integer
```
It seems that only seven new routines need to be coded.  I'd be willing to advise on how to do this using Gambit's fixnum/bignum implementation.

Brad

gambiteer:
Thank you for this pull request.
Perhaps I don't quite understand all the code, but it appears that most routines defined in this code are already defined in `gambit/lib/_num.scm`.  I went through the list of routines in this SRFI and put a * next to every routine I found in `_num.scm`:
```
(##namespace ("srfi/60#"
    bitwise-and             ;; *
    bitwise-ior             ;; *
    bitwise-xor             ;; *
    bitwise-not             ;; *
    bitwise-merge           ;; *
    any-bits-set?           ;; *
    bit-count               ;; *
    integer-length          ;; *
    first-set-bit
    bit-set?                ;; *
    copy-bit                ;; *
    bit-field
    copy-bit-field          ;; *
    rotate-bit-field
    reverse-bit-field
    integer->list
    list->integer
    booleans->integer
```
It seems that only seven new routines need to be coded.  I'd be willing to advise on how to do this using Gambit's fixnum/bignum implementation.

Brad

antoine-doucet:
Indeed, most of the functionalities are already built-in, even including their bignum counterparts. I would be glad to implement the missing routines and will surely take some advice !

Antoine

gambiteer:
OK, well, I haven't looked closely at your code yet, but here are some thoughts.

`first-set-bit` appears to be a renaming of built-in `first-bit-set`.  I would let the existing routines implement the error checking, checking whether Gambit is built with bignum support, etc.

`bit-field` seems like Gambit's `extract-bit-field`, perhaps with a different order of arguments.

The SRFI 60 implementation of  `rotate-bit-field` looks straightforward using existing bulk operations as you have done, I believe (although I haven't checked it in detail).

I would suggest using the name of Gambit's built-in procedures rather than SRFI 60 names (e.g., `bitwise-ior` instead of `logior`) in the implementation of new procedures.

In `list->integer` I suppose you can have general code when bignums are available, but you'll need specialized code using `##fx*?` to multiply partial results by 2 so you can check whether the result can not be contained in a fixnum and signal an error.  See also next comment.

For routines like `bitwise-reverse` and `list->integer`, where you have to construct an integer bit-by-bit, the code should have two branches depending on whether bignums are supported or not.  You should look at how `macro-exact-int-dispatch` is defined in `lib/_num#.scm` and how it's used in `lib/_num.scm`.

Hope this helps for now.

Brad

gambiteer:
These comments are based on the assumption that you want these routines to be error checked as well as those in `_num.scm`.
1.  I don't think `(checked-bitwise-op call)` is general enough (these operations apply to bignums as well as fixnums), and I don't think it's needed (see below).
2.  I think built-in procedures like `bitwise-and` should just be imported into the SRFI 60 namespace, I don't think you should have a new definition.  I don't know how to do this (or, indeed, whether anything needs to be "done").
3. If you really want to support the complete SRFI, then I think you should have the alternate names `logand`, etc.  I'd model them on the definitions in `_num.scm`.  For example, you find in `_num.scm`
```
(define-prim-nary (bitwise-and x y)
  -1
  (if (macro-exact-int? x) x '(1))
  (##bitwise-and2 x y)
  macro-force-vars
  macro-no-check
  (##pair? ##fail-check-exact-integer))
```
This may not look like it takes more than 2 arguments, but it does because of the `define-prim-nary` macro.  So perhaps you should have in the SRFI 60 implementation of `logand` like this
```
(define-prim-nary (logand x y)
  -1
  (if (macro-exact-int? x) x '(1))
  (##bitwise-and2 x y)
  macro-force-vars
  macro-no-check
  (##pair? ##fail-check-exact-integer))
```
to have precise error messages (i.e., ones that mention `logand` as the name of the function). 
4.  Model `copy-bit-field` on the version in `_num.scm`, which would give something like
```
(include "~~/lib/_gambit#.scm")

(define-prim (copy-bit-field to from start end)
  (macro-force-vars (to from start end)
    (macro-check-index
      start
      3
      (copy-bit-field to from start end)
      (macro-check-index-range-incl
        end
        4
        start
        ##max-fixnum
        (copy-bit-field to from start end)
        (cond ((##not (macro-exact-int? from))
               (##fail-check-exact-integer 2 copy-bit-field to from start end))
              ((##not (macro-exact-int? to))
               (##fail-check-exact-integer 1 copy-bit-field to from start end))
              (else
               (##copy-bit-field (##fx- end start) start from to)))))))
```
which expands to
```
(define copy-bit-field
  (lambda (to from start end)
    (if ('#<procedure #2 ##fixnum?> start)
        (if ('#<procedure #3 ##not> ('#<procedure #4 ##fxnegative?> start))
            (if ('#<procedure #2 ##fixnum?> end)
                (if (and ('#<procedure #3 ##not> ('#<procedure #5 ##fx<> end start))
                         ('#<procedure #3 ##not> ('#<procedure #5 ##fx<> ##max-fixnum end)))
                    (if ('#<procedure #3 ##not>
                         (if ('#<procedure #2 ##fixnum?> from) #t (if ('#<procedure #6 ##bignum?> from) #t #f)))
                        (##fail-check-exact-integer 2 '#<procedure #7 copy-bit-field> to from start end)
                        (if ('#<procedure #3 ##not>
                             (if ('#<procedure #2 ##fixnum?> to) #t (if ('#<procedure #6 ##bignum?> to) #t #f)))
                            (##fail-check-exact-integer 1 '#<procedure #7 copy-bit-field> to from start end)
                            ('#<procedure #8 ##copy-bit-field> ('#<procedure #9 ##fx-> end start) start from to)))
                    (##raise-range-exception 4 '#<procedure #7 copy-bit-field> to from start end))
                (if ('#<procedure #6 ##bignum?> end)
                    (##raise-range-exception 4 '#<procedure #7 copy-bit-field> to from start end)
                    (##fail-check-exact-integer 4 '#<procedure #7 copy-bit-field> to from start end)))
            (##raise-range-exception 3 '#<procedure #7 copy-bit-field> to from start end))
        (if ('#<procedure #6 ##bignum?> start)
            (##raise-range-exception 3 '#<procedure #7 copy-bit-field> to from start end)
            (##fail-check-exact-integer 3 '#<procedure #7 copy-bit-field> to from start end)))))
```
which looks like it may be correct.  The same with `copy-bit`.

I hope this is helpful.   If another way to communicate (e.g., on gitter) would be helpful, let me know.

Brad

gambiteer:
I apologize if my comments seem inconsistent.

antoine-doucet:
Thanks for the comments, very helpful !
Indeed the `prim-nary`  already takes care properly of multiple arguments, I then followed your ideas and rewrote the built-in primitive with precise error-message: some small change to what was done in `lib/_num`, according to the srfi's  argument-order / naming specifications.

I'll get started on the new primitives, following again what was done in the said file and your recommandation. I could write to you on Gitter if some specific question arise!

Thanks,
Antoine

gambiteer:
I've been updating the `*-bit-field` routines in `_num.scm`.

antoine-doucet:
I'll check out those update! 
For my part, I pushed a new commit that encapsulate my last changes, I think some update might be needed once I add some bignum/fixnum-overflow tests!

I believe it could be interesting to add the new primitives in the _num files, once completed.

Antoine

gambiteer:
A few more small comments.

I don't think `log2-binary-factors` will have precise error messages.

I'm not sure how to "shadow" a Gambit built-in function with a SRFI 60 function with the same name but different semantics.  (I think there's at least one case, but I can't find it now.)

Generally speaking `start` and `end` don't have to be fixnums.  You want to check that `start` is a nonnegative exact integer (new macro and `fail-check-nonnegative-exact-integer` for that in `_num.scm` and `_num#.scm`) and that `end` is an exact integer that is >= `start`.   E.g.,
```
> (extract-bit-field  10 (expt 2 100) -1)
1023
```
See the computation of `result-bit-length` in `##bignum.extract--bit-field`.

My opinion now is that Marc's convention of `size` and `position` (both of which need to be nonnegative exact integers) is better than SRFI 60's `start and `end` parameters, they're easier to check. I guess that's neither here nor there at this point.

All those algorithms that are "do something with one bit and then shift the argument/result one bit left/right" are quadratic in the size of the argument/result.  One can devise algorithms that are linear in the size of the result as follows, although I'm not sure that it's important given my doubts about how important `reverse-bit-field` and the conversions to and from lists are.

`reverse-bit-field` might be the most interesting.  The algorithms to compute the reversed field for fixnums are fine, they can't get too big.  But for bignums you could use `fdigits`, which are 8 bits in the C backend and 7 bits in the universal backend.  See `##bignum.fdigit-ref` and `##bignum.fdigit-set!`.  You might need to shift things so that the top bit is aligned on an fdigit boundary.  A compile-time macro can generate a u8vector table of size `##bignum.fdigit-base` such that `(u8vector-ref table i)` contains the reversed bits of `i`.  Then you can put together the (possibly unnormalized) bignum result fdigit by fdigit, mask off the top bits as needed, and call `##bignum.normalize` to finish the reversed bit field.   A similar cleanup is in `##bignum.extract-bit-field`.

gambiteer:
Antoine:

Are you thinking of making any more improvements to this pull request?  Or should I think about merging this pull request the way it is and improving it later?

Brad

antoine-doucet:
Hi Brad, 
I was surely planing on improving this pull request, I had some due date that are now behind!

The new macros are a good idea since the `start` `end` convention are really recurrent in the SRFI world ( I totally agree that the start/length convention make more sens, however).

I like your idea for reducing the space complexity of the algorithm, I think it is worth implementing even for the "less interesting" algo: I can see how some bad usage of multiple conversion could become problematic pretty quickly. 

I'll start by correcting/updating the type/bound checking and then go with `reverse-bit-field` at first.

Thanks,
Antoine

gambiteer:
I just added `macro-check-exact-integer` and `macro-check-nonnegative-exact-integer` and used them here https://github.com/gambit/gambit/commit/ec0a8c97dd110ffb3917c5d646efbdc88f6e2bec .  They may prove useful.

There's also a fixnum bit reversal routine in `_num.scm`:
```
          (define (bit-reverse x n)
            (declare (not interrupts-enabled))
            (do ((i 0 (##fx+ i 1))
                 (x x (##fxarithmetic-shift-right x 1))
                 (result 0 (##fx+ (##fx* result 2)
                                       (##fxand x 1))))
                ((##fx= i n) result)))
```
that I'll pull out later.

lib/srfi/60/60.scm
```diff
@@ -405,9 +405,9 @@
 ;;---------------------------------------------------------------------------
 ;; reverse-bit-field
 
-(define-prim (##reverse-bit-field n start end)
+(define-prim (reverse-bit-field n start end)
 
-  (define (bit-reverse k n)
+  (define (fixnum-bit-reverse k n)
     (let reverse-loop ((m (if (##fxnegative? n)
```
gambiteer, line 411: I've looked at Aubrey's code, and I don't understand why he pre-treats the argument n and then post-processes the result when n is negative.  Wouldn't he get the same result without this pre- and post-treatment?

lib/srfi/60/60.scm
```diff
@@ -423,15 +423,22 @@
...
+    (let reverse-loop ((m (if (##bignum.negative? n)
+                              (##bitwise-not n)
+                              n))
+                       (k (##fx- k 1))
+                       (rvs-acc 0))
+      (if (##fxnegative? k)
+          (if (##bignum.negative? n)
+              (##bitwise-not rvs-acc)
+              rvs-acc)
+          (reverse-loop (##arithmetic-shift m -1)
```
gambiteer, line 436: This line can turn a bignum into a fixnum.

antoine-doucet:
I was considering adding a `macro-check-exact-integer-range-incl` since it is quite common!
 Afterward, every name redefinition could be written cleanly with a `define-primitive` (which I believe is the `define-procedure` for `prim` definition)

I'll then be able to start the algorithmic improvements!

Antoine

PR #664, opened 2021-02-10
Dockerfile for gambit

Velythyl:
# Installing gambit 

Installing gambit can be a bit of a pain sometimes, especially on windows or uncommon linux distributions. Having to compile it from scratch even for quick demonstrations can also be annoying. Also, sometimes it is useful to have multiple gambit versions on hand, in order to test regressions, etc. 

# Installing gambit using Docker

I made a Dockerfile to greatly simplify installing Gambit. Using docker, one can install gambit by doing 

```bash
docker pull velythyl/gambit:latest
```

This pulls the latest gambit version from my own dockerhub profile https://hub.docker.com/repository/docker/velythyl/gambit 

# Running gambit using docker

Then, gambit can by used by doing

```bash
docker run -it velythyl/gambit
```

# What it actually does

This command above will open an ubuntu container running locally, in which gsi and gsc are installed. This also means that one could alias `gsi` in their own shell to `docker run -i velythyl/gambit gsi`, and it would boot you directly into the gambit repl, as if it was installed on your own machine.

# Changing the docker image name (replacing "velythyl" by "gambit")

Of couse, the commands above use the image found on my own dockerhub. You might want to push images to a Gambit profile on dockerhub. To do so, follow the instructions bellow:

# Building the docker image
 
To generate the docker image once a new gambit release is ready, simply do

```bash
docker built -t gambit -t gambit/gambit:latest -t gambit/gambit:<gambit release version> .
docker push gambit/gambit
```

# Note:

I verified with @SamuelYvon that everything works fine :)

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,0 +1,6 @@
+FROM gcc:9.3.0
+WORKDIR gambit_install
+COPY . .
+RUN make clean && ./configure --enable-single-host && make -j$(nproc) && make check && make modules && make doc && make install
+RUN ln -s /gambit_install/gsi/gsi /bin/gsi && ln -s /gambit_install/gsc/gsc /bin/gsc
+WORKDIR /workdir


Velythyl:
Note: the point of this Dockerfile idea was to make a whole build order for Mimosa

- Build gambit using the old version of GCC inside docker
- Build Mimosa `FROM` that version of gambit built in the line above
- Extract Mimosa from the docker
- Voilà, we just built Mimosa locally, without having to deal with a VM or anything

Dockerfile
```diff
@@ -0,0 +1,6 @@
+FROM gcc:9.3.0
+WORKDIR gambit_install
+COPY . .
```
laurenthuberdeau, line 3: `COPY . . ` is noop?

Velythyl: Nope! It copies the local gambit files into the dockerfile

laurenthuberdeau: Ah, yes, the source and destination aren't on the same machine.

Velythyl: Otherwise, the dockerfile only contains a "virgin" ubuntu install.

You could replace `COPY . .` by `RUN git clone https://github.com/gambit/gambit.git`, but I think that you can save the network I/O since you already have all the gambit files locally (aka if you have this Dockerfile, I assume you cloned gambit and are running the commands inside the gambit folder)

Velythyl: > Ah, yes, the source and destination aren't on the same machine.

exactly :)

lassik:
We have docker containers for lots of Scheme implementations at

* https://github.com/scheme-containers
* https://hub.docker.com/u/schemers

Including Gambit:
* https://github.com/scheme-containers/gambit


* `docker run -it schemers/gambit`
* `docker run -it schemers/gambit:head`

SamuelYvon:
@lassik, @Velythyl if this is the case, this should be documented on the readme or in Gambit's doc. I suggest adding instructions here.

Velythyl:
Additonally, I feel like Gambit should have its own Dockerfile in its own repo. My research advisor manages the Duckietown Foundation, and all of our repos use that model (having all the code for the repo + the Dockerfile for containerizing it in the same repo), and we use a LOT of dockers. Separating the Dockerfiles from their code makes everything too hard to maintain, in my opinion.

feeley:
Thanks for your contribution!

PR #668, opened 2021-03-02
Fix too few arguments passed to open_long_path

FredericHamel:
I fixed the call to `open_long_path` that had too few arguments. The mode value passed is 0. It's ignored if the "oflag" doesn't contain `O_CREAT`.


feeley:
Thanks for noticing that!

Issue #659, opened 2021-01-19
letrec* incorrectly reorders statements


fare:
R7RS says that `letrec*` should not reorder the side-effects, that will happen from left-to-right, but this function, when compiled (as opposed to interpreted, which works correctly), reorders line 300 below line 500, which is incorrect. Presumably, it detects that 400 and 500 do not "depend" on x (or anything else defined in the `letrec*`), whereas 300 does.

```
":" ; ( echo -e 'INTERPRETED' ; gsi letrec-star-reordered.scm; echo -e '\nCOMPILED' ; gsc letrec-star-reordered.scm && gsi -e '(load "letrec-star-reordered")' ) 2>&1 | less

(define (displayln . a) (for-each display a) (newline))

(define (reordered)
  (letrec* ((x (begin (displayln 100) 42))
            (_200 (displayln 200))
            (_300 (displayln 300 (list x)))
            (_400 (displayln 400))
            (y (begin (displayln 500) 23)))
    (displayln 600 (list y))))

(reordered)
```

fare:
Marc said on gitter: Concerning the implementation of `letrec*`, there are two implementations… a correct one in the interpreter (procedure `##comp-let-like-form` in `lib/_eval.scm`) and a buggy one in the compiler (procedure `pt-recursive-let` in `gsc/_ptree1.scm`).

gambiteer:
Fixed by 47f2eb7feb5deae94c0b96f5a50310d978e120ae

gambiteer:
Sometimes we make mistakes.  Reopening.

fare:
NB: the fix was immediately reverted in the next commit d1a6c6bf5ea019731c4264ed7a91c0386659a989 because it broke some regression tests... that I believe might need be updated.

SamuelYvon:
@fare 
```shell
$ ./gsc/gsc
Gambit v4.9.4-211-g97e4bff5

> (define (displayln . a) (for-each display a) (newline))
> (define (reordered)
   (letrec* ((x (begin (displayln 100) 42))
            (_200 (displayln 200))
            (_300 (displayln 300 (list x)))
            (_400 (displayln 400))
            (y (begin (displayln 500) 23)))
    (displayln 600 (list y))))
> (reordered)

100
200
300(42)
400
500
600(23)
```

fare:
Yes, this is the desired behaviour. As of 4.9.4-210-g420f1e5e, it works in the interpreted case, and fails in the compiled case with
```
100
200
[...]

PR #670, opened 2021-03-05
Fix for issue #659

SamuelYvon:
Removed reordering when we are creating the bindings for `let*`.  One can verify with the provided example of #659 that the `_300` variable is not reordered.


feeley:
Thank you for your contribution.  The checks (`make check`) don't succeed and this is a requirement for accepting the PR.  Could you please fix this?

SamuelYvon:
I'll investigate later; the only test that fails is the first one (`test1`). Removing it and running `make check` shows `All tests succeeded`

gambiteer:
I wondered whether the code in the interpreter and this new code is the same.

SamuelYvon:
It seems like this change breaks inlining... Back to drawing board.

fare:
The error is as follows:
```
make[1]: Entering directory '/d/a/gambit/gambit/tests'
------------ TEST 1 (debugging support)
../gsi/gsi.exe -:~~bin=../bin,~~lib=../lib,~~include=../include -f debug.scm > test1.out
diff test1.ok test1.out && rm -f test1.out
262,264c262,265
< #return(##cprc-prc 7) : ($code args i inner-rte j n proc rte)
< #return(##cprc-prc 8) : ($code end i inner-rte j k keys left rest? rte)
< #return(##cprc-prc 9) : ($code args entry-hook exec proc rte)
make[1]: *** [makefile:119: test1] Error 1
---
make: *** [makefile:208: check] Error 2
> #return(##cprc-prc 7) : ($code args inner-rte n proc reject-illegal-dsssl-parameter-list? rte)
> #return(##cprc-prc 8) : ($code args i inner-rte j n proc reject-illegal-dsssl-parameter-list? rte)
> #return(##cprc-prc 9) : ($code end i inner-rte j k keys left rest? rte)
> #return(##cprc-prc 10) : ($code args entry-hook exec proc rte)
make[1]: Leaving directory '/d/a/gambit/gambit/tests'
Error: Process completed with exit code 2.
```
It looks like some execution trace has changed. The change might quite possibly be legal and expected, after the compiler was changed to handle letrec* differently. How do I tell? And how do I get to run all the tests without stopping at the first error?

gambiteer:
make -i check

gambiteer:
I tried this version of the patch:
```
diff --git a/gsc/_ptree1.scm b/gsc/_ptree1.scm
index 71d68bf3..3d992653 100644
--- a/gsc/_ptree1.scm
+++ b/gsc/_ptree1.scm
@@ -1896,8 +1896,9 @@
             (let* ((var (car l))
                    (val (val-of var)))
               (if (or (prc? val)
-                      (not (varset-intersects? (bound-free-variables val)
-                                               vars-set)))
+                      (not (or *?
+                               (varset-intersects? (bound-free-variables val)
+                                                   vars-set))))
                 (loop1 (cdr l)
                        (cons var vars-b)
                        (cons val vals-b)
```
and it passed `make check` no problems, and it fixed issue #659.

I don't understand this code very well at all, but I'm included to push this.  Any thoughts?

gambiteer:
I pushed your fix as is.

fare:
It seems to me that the failing regression test caused by the fix is that some bytecode changed... which it probably should if the code uses letrec* which is now fixed. So maybe the fix actually works and the test must be changed? Who understands the failing test? Can we get this fixed? It really affects gerbil coding style badly.

SamuelYvon:
The code was pushed by @gambiteer directly, so your issue must come from somewhere else. See commit https://github.com/gambit/gambit/commit/47f2eb7feb5deae94c0b96f5a50310d978e120ae. However issue #659 lists the issue as open and mentions that its still a problem. Maybe the commit was reverted?

fare:
Yes, it was immediately reverted in d1a6c6bf5ea019731c4264ed7a91c0386659a989 for breaking regression tests... that I believe might just need to be updated.

SamuelYvon:
I see. The code has significantly changed from a quick glance, let me take a second go at it.

fare:
Force-push? Ouch, commit wars are bad... also, I don't see a fix in master currently. Why close? Or is the closing a subtle side-effect of the force-push?

PR #678, opened 2021-04-05
Remove deprecated macros in configure.ac

FredericHamel:
I didn't generate the configure script because in this PR because I don't have a usable macOS machine of my own.

This should be tested on macOS before merging this PR.

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -27,7 +27,10 @@ else
   echo "/usr/local/$PACKAGE_NAME" ;
 fi`)
 
-AC_CANONICAL_SYSTEM
+# This replace AC_CANONICAL_SYSTEM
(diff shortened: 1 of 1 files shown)


FredericHamel:
New update this doesn't break the `--enable-cplusplus`. It might have been the g++ version. It works with the following version:
```                                                                  
g++ (GCC) 11.0.0 20200813 (experimental)
Copyright (C) 2020 Free Software Foundation, Inc.
This is free software; see the source for copying conditions.  There is NO
warranty; not even for MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.
```

feeley:
Thanks!

PR #683, opened 2021-04-22
This adds --enable-c-opt-rts to control the RTS opt level.

iains:
The option behaves in a similar manner to --enable-c-opt.

--enable-c-opt-rts=yes will use the highest opt the compiler
supports (typically O3 for GCC, for example).

--enable-c-opt-rts=no will use "-O1".

--enable-c-opt-rts="-x -y -z" will use "-x -y -z" as the
optimisation settings for the RTS.

Omitting the option will behave the same as --enable-c-opt-rts=yes

### DIFF
--- a/configure
+++ b/configure
@@ -954,6 +954,7 @@ enable_option_checking
 enable_cplusplus
 enable_guide
 enable_c_opt
+enable_c_opt_rts
 enable_pic
 enable_debug
 enable_debug_log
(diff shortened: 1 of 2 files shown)


iains:
I guess something in the regenerated include/config.h.in  is conflicting...
... not sure what you usually use to regenerate.

I used autoreconf -fv  (with autoconf 2.69)

feeley:
Thanks for the PR.  The changes look good except for `include/config.h.in` which should not be part of the commit since that file is managed manually.  Maybe it should be managed by autoreconf but that's best left for a future change because I fear it might break a few things at this point.  Could you fix and resubmit your PR?

iains:
ah, OK wil do.

PR #684, opened 2021-04-22
This adds --enable-c-opt-rts to control the RTS opt level.

iains:
The option behaves in a similar manner to --enable-c-opt.

--enable-c-opt-rts=yes will use the highest opt the compiler
supports (typically O3 for GCC, for example).

--enable-c-opt-rts=no will use "-O1".

--enable-c-opt-rts="-x -y -z" will use "-x -y -z" as the
optimisation settings for the RTS.

Omitting the option will behave the same as --enable-c-opt-rts=yes

### DIFF
--- a/configure
+++ b/configure
@@ -954,6 +954,7 @@ enable_option_checking
 enable_cplusplus
 enable_guide
 enable_c_opt
+enable_c_opt_rts
 enable_pic
 enable_debug
 enable_debug_log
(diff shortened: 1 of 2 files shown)


iains:
hmm .. not sure what's happening here, this was restricted to the configure changes only - and those should not have altered the default (without the option).  Sorry don't have cycles to poke at this more just now.

feeley:
I'm not sure why you retracted your PR.  The CI errors are "expected" because of a problem with the Windows CI that hasn't been resolved.  So I think your PR was good.

iains:
Ah, OK I didn't know that :)

feeley:
Thank you for your contribution and also for helping understand why gcc 10.3 miscompiles Gambit on macOS when the runtime system is compiled as a shared library (https://gcc.gnu.org/bugzilla/show_bug.cgi?id=100152).  As you mention the issue can be avoided by using the new `--enable-c-opt-rts=no` configure option.

PR #689, opened 2021-05-13
Add some Floyd procedures from SRFI 1

lassik:
To start off, here are three list classification procedures and the cycle-aware `length+` procedure from SRFI 1. Does this look reasonable? Coding style can probably be improved.

Once these procedures are OK, we can merge them, or I can add all the SRFI 1 stuff by pushing more commits to the same PR.

### DIFF
--- a/lib/gambit#.scm
+++ b/lib/gambit#.scm
@@ -131,6 +131,7 @@ cfun-conversion-exception-procedure
 cfun-conversion-exception?
 char-foldcase
 circular-list
+circular-list?
 clear-bit-field
 close-port
 command-line
@@ -194,6 +195,7 @@ display-procedure-environment
 divide-by-zero-exception-arguments
 divide-by-zero-exception-procedure
 divide-by-zero-exception?
+dotted-list?
 drop
 emergency-exit
 eof-object
@@ -460,6 +462,7 @@ keyword-hash
 keyword?
 last
 last-pair
+length+
 length-mismatch-exception-arg-id
 length-mismatch-exception-arguments
 length-mismatch-exception-procedure
@@ -616,6 +619,7 @@ process-times
 processor-id
 processor?
 promise?
+proper-list?
 protocol-info
 protocol-info-aliases
 protocol-info-name
--- a/lib/gambit/list/list.scm
+++ b/lib/gambit/list/list.scm
@@ -101,12 +101,14 @@
   (macro-force-vars (obj)
     (##null? obj)))
 
-;; Floyd's tortoise and hare algorithm for cycle detection.
+;; Floyd's tortoise and hare algorithm for cycle detection
+
+;; https://en.wikipedia.org/wiki/Cycle_detection
 
 ;; These procedures may get into an infinite loop if another thread
 ;; mutates "lst" (if lst1 and lst2 each point to disconnected cycles).
 
-(define-procedure (floyd-tail lst)
+(##define (##possibly-cyclic-tail lst)
   (include "~~lib/gambit/prim/prim#.scm") ;; map fx+ to ##fx+, etc
   (let loop ((fast lst) (slow lst))
     (macro-force-vars (fast)
@@ -126,12 +128,12 @@
                     (else
                      fast))))))))
 
-(define-procedure (floyd-length lst)
+(##define (##possibly-cyclic-length lst)
   (include "~~lib/gambit/prim/prim#.scm") ;; map fx+ to ##fx+, etc
   (let loop ((len 0) (fast lst) (slow lst))
     (macro-force-vars (fast)
       (cond ((null? fast)
-             (* 2 len))
+             (fx* 2 len))
             ((not (pair? fast))
              #f)
             (else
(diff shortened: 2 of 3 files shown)


feeley:
Thanks for the PR.  Here are my comments.

1) Avoid using `floyd` in the procedure names as this relates to the implementation.  Names like `possibly-cyclic-tail` or `general-tail`, etc would be better since the implementation could change in the future.  Adding a reference to https://en.wikipedia.org/wiki/Cycle_detection will help document how the algorithm works.

2) `floyd-tail` and `floyd-length` should not be defined with `define-procedure` as this is for exported procedures that do type checking.  I guess it doesn't hurt to use `define-procedure` but it leaves the impression that this is an exported procedure.

3) `floyd-length` is using generic arithmetic (`+`, `*`, ...) to compute the length.  It is safe to use fixnum operations here as the length will never exceed a fixnum.

4) The procedure `floyd-tail` is called in non-tail position by `proper-list?`, `circular-list?` and `dotted-list?`.  This is OK in this particular situation because `floyd-tail` does not raise exceptions.  Had this been the case it would have been impossible (or hard) to ensure that the exception is raised in `floyd-tail` with the continuation passed to `proper-list?`, `circular-list?` or `dotted-list?`.  I'm just explaining this so that care is taken when defining more advanced procedures of SRFI-1.

lassik:
Thanks for the detailed review, very well explained. The above commit fixes most of those things.

I changed the helper procedures into `##possibly-cyclic-tail` and `##possibly-cyclic-length`.

Trying this:

```
(##namespace ("##"))               ;; in ## namespace by default
(##include "~~lib/_gambit#.scm")   ;; for define-procedure, etc
(##include "list#.scm")            ;; use correct namespace for exported procedures
```

in `list.scm` caused the build to fail. There's quite a bit of existing code in the list library so it's no wonder; we should do a full code review of the library once we have all the SRFI 1 procedures.

I didn't yet add the tests since I'm not sure what to do about the SRFI 1 tests. Since most of the stuff in `(gambit list)` is from SRFI 1, it would be ideal to re-use the SRFI's tests as-is. Then they could be easily shared with other Scheme implementations.

The https://github.com/srfi-explorations/srfi-test repo has tests aggregated from multiple Scheme implementations as well as SRFIs' reference implementations. Each SRFI's tests are bundled into an `.scm` file with the intent that those `.scm` files would be easy to import into Scheme implementations' own repos with no changes (each implementation would add its own `.sld` or similar wrapper to run the tests in its native way).

feeley:
Thanks for the corrections.  Concerning the tests, the tests from srfi-test should be seen as complementary test because Gambit has additional constraints on the behaviour of the procedures.  For example procedures need to raise exceptions in tail position (something that isn't required by SRFI 1 and is not something that can be tested in a portable way).  So I will add tests in a separate commit.

I agree that the `##` namespace issue should be addressed in a later code review.  In addition, the repeated `(include "~~lib/gambit/prim/prim#.scm")` should be removed from each procedure's body and put at the top of the file.

lassik:
Great.

I hope we can convince the other Scheme implementers to add tail-position checks to the cross-implementation testing API. In implementations that don't care about the issue, `test-tail-error` can be an alias for the ordinary `test-error`.

> In addition, the repeated `(include "~~lib/gambit/prim/prim#.scm")` should be removed from each procedure's body and put at the top of the file.

Right, I was wondering why it's repeated :-)

feeley:
Yes the `(include "~~lib/gambit/prim/prim#.scm")` is repeated because it will cause conflicts with the other procedures in `list.scm` that haven't yet been updated to use the new coding style.

I agree that it would be nice if `test-tail-error` existed, even if it is partially implemented by most Scheme implementations.  Another possible difference is that Gambit sometimes adds optional parameters so the test for wrong number of arguments will also be non-portable.

Commit a445f68da03e9ce50a386b2fed98f4bc071ef3b4 adds the unit tests for the new procedures.  There are now officially more than `(expt 2 8)` unit tests!  Unfortunately, that covers only 1/4 of all the procedures defined by Gambit...

lassik:
> Yes the `(include "~~lib/gambit/prim/prim#.scm")` is repeated because it will cause conflicts with the other procedures in `list.scm` that haven't yet been updated to use the new coding style.

> * `floyd-tail` and `floyd-length` should not be defined with `define-procedure` as this is for exported procedures that do type checking.  I guess it doesn't hurt to use `define-procedure` but it leaves the impression that this is an exported procedure.

Could we have separate `define-procedure` and `define-export-procedure` versions of the macro?

> I agree that it would be nice if `test-tail-error` existed, even if it is partially implemented by most Scheme implementations. Another possible difference is that Gambit sometimes adds optional parameters so the test for wrong number of arguments will also be non-portable.

I'll ask a few of them about it.

> Commit [a445f68](https://github.com/gambit/gambit/commit/a445f68da03e9ce50a386b2fed98f4bc071ef3b4) adds the unit tests for the new procedures. There are now officially more than `(expt 2 8)` unit tests! Unfortunately, that covers only 1/4 of all the procedures defined by Gambit...

This is probably a problem with all Scheme implementations. Thank you for writing the new ones!

feeley:
I noticed that the "primitives" for the new list procedures were not defined, so I added those in commit f8ce09034b02cc0b324476e19254ccf94a3b4443 .

Primitives are versions of the runtime library procedures that don't check type errors and they are meant for writing the runtime library itself.  They have the same name as the type checking procedures but with a `##` prefix.  When in the scope of a `(include "~~lib/gambit/prim/prim#.scm")` the use of a procedure like `proper-list?` will be replaced with `##proper-list?`, which is convenient for defining runtime library procedures without having to use `##` all over the code.  Primitives also solve the problem that (in an R5RS system) mutations of `proper-list?` will not affect the code that calls `##proper-list?` (or `proper-list?` in the scope of a `(include "~~lib/gambit/prim/prim#.scm")`).

> Could we have separate `define-procedure` and `define-export-procedure` versions of the macro?

What would it do?  The actual "exporting" is done elsewhere (using `namespace`).

lassik:
Makes sense.

If `make check` has a better way to figure out what's exported and what's not then a separate `define-export-procedure` is not needed.

PR #691, opened 2021-05-15
Add some build products to .gitignore

lassik:


.gitignore
```diff
@@ -140,6 +140,7 @@ include/gambit.h
 include/gambit-not*.h
 include/makefile
 lib/_eval.o
+lib/_gambit.js
```
feeley, line 143: Might as well add the other universal backend targets:

lib/_gambit.py
lib/_gambit.php
lib/_gambit.rb
lib/_gambit.java
lib/_gambit.go

feeley (changes requested):

.gitignore
lassik: Added in a new commit.

PR #696, opened 2021-05-26
Add filter and remove (SRFI 1)

lassik:
Is the helper macro OK? Could use gensyms, but it's not intended for general use.

### DIFF
--- a/lib/gambit/list/list.scm
+++ b/lib/gambit/list/list.scm
@@ -607,15 +607,16 @@
                 (macro-check-list x 2 (assoc obj lst c)
                   #f))))))))
 
-(define-macro (##macro-filter prim-call list var test-expr)
+(define-macro (macro-filter prim-call list var test-expr)
   `(let ()
      (define (filter-tail list)
        (if (pair? list)
            (let* ((,var (car list))
+                  (matches? ,test-expr)
                   (old-tail (cdr list))
                   (new-tail (filter-tail old-tail)))
              (if (and (or (pair? new-tail) (null? new-tail))
-                      ,test-expr)
+                      matches?)
                  (if (eq? old-tail new-tail)
                      list
                      (cons x new-tail))
@@ -628,13 +629,13 @@
                         new-list))))
 
 (define-prim&proc (filter (pred procedure) (list object))
-  (##macro-filter (filter pred list) list x (pred x)))
+  (macro-filter (filter pred list) list x (pred x)))
 
 (define-prim&proc (remove (pred procedure) (list object))
-  (##macro-filter (remove pred list) list x (not (pred x))))
+  (macro-filter (remove pred list) list x (not (pred x))))
 
 (define-prim&proc (remq (elem object) (list object))
-  (##macro-filter (remq elem list) list x (not (eq? x elem))))
+  (macro-filter (remq elem list) list x (not (eq? x elem))))
 
 ;;;----------------------------------------------------------------------------
 
--- a/lib/gambit/list/list.scm
+++ b/lib/gambit/list/list.scm
@@ -607,22 +607,34 @@
                 (macro-check-list x 2 (assoc obj lst c)
                   #f))))))))
 
-(define-prim (##remq elem lst)
+(define-macro (##macro-filter prim-call list var test-expr)
+  `(let ()
+     (define (filter-tail list)
+       (if (pair? list)
+           (let* ((,var (car list))
+                  (old-tail (cdr list))
+                  (new-tail (filter-tail old-tail)))
+             (if (and (or (pair? new-tail) (null? new-tail))
+                      ,test-expr)
+                 (if (eq? old-tail new-tail)
+                     list
+                     (cons x new-tail))
+                 new-tail))
+           list))
+     (let ((new-list (filter-tail ,list)))
+       (macro-if-checks (if (or (pair? new-list) (null? new-list))
+                            new-list
+                            (macro-fail-check-list 2 ,prim-call))
+                        new-list))))
 
-  (include "~~lib/gambit/prim/prim#.scm") ;; map fx+ to ##fx+, etc
+(define-prim&proc (filter (pred procedure) (list object))
+  (##macro-filter (filter pred list) list x (pred x)))
 
-  (define (rem elem lst)
-    (if (pair? lst)
-        (let* ((lst2 (rem elem (cdr lst)))
-               (x (car lst)))
-          (if (eq? x elem)
-              lst2
-              (if (eq? lst2 (cdr lst))
-                  lst
-                  (cons x lst2))))
-        '()))
+(define-prim&proc (remove (pred procedure) (list object))
+  (##macro-filter (remove pred list) list x (not (pred x))))
 
-  (rem elem lst))
+(define-prim&proc (remq (elem object) (list object))
+  (##macro-filter (remq elem list) list x (not (eq? x elem))))
 
 ;;;----------------------------------------------------------------------------
 
--- a/lib/gambit#.scm
+++ b/lib/gambit#.scm
@@ -279,6 +279,7 @@ file-number-of-links
 file-owner
 file-size
 file-type
+filter
 finite?
 first-bit-set
 fixnum->flonum
@@ -676,6 +677,8 @@ readtable-write-extended-read-macros?
 readtable-write-extended-read-macros?-set
 readtable?
 real-time
+remove
+remq
 rename-file
 repl-display-environment?
 repl-error-port
--- a/lib/gambit/list/list#.scm
+++ b/lib/gambit/list/list#.scm
@@ -47,8 +47,10 @@ cdddr
 cddr
 cdr
 circular-list
+circular-list?
 cons
 cons*
+dotted-list?
 drop
 fold
 fold-right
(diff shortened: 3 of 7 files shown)


lassik:
The `(lst object)` args should probably specify the type as pair-or-null instead of `object`.

lib/gambit/list/list.scm
```diff
@@ -607,22 +607,26 @@
...
-              lst2
-              (if (eq? lst2 (cdr lst))
-                  lst
-                  (cons x lst2))))
-        '()))
-
-  (rem elem lst))
+(define-macro (##filter-macro lst var test-expr)
+  `(let recurse ((lst ,lst))
+     (if (pair? lst)
```
feeley, line 612: I tend to not use named `let` for recursions and prefer to do `(let () (define (recurse ...) ...) (recurse ...))`.  I also avoid the name `recurse` that sounds like repeated cursing...

lib/gambit/list/list.scm
```diff
@@ -607,22 +607,26 @@
...
+         (let* ((lst2 (recurse (cdr lst)))
+                (,var (car lst)))
+           (if ,test-expr
+               (if (eq? lst2 (cdr lst))
```
feeley, line 616: This is not quite right in a multithreaded Scheme because the `cdr` of `lst` could have been modified by a second thread to make the test `(eq? lst2 (cdr lst))` succeed (it is very unlikely, but still possible).  It is better to do `(cdr lst)` once and save the value in a `rest` variable in the `let*`, and then test `(eq? lst2 rest)`.

feeley (changes requested):

lib/gambit/list/list.scm
```diff
@@ -607,22 +607,26 @@
...
+                   lst
+                   (cons x lst2))
+               lst2))
+         '())))
+
+(define-prim&proc (filter (pred procedure) (lst object))
```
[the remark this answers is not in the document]
lassik: Yes, I knew about `,h` - neat feature.

lassik:
OK, the new commit should fix all of the issues in the review.

lib/gambit/list/list.scm
```diff
@@ -607,22 +607,34 @@
...
-               (x (car lst)))
-          (if (eq? x elem)
-              lst2
-              (if (eq? lst2 (cdr lst))
-                  lst
-                  (cons x lst2))))
-        '()))
-
-  (rem elem lst))
+(define-macro (##macro-filter prim-call list var test-expr)
```
feeley, line 610: The name of the macro should be `macro-filter` without the `##`.

lib/gambit/list/list.scm
```diff
@@ -607,22 +607,34 @@
...
+                     (cons x new-tail))
+                 new-tail))
+           list))
+     (let ((new-list (filter-tail ,list)))
+       (macro-if-checks (if (or (pair? new-list) (null? new-list))
+                            new-list
+                            (macro-fail-check-list 2 ,prim-call))
+                        new-list))))
+
+(define-prim&proc (filter (pred procedure) (list object))
```
feeley, line 630: I think it is wrong to indicate that `list` is just an `object`.  It would be better to indicate `proper-list` and to have a `macro-check-proper-list` macro that doesn't do any type checking since it will be done later.  At least that way the procedure header contains a type indication that could be used by other tools (like a document generator).  For now the `macro-check-proper-list` macro could be defined locally in `list.scm` like this:

```
(define-syntax macro-check-proper-list
  (lambda (src)
    (##deconstruct-call
     src
     5
     (lambda (arg-src pos-src call-src body-src)
       body-src))))
```

Identical definitions should exist for `dotted-list` and `circular-list`.  To avoid confusion the current macros `macro-check-list` and `macro-check-pair-list` (which do a type check for `null?` and `pair?` respectively but with an exception message saying a "LIST" was expected) should be renamed to `macro-check-empty-proper-list` and `macro-check-nonempty-proper-list`. That can wait for another PR.

feeley (changes requested):

lib/gambit/list/list.scm
lassik: OK, changed.

lib/gambit/list/list.scm
lassik: I tried to add this macro and give `proper-list` as the type in `define-prim&proc` but got segfaults and didn't manage to figure out why it's not working. Should we also add `proper-list` to `##type-exception-names` in `_repl.scm`?

lassik: @feeley Adding `proper-list` seems more difficult than the rest of this PR. Can you merge this and add the `proper-list`/`dotted-list`/`circular-list` checks? If you can add each to one procedure, I can fill in the rest by looking at how those examples do it.

feeley:
I will adjust according to my comments.

feeley:
I'm looking into this at the moment.  One issue is what to call the various types of lists in error messages to highlight the fact that a circular list or dotted list are not allowed.  There are also "pair lists" that are used by `assoc` (as association lists) and "string lists" for `concatenate`, etc.   Here's how error messages look like currently and a possible new way:

```
% gsi -:d1 -e "(member 1 '(2 3 . 4))"
*** ERROR IN (string)@1.1 -- (Argument 2) LIST expected
(member 1 '(2 3 . 4))
% gsi -:d1 -e "(assoc 1 '(2 3))"
*** ERROR IN (string)@1.1 -- (Argument 2) PAIR LIST expected
(assoc 1 '(2 3))


% gsi/gsi -:d1 -e "(member 1 '(2 3 . 4))"
*** ERROR IN (string)@1.1 -- (Argument 2, list) PROPER LIST expected
(member 1 '(2 3 . 4))
% gsi/gsi -:d1 -e "(assoc 1 '(2 3))"
*** ERROR IN (string)@1.1 -- (Argument 2, alist) PAIR PROPER LIST expected
(assoc 1 '(2 3))
% gsi/gsi -:d1 -e "(length+ '(2 . 3))"
*** ERROR IN (string)@1.1 -- (Argument 1, clist) PROPER or CIRCULAR LIST expected
(length+ '(2 . 3))
```

So here are possible names for the 3 kinds of plain lists:

```
short and sweet               verbose and precise
LIST                          PROPER LIST
LIST or CIRCULAR LIST         PROPER LIST or CIRCULAR LIST
LIST or DOTTED LIST           PROPER LIST or DOTTED LIST
```

The problem with being precise is that some novices will not understand why "PROPER" is there.

Any suggestions?

lassik:
Thanks for getting back to this! Would it be friendlier to novices if `(member 1 "234")` says `LIST expected`, but `(member 1 '(2 3 . 4))` says `PROPER LIST expected`? So the `LIST` case would be triggered if a non-list is given as the argument, and the `PROPER LIST` case happens if things go wrong during traversal.

feeley:
That's an interesting idea.  However it means that there are two possible exceptions from those procedures.  On the other hand they are both instances for which `type-exception?` is true, so it shouldn't be more complex to catch the exceptions.  I'll prototype it.

lassik:
Thanks. I agree that conforming to `type-excpetion?` is more important than the precise exception being raised.

feeley:
Please check out the latest commit of list.scm (https://github.com/gambit/gambit/blob/master/lib/gambit/list/list.scm).

All of the list procedures have been refactored to use define-procedure, define-primitive or define-prim&proc.  The ## prefix is now gone from the file, but for a couple of minor exceptions.  It certainly looks more like "normal" Scheme code now and probably more portable to other implementations of Scheme.

The precise handling of exceptions ("LIST" vs "PROPER LIST" distinction) was not easy to handle concisely, so some of the definitions are more complex than would appear to be strictly necessary at first glance.  Here are some examples:

```
> (assoc 5 'not-a-pair-or-null)                      
*** ERROR IN (stdin)@1.1 -- (Argument 2, alist) LIST expected
(assoc 5 'not-a-pair-or-null)

> (assoc 5 '((1 . 11) (2 . 22) . not-a-pair-or-null))
*** ERROR IN (stdin)@2.1 -- (Argument 2, alist) PROPER LIST expected
(assoc 5 '((1 . 11) (2 . 22) . not-a-pair-or-null))

> (assoc 5 '((1 . 11) not-a-pair (2 . 22)))          
*** ERROR IN (stdin)@3.1 -- (Argument 2, alist) PAIR LIST expected
(assoc 5 '((1 . 11) not-a-pair (2 . 22)))

> (fold vector 0 '(1 2) '(3 4))
#(2 4 #(1 3 0))

> (fold vector 0 'not-a-pair-or-null '(3 4))
*** ERROR IN (stdin)@5.1 -- (Argument 3, clist1) LIST expected
(fold '#<procedure #2 vector> 0 'not-a-pair-or-null '(3 4))

> (fold vector 0 '(1 2) 'not-a-pair-or-null)       
*** ERROR IN (stdin)@6.1 -- (Argument 4) LIST expected
(fold '#<procedure #2 vector> 0 '(1 2) 'not-a-pair-or-null)
```

Note that the name of the parameter is part of the error message when it is available (I looked at the R7RS and SRFI 1 for the parameter names).  For `map`, `for-each`, `fold`, `fold-right` there is no name for the lists after the first in the documentation, so there is none in the error message, just the argument position.

I added a few unit tests for some special cases I think are relevant.

The list.scm file is thus a good source of inspiration to continue the implementation of the SRFI 1 procedures.

PR #699, opened 2021-05-29
Improve try Gambit build compatibility across platforms.

Negdayen:
GNU sed does not support -I and is fussy about spaces after -i


feeley:
Thanks for your contribution!

PR #700, opened 2021-06-06
Implement SRFI 26.

Negdayen:


lassik:
Great. Maybe merge #652 and rebase this PR on top of the new master?

feeley:
Actually, I have accepte PR #652 which I had forgotten to do.  So please refresh your PR accordingly.

lib/srfi/26/expand-cut.scm
```diff
@@ -0,0 +1,46 @@
+(define (expand-cut id close? list<argument-or-slot>)
+
+  (define (gs s) (gensym (string->uninterned-symbol s)))
```
feeley, line 3: Using `gensym` directly, as in `(gensym 'slot)`, is the preferred style.  It isn't clear why the call to `string->uninterned-symbol` is there.

lib/srfi/26/expand-cut.scm
```diff
@@ -0,0 +1,46 @@
...
+
+  (let cons-closure ((l<argument-or-slot> list<argument-or-slot>)
```
feeley, line 5: The use of angle brackets within the identifiers is distracting.  Moreover the use of `l` as an identifier (with or without the angle brackets) should be avoided because it is easy to confuse with a `1`.  `lst` would be preferable.

lib/srfi/26/26#.scm
```diff
@@ -0,0 +1,9 @@
+(##namespace ("srfi/26#" cut cute))
+
+(define-macro (cut . list<argument-or-slot>)
```
feeley, line 3: The use of `define-macro` should be avoided because the location information of the `cut` arguments will be lost.  Using `##define-syntax` avoids this issue.  Please see `lib/_test/_test#.scm` and `lib/_test/test-expand.scm` for an example of how this should be done.

lib/srfi/26/expand-cut.scm
```diff
@@ -0,0 +1,46 @@
...
+                             (cons parameter formals)
+                             (cons parameter procedure-application))))
+
+            ((<...>)
+             (if (null? (cdr l<argument-or-slot>))
+                 (let ((rest-parameter (gs "rest-slot")))
+                   `(##let ,let-over-lambda-bindings
+                      (##lambda (,@(reverse formals) . ,rest-parameter)
+                        (##apply ,@(reverse procedure-application) ,rest-parameter))))
+                 (error "<...> in non-tail position:" (cons id list<argument-or-slot>))))
```
feeley, line 33: Using `error` is not ideal because the error message will not contain location information.  It is better to use a call to `##raise-expression-parsing-exception` (but this will require using `##define-syntax` instead of `define-macro`).  An alternative is to generate the call

    `(syntax-error "<...> in non-tail position:" ,@(cons id list<argument-or-slot>))

feeley (changes requested):

lib/srfi/26/expand-cut.scm
Negdayen: Original implementation comes from my library of Gambit utils, which just uses define-runtime-macro everywhere instead of define-macro and stores all syntax as a part of the runtime.  My original thought is this might save permanently allocating a symbol or two, although I'm not sure if that's accurate (for this version it isn't helpful and will change).

lib/srfi/26/26#.scm
Negdayen: I ended up using the syntax->list form instead of the ##deconstruct-call forms I saw in _test, I think this is solved now.

feeley:
Thanks for your contribution!

PR #703, opened 2021-06-13
Implement SRFI 2: and-let*

Negdayen:


lib/srfi/2/2#.scm
```diff
@@ -0,0 +1,7 @@
+(##namespace ("srfi/2#" and-let*))
+
+(##define-syntax and-let*
+  (lambda (src)
+    (include "syntax-utils.scm")
```
feeley, line 5: Currently and for the foreseeable future, a `##define-syntax` form will cause the `(lambda (src) ...)` to evaluate to an interpreted procedure.  This is because the `lambda` is evaluated at macro expansion time (essentially by calling `eval`).  This means that all the code that is `include`d here will also be interpreted which could be slow if it represents a lot of code or a complex expansion algorithm.  Although this is probably not the case here it is better to adopt a more efficient style based on `import` that allows the expansion process of the macro to be compiled code.

This can be achieved by turning the "expansion algorithm" into a module and `import`ing it rather than `include`ing it.  Most Gambit modules that export syntax are written in this style.  For example the `_match` module's `_match#.scm` file contains:

```
(##define-syntax match
  (lambda (src)
    (##import _match/match-expand)
    (match-expand src
                  #f    ;; use-question-mark-prefix-pattern-variables?
                  #t    ;; use-exhaustive-cases?
                  #f))) ;; use-else?
```

and the `_match/match-expand` module contains the definition of the `match-expand` procedure that implements the expansion algorithm.  The file `lib/_match/makefile` contains the following directives to compile both the `_match` and `match-expand` modules (at the line binding `MAIN_MODULES`):

```
herefromlib = _match
libfromhere = ..
SUBDIRS =
HEADERS_SCM = _match\#.scm match-expand\#.scm
MODULES_SCM = _match.scm match-expand.scm test.scm
MAIN_MODULES = _match match-expand
OTHER_RCFILES = makefile

include $(libfromhere)/module-common.mk
```

The macro definition for `and-let*` should be changed to:

```
(##define-syntax and-let*
  (lambda (src)
    (import srfi/2/expand)
    (and-let*-expand src)))
```

and the definition of the `and-let*-expand` procedure should be put in the `srfi/2/expand` module.  The `lambda` is still creating an interpreted procedure but when `and-let*-expand` is called there will be a transition to fast compiled code.  Another option to avoid the small amount of interpretation when the macro is called is to write the macro definition this way:

```
(##define-syntax and-let*
  (let ()
    (import srfi/2/expand)
    and-let*-expand))
```

The `(let ...)`, which is interpreted a single time regardless of the number of calls to the `and-let*` macro, will return the compiled procedure so that the macro expander calls it directly.  I prefer the first style because it allows slightly more flexibility, like having multiple macros that share the same expansion procedure whose behaviour is selected with parameters (like you have done for SRFI 26).

By the way I notice that SRFI 26 could also be updated to follow the same style.

feeley:
Thanks!

Issue #635, opened 2020-09-07
Build fails with --enable-cplusplus on Windows 10 with MSYS2 MinGW


Negdayen:
Not sure if this is supported; my understanding of the GNU Build System and building C/C++ is pretty weak in general.  Compiling without --enable-cplusplus works fine.  I added -fpermissive but it didn't seem to help much.

```
make stamp
make[1]: Entering directory '/home/jh185336/me/me/gambit-scheme'
make[2]: Entering directory '/home/jh185336/me/me/gambit-scheme/include'
*** include/stamp.h untouched (___STAMP_VERSION did not change)
make[2]: Leaving directory '/home/jh185336/me/me/gambit-scheme/include'
make[1]: Leaving directory '/home/jh185336/me/me/gambit-scheme'
making all in include
make[1]: Entering directory '/home/jh185336/me/me/gambit-scheme/include'
make[1]: Leaving directory '/home/jh185336/me/me/gambit-scheme/include'
making all in bin
make[1]: Entering directory '/home/jh185336/me/me/gambit-scheme/bin'
rm -f gambvcs.bat
if test ".bat" = ""; then \
  cp gambvcs.unix gambvcs.bat; \
  chmod +x gambvcs.bat; \
else \
  sed "s/\$/\r/g" gambvcs.bat.windows > gambvcs.bat; \
fi
rm -f gambuild-C.bat
if test ".bat" = ""; then \
  cp gambuild-C.unix gambuild-C.bat; \
  chmod +x gambuild-C.bat; \
else \
  sed "s/\$/\r/g" gambuild-C.bat.windows > gambuild-C.bat; \
fi
rm -f gambuild-js.bat
if test ".bat" = ""; then \
  cp gambuild-js.unix gambuild-js.bat; \
  chmod +x gambuild-js.bat; \
else \
  sed "s/\$/\r/g" gambuild-js.bat.windows > gambuild-js.bat; \
fi
rm -f gambuild-python.bat
if test ".bat" = ""; then \
  cp gambuild-python.unix gambuild-python.bat; \
  chmod +x gambuild-python.bat; \
else \
  sed "s/\$/\r/g" gambuild-python.bat.windows > gambuild-python.bat; \
fi
rm -f gambuild-ruby.bat
if test ".bat" = ""; then \
  cp gambuild-ruby.unix gambuild-ruby.bat; \
  chmod +x gambuild-ruby.bat; \
else \
[...]

PR #708, opened 2021-06-19
Fix gai_code_to_string type error when compiling under MINGW with --enable-cplusplus

Negdayen:
gambit/gambit#635

The implementation for gai_strerror seems to vary on Windows/MINGW, it can return a string encoded as UTF-16 depending on #define UNICODE.


feeley:
Thanks for the fix!

PR #710, opened 2021-06-26
Fix all byte-compiler warnings in gambit.el. Fix a docstring.

pierre-rouleau:
- Moved defvar before their use.
- Add forward declaration defvar.
- Replace deprecated or interactive functions/macros by appropriate ones.
- Declare XEmacs functions when in XEmacs but prevents warnings in Emacs.
- No change in behaviour.

### DIFF
--- a/misc/gambit.el
+++ b/misc/gambit.el
@@ -60,7 +60,7 @@
               (set-face-background face gambit-highlight-color)
               (progn
                 ;(make-face-bold face)
-                (set-face-underline-p face t))))
+                (set-face-underline face t))))
         (error (setq face nil)))
     face)
   "Face of overlay for highlighting Scheme expressions.")
@@ -80,6 +80,7 @@
 (require 'cmuscheme)
 
 ;;;----------------------------------------------------------------------------
+(defvar calculate-lisp-indent-last-sexp) ; forward declare, prevent warning
 
 (defun gambit-indent-function (indent-point state)
   (let ((normal-indent (current-column)))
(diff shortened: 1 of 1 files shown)


lassik:
I'm happy to help submit this into [MELPA](https://melpa.org/).

Have you looked at [geiser-gambit](https://melpa.org/#/geiser-gambit), and does `gambit.el` have features that are missing from it?

pierre-rouleau:
@lassik Hi!   I'm quite new to Gambit so I can't answer your question.  I was going to look into geyser-gambit but I have not done it yet.  What I'm doing is integrating support for several languages into my system [PEL](https://github.com/pierre-rouleau/pel#readme) and I was also going to start looking for adding iMenu support for Gambit as well as some others like Gerbil.  I can't object to support to MELPA, but I personally do not need it (for files not turned into elope package my system copies them in a directory that is in my load-path).

pierre-rouleau:
Also I should have mentioned that I did not provide backward compatibility to old versions of Emacs that do not support:
- ``with-current-buffer`` : introduced in Emacs 20.1
- ``set-face-underline``  : introduced on 1999-07-21,  replaced set-face-underline-p which was made obsolete in Emacs 24.4

I assumed these are old enough so changing the code would not cause problems.

pierre-rouleau:
I am also planning to replace the user definable defvar by customizable defcustom in a gambit defgroup. 
Is there any objection?

lassik:
> I am also planning to replace the user definable defvar by customizable defcustom in a gambit defgroup.
> Is there any objection?

No objection from me.

feeley:
I'm not an elisp expert, so I don't know what that means... but if it works with "old" (~5 years ago) releases of emacs I'm ok with it.

PR #712, opened 2021-07-07
Implement SRFI 219 (Define higher-order lambda)

lassik:
https://srfi.schemers.org/srfi-219/srfi-219.html

This implementation doesn't work because the syntax-rules macro cannot use the renamed `native-define` at the expansion site. There's probably a way to implement it (perhaps using `define-macro`). Note that the SRFI exports a `define` which shadows / conflicts with the native `define` from the `(scheme base)` library.


feeley:
I'm not sure what issue you are having but probably the best thing is to use `##define` as the "native `define`".

lassik:
It works! Tests pass.

lassik:
Should `219.scm` and/or `219#.scm` be made?

feeley:
Yes that would be best.

lassik:
How is the naming conflict between the `##` version of `define` and the 219 version of `define` handled in them? They cannot place the new `define` in the `##` namespace.

gambiteer:
This is cool.  I look forward to it working.  Brad

lassik:
I tried to implement using namespaces by following the example of SRFI 26 (`cut`/`cute`), but it doesn't work. I get the error:

```
$ gsi lib/
> (import (srfi 219))
> (define ((foo a) b) (string-append a b))
*** ERROR IN (stdin)@2.2 -- Unbound variable: srfi/219#define
```

feeley:
The problem is the file `219.sld` which lacks the definition of the `define` macro.  It should be:

```
(define-library (srfi 219)
  (export define)
  (include "219#.scm"))
```

lassik:
The latest commit fixes the problem, and now passes tests and works in `gsi` scripts and REPL.

`gsc -exe foo.scm` works when `foo.scm` is:

```scheme
(import (srfi 219))
(define ((greet-with-prefix prefix) suffix)
  (string-append prefix " " suffix))
(define greet (greet-with-prefix "hello"))
(display (greet "world"))
(newline)
```

but the compiled executable doesn't work (exit code 70 with no error message; `-:d+` `-:d-` `-:dc` `-:dr` do not show an error either).

feeley:
Did you compile with `-nopreload`? That's typically the problem...

lassik:
`gsc -exe -nopreload foo.scm` solves the problem.

feeley:
Great!

PR #714, opened 2021-07-17
Start (gambit unstable) library with make-inexact-real

lassik:
The macro `make-inexact-real` is copy-pasted from `_num#.scm` with minor adjustments.

The current commit doesn't work:

```
gsi/gsi lib/
Gambit v4.9.3-1472-g3bcb3a7f

> (import (gambit unstable))
> make-inexact-real
*** ERROR IN (stdin)@2.1 -- Unbound variable: make-inexact-real
```


lib/gambit/unstable/unstable#.scm
```diff
@@ -0,0 +1,17 @@
+;;;============================================================================
+
+;;; File: "promise#.scm"
```
feeley, line 3: should be `unstable#.scm`

lib/gambit/unstable/unstable.scm
```diff
@@ -0,0 +1,62 @@
...
+          (begin
+            (vector-set! t i (exact->inexact (expt 10 i)))
+            (loop (- i 1)))))
+
+    `',(list->f64vector (vector->list t))))
+
+(define exact-10^n-table (macro-make-exact-10^n-table))
+
+(define-prim&proc (make-inexact-real (negative? boolean)
+                                     (mantissa uinteger)
```
feeley, line 38: There is no type checker for "uinteger"...  instead use "nonnegative-exact-integer" or accept any integer and xor the sign with the `negative?` flag.

lib/gambit/unstable/unstable.scm
```diff
@@ -0,0 +1,62 @@
...
+                                     (exponent uinteger))
```
feeley, line 39: There is no type checker for "uinteger"... instead use "nonnegative-exact-integer".

lib/gambit/unstable/unstable.scm
```diff
@@ -0,0 +1,62 @@
...
+
+
+  (let ((n (if (and (##fixnum? mantissa)
```
feeley, line 42: Avoid using the `##` prefix... that's one of the reasons `define-prim&proc`.  A reference to `fixnum?` will be mapped to `##fixnum?`, and `(primitive (foo a b))` is equivalent to `(##foo a b)`.

lassik:
Fixed, but still getting `Unbound variable: make-inexact-real` after importing the library.

lib/gambit/unstable/unstable.scm
```diff
@@ -0,0 +1,60 @@
...
+            (vector-set! t i (exact->inexact (expt 10 i)))
+            (loop (- i 1)))))
+
+    `',(list->f64vector (vector->list t))))
+
+(define exact-10^n-table (macro-make-exact-10^n-table))
+
+(define-prim&proc (make-inexact-real (negative? boolean)
+                                     (mantissa nonnegative-exact-integer)
+                                     (exponent nonnegative-exact-integer))
```
feeley, line 39: Sorry, the exponent should be simply "exact-integer".

Moreover, maybe the `negative?` parameter should come last and default to `#f`.  This would make even more sense if `mantissa` is allowed to be a negative integer (either in the first version of the API or somewhere down the road).

Also the `exponent` could default to `0`, so `(make-inexact-real 123)` would give `123.0`.  This would be a good motivation for accepting negative integers so that `(make-inexact-real -123)` would give `-123.0`.

lib/gambit/unstable/unstable.scm
```diff
@@ -0,0 +1,60 @@
...
+      (if (not (< i 0))
+          (begin
+            (vector-set! t i (exact->inexact (expt 10 i)))
+            (loop (- i 1)))))
+
+    `',(list->f64vector (vector->list t))))
+
+(define exact-10^n-table (macro-make-exact-10^n-table))
+
+(define-prim&proc (make-inexact-real (negative? boolean)
```
feeley, line 37: I'm not sure why `make-inexact-real` is not accessible.  Compiling with `gsc -expansion ...` will reveal what name is actually being used.

lib/gambit/unstable/unstable.scm
lassik: True. If both `mantissa` and `exponent` are allowed to be negative, and the separate `negative?` argument is dropped, that would be ideal IMHO.

lassik: Let's revisit the issue of which arguments to make optional once we've agreed on the full range of available arguments. If we're turning this into a SRFI, an argument for the R7RS precision (`e`, `s`, `f`, `d`, `l`) should be added.

lib/gambit/unstable/unstable.scm
lassik: `gsc -expansion unstable.scm` shows that `(define-prim&proc (make-inexact-real ...) ...)` is not expanded, even if changed to `##define-prim&proc`.

lib/gambit/unstable/unstable.scm
feeley: You can't eliminate the `negative?` parameter because of `-0.0`.

lib/gambit/unstable/unstable.scm
feeley: Of course! There's a `(##include "~~lib/_gambit#.scm")` needed at the top of the file because `define-prim&proc` is a macro defined there.

lassik: Right. `(include "unstable.scm")` was also needed instead of `(include "unstable#.scm")`.

lassik:
I perused the old draft at https://github.com/pre-srfi/make-float, and according to it, `mantissa exponent precision` should be the only arguments we need. In that case we could make `exponent` and `precision` optional.

lib/gambit/unstable/unstable.scm
```diff
@@ -0,0 +1,50 @@
...
+           (fx< exponent
+                (f64vector-length exact-10^n-table)))
+      (if (fx< exponent 0)
+          (fl/ (fixnum->flonum mantissa)
+               (f64vector-ref exact-10^n-table
+                              (fx- exponent)))
+          (fl* (fixnum->flonum mantissa)
+               (f64vector-ref exact-10^n-table
+                              exponent)))
+      (exact->inexact (* mantissa (expt 10 exponent)))))
```
feeley, line 48: The calls `(primitive (exact-int.* mantissa ...))` and `(primitive (exact-int.expt 10 exponent))` would be better/faster here.  They also require fewer library dependencies (i.e. only the bignum specific operations rather than generic operations that require complex number support).

lib/gambit/unstable/unstable.scm
```diff
@@ -0,0 +1,50 @@
...
+        (loop (- i 1) (cons (exact->inexact (expt 10 i)) t))
+        `',(list->f64vector t))))
+
+(define exact-10^n-table (macro-make-exact-10^n-table))
+
+(define-prim&proc (make-inexact-real (mantissa exact-integer)
+                                     (exponent exact-integer 0)
+                                     (precision object #f))
+  (if (and (fixnum? mantissa)
+           (##fixnum->flonum-exact? mantissa)
```
feeley, line 35: `(primitive (fixnum->flonum-exact? mantissa))`

lib/gambit/unstable/unstable.scm
```diff
@@ -0,0 +1,50 @@
...
+  (define max-exact-power-of-10 22)
+  ;; (floor (inexact->exact (/ (log (expt 2 (macro-flonum-m-bits-plus-1)))
+  ;;                           (log 5))))
+
+  (let loop ((i max-exact-power-of-10) (t '()))
+    (if (>= i 0)
+        (loop (- i 1) (cons (exact->inexact (expt 10 i)) t))
+        `',(list->f64vector t))))
+
+(define exact-10^n-table (macro-make-exact-10^n-table))
```
feeley, line 29: This should be moved into `make-inexact-real` to avoid using a global variable.  On the other hand, the same table is used in `lib/_num.scm` so maybe making it a global in `lib/_num.scm` would be a better choice in the future.

lassik:
There should be a way to make negative zero as well.

lib/gambit/unstable/unstable.scm
lassik: The whole routine could be put into `lib/_num.scm`, which could be re-exported from one of the `(gambit ...)` libraries, as well as a future `(srfi ...)` library.

lassik:
I fixed the code as above, and brought back the `negative?` argument since there's no obvious way to make negative zero otherwise.

lassik:
+/- infinity and NaN still cannot be made (at least not without abusing arcane knowledge about the float representation), but they can be entered as literals into a parser routine; a parser should have no need to call `make-inexact-real` to make infinities and Nans.

lassik:
@feeley As far as I can tell, the PR in its current form works and solves all the above issues. What do you think?

feeley:
In the interest of moving forward, and because it is "unstable" I've decided to accept the PR, even though I'm not sure what is the purpose of this procedure.  Also, up to now the `(gambit ...)` libraries have all contained procedures that are statically linked with the runtime library, and `(gambit unstable)` does not fit.  So maybe the statically linked procedures should be put in `(_gambit ...)` and optional libraries that are Gambit specific should go into `(gambit ...)`.  It could also be `(gambit _XXX)` vs `(gambit XXX)`.

lassik:
The point is that user-written parsers can turn text into numbers using the same numerical semantics that the underlying implementation of Scheme `read` is using, without incurring the complexities of `read` or `number->string` on numbers parsed from a non-Scheme syntax. For example, a JSON parser can gather all the digits from a JSON number into a Scheme integer, then call `make-inexact-real` on that integer.

The point about static linking is too sophisticated for me to understand, but in general a `_` prefix carries a strong connotation of "internal -- do not use" to C programmers.

feeley:
The main problem with the `make-inexact-real` API is that it requires bignum support.  Some Schemes don't have bignum support and for those that do this could create a dependency to a large library of code.  An API based on strings may avoid the bignum library dependency at the cost of custom code to operate on large integers (I'm not sure how many bits would be required).

For the `(gambit ...)` libraries I now think the underscore prefix is not really useful.  The fact that some of the libraries are statically linked can be seen as an "implementation detail" that doesn't have to be exposed in the naming scheme.  What is important is that `(import (gambit X))` works for all available `X`.

lassik:
Does `make-inexact-real` require bignum support because a Scheme implementation can support more numerical range in flonums than in fixnums, which would mean that not every flonum can be represented by a fixnums mantissa and exponent? The string API could be a good idea, e.g. parse all digits `0`..`9` from the string and ignore all other characters (e.g. `.` and `,` and `_`) , assuming the string has been pre-validated for correct syntax.

The Gambit manual could have a table showing which libraries are statically linked, and which are not.

feeley:
The fundamental problem is that the `make-inexact-real` API needs on the order of 1024 significant bits for the `mantissa` parameter to correctly round the final inexact real:

```
(define expo 1023)
(define mant 53)
(define big (expt 2 expo))
(define x1 (+ big (expt 2 (- expo mant))))
(define x2 (+ x1 1))
(pp (exact->inexact x1))
(pp (exact->inexact x2))
(pp x1)
(pp x2)
;; output:
;; 8.98846567431158e307
;; 8.988465674311582e307
;; 89884656743115805365666807213050294962762414131308158973971342756154045415486693752413698006024096935349884403114202125541629105369684531108613657287705365884742938136589844238179474556051429647415148697857438797685859063890851407391008830874765563025951597582513936655578157348020066364210154316532161708032
;; 89884656743115805365666807213050294962762414131308158973971342756154045415486693752413698006024096935349884403114202125541629105369684531108613657287705365884742938136589844238179474556051429647415148697857438797685859063890851407391008830874765563025951597582513936655578157348020066364210154316532161708033
```

Note that here the `x1` and `x2` exact integers are about 1024 bits long and only differ in the least significant bit.  But that last bit is enough to make `x2` slightly closer to the next higher up IEEE754-64 number.

I'm not sure how this can be done without bignums but clearly some programming languages can convert the string representation of a floating point number to the IEEE754-64 value with correct rounding and without a full blown bignum library.

lassik:
OpenBSD uses [a `strtod()` routine written by David Gay](https://github.com/openbsd/src/blob/2207c4325726fdc5c4bcd0011af0fdf7d3dab137/lib/libc/gdtoa/strtod.c). It doesn't look easy. His routine is mentioned as being similar to this paper by the Chez folks: https://legacy.cs.indiana.edu/~dyb/pubs/FP-Printing-PLDI96.pdf

lassik:
For the benefit of things like TinyScheme, it would be good if `make-inexact-real` takes a string that can be passed to `strtod()`, either as-is, or by concatenating substrings into a suitable string.

lassik:
Proof of concept code:

```C
#include <stdio.h>
#include <stdlib.h>

static size_t copy_sign(char **dst, const char *str, size_t strpos,
                        size_t strlim) {
  if (str[strpos] == '+')
    strpos++;
  else if (str[strpos] == '-')
    *(*dst)++ = str[strpos++];
  return strpos;
}

static size_t copy_digs(char **dst, const char *str, size_t strpos,
                        size_t strlim) {
  while (strpos < strlim) {
    int ch = str[strpos];
    if ((ch >= '0') && (ch <= '9'))
      *(*dst)++ = str[strpos];
    strpos++;
  }
  return strpos;
}

double make_inexact_real(const char *str, size_t int_start, size_t frac_start,
                         size_t exp_start, size_t limit) {
  char buf[64];
  char *dst = buf;
  size_t b = 0;
  size_t i = 0;

  i = int_start;
  i = copy_sign(&dst, str, i, frac_start);
  i = copy_digs(&dst, str, i, frac_start);
  *dst++ = '.';
  i = copy_digs(&dst, str, i, exp_start);
  *dst++ = 'e';
  i = copy_sign(&dst, str, i, limit);
  i = copy_digs(&dst, str, i, limit);
  *dst = '\0';
  printf("evaluating strtod(\"%s\", NULL)\n", buf);
  return strtod(buf, NULL);
}

int main(void) {
  printf("the value is %e\n",
         make_inexact_real("-123_4_5_6.7_8e-90", 0, 10, 14, 17));
  return 0;
}
```

feeley:
I agree that an implementation that doesn't require bignum support would be nice.  This would require an API based on strings.  But this opens up a big can of worms... should the API accept `.` and `,` (for the decimal point), what to do with `_`, and many other things.  Also I would like to avoid using `substring` all over the place because this slows things down and causes a lot of garbage to be allocated.

It is a bit sad to "dumb down" the API (to strings) when Scheme has bignums, but I see advantages (support for tiny Schemes and also even big Schemes with an intelligent static linker could notice that the bignum library is not needed).

lassik:
In my experience, every syntax is prone to have a slightly different set of conventions for separators, delimiters, etc. The above C code ignores all characters except decimal digits (and +/- sign at the start of the integer/exponent part) which means the string has to be pre-validated before passing it in.

We could also accept the same syntax that C's `strtod()` accepts, but I suspect that `strtod()` implementations are not consistent in how they handle junk characters.

feeley:
Perhaps what is needed is for `make-inexact-real` to accept a string and be very tolerant of the syntax (basically allow many variations on the FP syntax which would have to be documented precisely).  That way in many circumstances there is no need to extract parts of the original representation to pass them to `make-inexact-real`.  This would only be needed in unusual cases.

feeley:
Yes your POC seems to be in the spirit of what I'm suggesting.  However, there would have to be a pure Scheme version of this too.

feeley:
For completeness the inverse function (for doing output) would also be required.

lassik:
Would this family of procedures work out:

* `(parse-unsigned-integer string start end)`
* `(parse-signed-integer string start end)`
* `(parse-inexact-real string int-start frac-start exp-start end)`

All of them would ignore characters other than `0`..`9`. `parse-signed-integer` and `parse-inexact-real` would additionally accept a `+` or `-` in the right place.

lassik:
Output is tricky since it has to adhere to specific conventions which vary. The formatted output SRFIs (there are about 3-5 of varying complexity) can display floats, and `number->string` will work if the precise output format doesn't matter.

feeley:
Don't use the term `parse` if no errors are detected... I would use `(string->nonnegative-integer string [start [end]])` which always returns a nonnegative integer (in particular 0 when the string contains no digits), etc

feeley:
Actually you could only have `(string->integer string [start [end]])` and the `+` and `-` signs would be interpreted appropriately.  So when there is no `+` or `-` sign you will get a nonnegative integer.  Also, when there is more than one `+` or `-` then the "last one" wins... keep it simple because the parsing and error checking is done before calling these procedures.

lassik:
Parsing signed and unsigned using the same procedure is a good idea, because Scheme doesn't have separate `signed int` and `unsigned int` types. There are many situations where you only want unsigned integers even in Scheme, but since these procedures require pre-validated input anyway, it doesn't matter.

feeley:
Hmmm... actually `(string->exact-integer string [start [end]])` is a better name, in particular if there's also `(string->inexact-real string [start [end]]`.  Interestingly, `string->inexact-real` can simply find the location of the exponent marker and decimal point and then use `string->exact-integer` to extract 2 integers and do the same bignum-based computation of the original `make-inexact-real` (for a pure-Scheme implementation).

lassik:
Yes, I orginally separated the signed and unsigned procedures because they can serve as subroutines of `make-inexact-real`.

Here's a first cut:

```Scheme
(define (string->integer string start end)
  (let loop ((i start) (magnitude 0) (neg? #f))
    (if (< i end)
        (let* ((c (string-ref string i))
               (d (digit-value c)))
          (cond (d (loop (+ i 1) (+ d (* 10 magnitude)) neg?))
                ((char=? c #\+) (loop (+ i 1) magnitude #f))
                ((char=? c #\-) (loop (+ i 1) magnitude #t))
                (else (loop (+ i 1) magnitude neg?))))
        (if neg? (- magnitude) magnitude))))

;; (string->integer "- 123_4_abc5_65a" 0 14) => -123456
```

feeley:
An amusing thought (not a real proposal)... `string->exact-integer` could return a one's complement representation of the number (for negative integers).  So `"-0"` would be converted to `-1` and `"-123"` would be converted to `-124`.  This would allow knowing if there was a negative sign before the integer (useful for `string->inexact-real`).  Yay for one's complement!

Your first cut is simple... but rather inefficient for large integers... a divide-and-conquer approach is better (see `substring->uinteger-aux` in `lib/_num.scm`).  However it isn't clear how to write a divide and conquer algorithm when some characters are not digits.

lassik:
Maybe `string->integer` should return the sign and the magnitude as two values, the way `truncate/` returns the quotient and the remainder.

The divide-and-conquer could work fine if we can assume a uniform distribution of digits in each substring (and accept bad performance in case some substrings have few to no digits). This should correspond to real usage (e.g. using `_` to separate digit groups at regular intervals, and `substring->uinteger-aux` looks like it can handle that.

feeley:
Nah... `string->exact-integer` should just return a single value.  The overhead of multiple values (both run time performance and code complexity) is not worth it for just detecting minus zero... (which is better done when parsing the FP number representation, which has to be done anyway to know where the decimal point and exponent part are).

For reference, with a 10000 digit string of digits, Gambit's `string->number` which uses a divide-and-conquer algorithm is 20x faster than the code you gave above.

A divide-and-conquer approach can be implemented by accumulating digits into a fixnum variable and when k digits (say k=8) are accumulated then add them to a list, then do the DAC on the list.  So a number of digits counter and a list of fixnums needs to be added to your code, and a fast path return when the list of fixnums is empty.

lassik:
Perhaps the integer parser should ignore the sign altogether? It complicates the API and slows down the parser.

Since the strings have to be pre-validated, the validator can easily remember whether or not there was a sign, and store it in a separate boolean variable. Negating a bignum is a fast operation, right?

feeley:
Also for the name of these procedures I would add the qualifier `tolerant`.  This is because `string->number` is already defined to do a check of the syntax, so it is reasonable to expect `string->exact-integer` to have a similar API and also do this check.  So the procedures would be

```
(tolerant-string->exact-integer string [start [end]])
(tolerant-string->inexact-real string [start [end]])
```

The handling of the sign by `tolerant-string->exact-integer` is for generality (so it can handle negative integers without having to define an extra procedure for this) and also for consistence with `tolerant-string->inexact-real` which does handle the sign.  The performance difference is negligeable (for short strings most of the time will be spent in `digit-value` to determine if the Unicode character is a digit or not, and its value).

lassik:
Tolerant is a good idea. "Lenient" is a slightly more precise word; a web search for "lenient parser" finds many things, including the famous BeautifulSoup HTML parser.

Consistency in sign handling is a good point.

How does the inexact-real scanner find where the fractional and exponent start? I think passing the substring positions is a reasonable solution; setting a position equal to the next position indicates a zero-length substring, which is handled like the numerical value `0`.

feeley:
`lenient` is good (and shorter!).  I'm not sure what you mean concerning finding fractional and exponent parts... the parser just goes from left to right and finds where each part starts/ends.  Read `lib/_num.scm` for the details which are quite complex because it supports rational, complex numbers and even the "old" `123####` syntax.

lassik:
That means it needs to recognize `.`, `e` and `E`. In particular, the decimal point (period vs comma) is not agreed upon internationally, though most programming languages do use a period. In human communication, a comma is used in about half of all countries worldwide IIRC. Passing the substring positions means we can delegate all these concerns to the caller.

feeley:
Ok... you mean that the API of `lenient-string->inexact-real` would accept the position of the decimal point and exponent marker as parameters?  My position is that this procedure should be "lenient" and accept different variations (both `.` and `,` would be acceptable decimal points).  Perhaps I need to revise this somewhat because `,` is sometimes also used as a thousand separator...  So maybe for this case an extra parameter could be given: the character to use as a decimal point (other characters would be ignored).

lassik:
True, `(lenient-string->inexact-real string separator start end)` would work, but then it's hard for it to use the integer parser as a subroutine.

lassik:
The thing is that any number validator has to find the exact substring start and end positions anyway. So it knows them already, and could just pass them in.

feeley:
Yes.  So one option would be to add an optional parameter that is the exponent (of 10) "bias".  So it would be set to -k where k is the number of decimals, and `lenient-string->inexact-real` ignores the decimal point.

lassik:
If the integer parser takes `start end` offsets, the float parser would only need two extra offsets: `start frac-start exp-start end`. If a char or bias argument is added, that alternative has only one less arg, and all the args are no longer string offsets.

feeley:
Actually, the higher level parser would first extract the exponent part and find k, then call `lenient-string->inexact-real` on the string without the exponent part.  So the `lenient-string->inexact-real` API could completely ignore the presence of an exponent part.

feeley:
This is actually quite neat because both `lenient-string->inexact-real` and `lenient-string->exact-integer` would only "convert" an integer (sequence of digits with a possible sign), but `lenient-string->inexact-real` would multiply by a power of 10 (given in an extra parameter).

lassik:
A typical validator keeps track of the string offset as it runs through the string. It's simple to add code to save important offsets in variables. This would obviate the need to calculate `k` and do a separate conversion of the exponent. It could simply save the offsets and pass them to `lenient-string->inexact-real ` (when there's no exponent, the same offset is passed for `exp-start` and `end`).

feeley:
I think the API I propose is simpler and doesn't leak the implementation.

feeley:
Moreover it is possible to only define a single procedure and make it so it's API supports the Scheme "exact real" syntax:

```
> #e1.234e5 
123400
```

The idea is to pass the exactness through the exponent parameter and have it default to 0.  An exact exponent parameter will return an exact result and an inexact exponent will return an inexact real:

```
> (lenient-string->real "- 1_2,,3.45")
-12345
> (lenient-string->real "- 1_2,,3.45" 2)
-1234500
> (lenient-string->real "- 1_2,,3.45" -2.0)
-123.45
```

lassik:
Small Schemes that don't have bignums probably also don't have ratios. Would this truncate exact reals into integers?

feeley:
I guess that could be implementation dependent.  The "integer" expressed by the string (without taking the exponent parameter into account) would have to fit in a number supported by the Scheme system, and the final result too.

lassik:
POSIX also promises that `strtod()` does range checking: https://pubs.opengroup.org/onlinepubs/9699919799/functions/strtod.html#tag_16_587_04 If the exponent is parsed from the string the same thing as the other info, Schemes that rely on libc can get range checking for free.

feeley:
For the case where the exponent bias is inexact a call to `strtod` is a possible implementation (after rewriting the string and adding the exponent part).  Note that the exponent part would be restricted to (exact or inexact) integers.  This is the same constraint as the parameters of `quotient`:

```
> (quotient 10 3)
3
> (quotient 10 3.0)
3.
> (quotient 10 3.5)
*** ERROR IN (stdin)@7.1 -- (Argument 2) INTEGER expected
(quotient 10 3.5)
```

lassik:
Good point. Hmm. Maybe we should try to implement each case and see how it goes.

PR #730, opened 2022-01-11
First impressions of a JavaScript six wrapper to the DOM API.

belmarca:
Such a module is desirable and this is a first step in this direction.


belmarca:
Squash the commits, no need to keep that history.

lib/_ffi/js/dom.scm
```diff
@@ -7,25 +7,25 @@
 
 ;;;============================================================================
 
-(define (select selector)
+(define (querySelector selector)
```
feeley, line 10: The naming is not consistent.  It should be called `document.querySelector` and there should be a seperate procedure `(querySelector elem selector)`.  Same for `getElementById` and `createElement`.  It might make sense to have an exported `document` variable that is initialized to `\document` so this can be done: `(createElement document "div")`.

lib/_ffi/js/dom.scm
```diff
@@ -7,25 +7,25 @@
...
 
-(define (insert-adjacent-html elem pos html)
+(define (insertAdjacentHTML elem pos html)
   \(`elem).insertAdjacentHTML(`pos, `html))
 
-(define (log msg)
+(define (console.log msg)
   \console.log(`msg))
 
 (define (alert msg)
```
feeley, line 31: Should add `prompt` and `confirm` too.  It would add convenience to allow any number of parameters and handle them the same way as `print`:

```
(define (alert . rest)
  \alert(`(with-output-to-string (lambda () (apply print rest))))))
```

feeley: A bit better implementation:

```
(define (alert . rest)
  \alert(`(with-output-to-string (lambda () (##print-aux rest)))))
```

belmarca:
Closing for staleness. Will reopen in due time.

PR #732, opened 2022-01-20
fixing geiser:load-file

targit:
The load function wasn't properly imported into the _geiser-namespace,
thus load-file didn't correctly.
I wonder if there is a reason (besides speed) why the code uses only primitive functions,
which load clearly is not.

### DIFF
--- a/lib/_geiser/_geiser.scm
+++ b/lib/_geiser/_geiser.scm
@@ -13,6 +13,7 @@
 
 (##namespace ("_geiser#"))                ;; in _geiser#
 (##include "~~lib/gambit/prim/prim#.scm") ;; map fx+ to ##fx+, etc
+(##namespace ("" load))                   ;; a non-primitive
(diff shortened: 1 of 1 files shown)


feeley:
The code uses primitives probably to isolate the geiser module from side-effects to the global variables containing predefined procedures.  For example, if a program does `(set! load (lambda x (println "hello!")))` then loading files will no longer work.  It would be possible to use the `##load` procedure directly to avoid this problem, but it has many parameters:

```
(define (##load
         path-or-settings
         script-callback
         clone-cte?
         raise-os-exception?
         linker-name
         quiet?)
  ...)
```

The `load` procedure is defined like this:

```
(define (load path-or-settings)
  (macro-force-vars (path-or-settings)
    (##load path-or-settings
            (lambda (script-line script-path) #f)
            #t
            #t
            #f
            #f)))
```

targit:
Thank you for that detailed explanation.

PR #733, opened 2022-01-25
Cygwin: Unbreak build

okuoku:
Fix Cygwin build by:

- Include `<cygwin/fs.h>` required to use `renameat2`
- Use `.unix` version of `bin/` script


okuoku:
Currently, cygwin build require `--enable-debug` to run `gsc` compiled executable correctly at Cygwin 3.3.3. I'm not sure #192 still relevant but still investigating.

Issue #735, opened 2022-02-02
universal backend: can't load js code from a worker


sublambda:
the browser implementation of ##os-load-object-file (_univlib.scm) relies on dom access to load js code, so it will fail in a worker. 

proposed fix:
```
--- a/lib/_univlib.scm
+++ b/lib/_univlib.scm
@@ -4170,13 +4170,21 @@ def @os_condvar_select@(devices_scm, timeout_scm):
         @trampoline@(@r0@);
       }
 
-      var script = document.createElement('script');
-      script.type = 'text/javascript';
-      script.src = path;
-      script.onload = onload;
-      script.onerror = onerror;
-      document.head.append(script);
-
+      if (self.document === undefined) {
+        try {
+          self.importScripts(path);
+          onload();
+        } catch(e) {
+          onerror();
+        }
+      } else {
+        var script = document.createElement('script');
+        script.type = 'text/javascript';
+        script.src = path;
+        script.onload = onload;
+        script.onerror = onerror;
+        document.head.append(script);
+      }
       return 0; // ignored
     }
   } else {
```

sublambda:
`self.document` is a common way to test; another is to check for existence of `self.DedicatedWorkerGlobalScope`.

PR #736, opened 2022-02-02
enable ##os-load-object-file (_univlib.scm) for js workers

sublambda:
re issue https://github.com/gambit/gambit/issues/735


feeley:
Thanks for contributing to Gambit!

sublambda:
thanks for the quick merge and thank you for gambit, really great stuff!

PR #762, opened 2022-08-05
Update manual's TCP Client with TLS encryption example

vesten:
Make manual's TCP Client with TLS encryption example a more complete stand alone procedure (attr: @feeley 4Aug2022 gitter.im/gambit/gambit)


gambiteer:
I like it, but I'm going to let Marc review and approve it, I'm not an expert in this area.

vesten:
Thanks. And thanks for your review.

PR #766, opened 2022-08-31
Reference python expander in SIX makefile.

belmarca:
This is required for the library to be properly installed.


belmarca:
Fixes for the crashing tests will come in an upcoming PR.

PR #768, opened 2022-09-06
Try to fix CI issue

feeley:


.github/workflows/ci.yml
```diff
@@ -68,6 +68,12 @@ jobs:
         ./configure --enable-multiple-threaded-vms --enable-gambitdir=~~execdir/.. --prefix=$(pwd)/dist-mt
         echo "#undef HAVE_CLOCK_GETTIME" >> include/config.h
         make -j`nproc` core
+        gsi/gsi -v
+        make check
+        cd lib/_define-library
+        make modules
+        cd ../..
+        make modules
```
gambiteer, line 76: Do you really want `make modules` twice here?

feeley: Yes, the first `make modules` is done in `lib/_define-library` so it only builds that set of modules.  Then the other `make modules` will build the rest of the modules.

PR #783, opened 2022-09-21
Issue 650

Negdayen:
I think this is an improvement;  I see about half as many segfaults with this change and receive less warnings about race conditions from TSan.

### DIFF
--- a/lib/setup.c
+++ b/lib/setup.c
@@ -457,12 +457,14 @@ ___virtual_machine_state ___vms;)
 
 #define SYNC_INIT_MSG(var) var = SYNC_WAITING
 
-#define SYNC_GET_MSG(expr) \
+#define SYNC_GET_MSG(ps,field,var) \
 do { \
-     ___MUTEX_LOCK(___ps->sync_mut); \
-     while ((expr) == SYNC_WAITING)  \
-       ___sync_wait (___ps); \
-     ___MUTEX_UNLOCK(___ps->sync_mut); \
+     ___MUTEX_LOCK(ps->sync_mut); \
+     while (ps->field == SYNC_WAITING) \
+       ___sync_wait(ps); \
+     var = ps->field; \
+     SYNC_INIT_MSG(ps->field); \
+     ___MUTEX_UNLOCK(ps->sync_mut); \
    } while (0)
 
 #define SYNC_SEND_MSG(ps,field,val) \
@@ -528,8 +530,7 @@ ___sync_op_struct *sop_ptr;)
       int sid1;
       ___sync_op_struct sop1;
 
-      SYNC_GET_MSG(sid1 = ___ps->sync_id1);
-      SYNC_INIT_MSG(___ps->sync_id1);
+      SYNC_GET_MSG(___ps, sync_id1, sid1);
 
       sop1 = ___ps->sync_op1;
 
@@ -562,8 +563,7 @@ ___sync_op_struct *sop_ptr;)
           int sid2;
           ___sync_op_struct sop2;
 
-          SYNC_GET_MSG(sid2 = ___ps->sync_id2);
-          SYNC_INIT_MSG(___ps->sync_id2);
+          SYNC_GET_MSG(___ps, sync_id2, sid2);
 
           sop2 = ___ps->sync_op2;
 
@@ -625,8 +625,7 @@ ___sync_op_struct *sop_ptr;)
        * Wait for parent to reply with winning operation.
        */
 
-      SYNC_GET_MSG(sid = ___ps->sync_id0);
-      SYNC_INIT_MSG(___ps->sync_id0);
+      SYNC_GET_MSG(___ps, sync_id0, sid);
 
       sop = ___ps->sync_op0;
     }
--- a/lib/setup.c
+++ b/lib/setup.c
@@ -438,8 +438,6 @@ ___virtual_machine_state ___vms;)
  * Synchronous operations.
  */
 
-#define WASTE_TIME() ___CPU_RELAX()
-
 #define COMBINING_OP(op) ((op)&3)
 #define COMBINING_AND 1
 #define COMBINING_ADD 2
@@ -459,23 +457,12 @@ ___virtual_machine_state ___vms;)
 
 #define SYNC_INIT_MSG(var) var = SYNC_WAITING
 
-#define SYNC_GET_MSG_SPIN_TIMEOUT 2000
-
 #define SYNC_GET_MSG(expr) \
 do { \
-     int loops_left = SYNC_GET_MSG_SPIN_TIMEOUT; \
-     while ((expr) == SYNC_WAITING) \
-       { \
-         WASTE_TIME(); \
-         if (--loops_left == 0) \
-           { \
-             loops_left = SYNC_GET_MSG_SPIN_TIMEOUT; \
-             ___MUTEX_LOCK(___ps->sync_mut); \
-             if ((expr) == SYNC_WAITING) \
-               ___sync_wait (___ps); \
-             ___MUTEX_UNLOCK(___ps->sync_mut); \
-           } \
-       } \
+     ___MUTEX_LOCK(___ps->sync_mut); \
+     while ((expr) == SYNC_WAITING)  \
+       ___sync_wait (___ps); \
+     ___MUTEX_UNLOCK(___ps->sync_mut); \
    } while (0)
 
 #define SYNC_SEND_MSG(ps,field,val) \


Negdayen:
Oops... only the first commit seems like an improvement.  I mistakeningly didn't rebuild the second commit and just noticed it has performance issues.

Negdayen:
Thanks for the insights.

I've been experimenting with augmenting Gambit's lock implementations with Helgrind/TSan annotations so that the tools understand Gambit better.  But on further investigation, I agree I'm not sure how effective it will be.  For example, I am receiving a lot of warnings about one OS-level thread trying to unlock a lock owned by another OS-level when it comes to the primitive ticket locks, but I think this is fine, as the semantics for ticket locks aren't going to be the same as a mutex.  And if my understanding is correct, Scheme-level threads can migrate between the OS-level threads, so it should be normal for a lock to happen on one OS-level thread and the unlock to happen on a different OS-level thread, because it is actually the same Scheme-level thread.

feeley:
It is not just because of migration.  The Gambit VM "processors" (which are OS threads) use mutexes for synchronization between processors.  So processor 1 can lock a mutex, then try to lock it again (causing it to block) and it is processor 2 that unlocks this mutex to let processor 1 continue executing.  This pattern avoids having to use condition variables.

For the debugging of SMP, we should come up with a test that fails (segfault or other) and that is "as simple as possible".  Then we should work on finding out why this test fails.  Finding this minimal failing program is not easy, but here is a first try:

```
$ cat test1.scm
(declare (standard-bindings) (extended-bindings) (not safe))

(define (test)
  (let loop ((i 0) (result 0))
    (if (fx< i 1000000000)
        (loop (fx+ i 1) (cons i i))
        (pair? result))))

(define p (##get-parallelism-level))

(pp (list parallelism-level: p))

(define threads (map (lambda (id) (thread test)) (iota p)))

(pp (map thread-join! threads))
```

This program starts as many threads as the parallelism level (which can be changed to N with the `-:pN` runtime option).  Each thread allocates many pairs that are all garbage.  This exercices the garbage collector.  I have tested this program by compiling it first, so that any failure that could be caused by a bug in the interpreter (i.e. `eval`) is excluded.  What I observe on macOS is that the program does not fail when N=1, it occasionally fails with N=2, and it consistently fails with N>=3:

```
$ ./configure --enable-single-host --enable-multiple-threaded-vms --enable-smp CC=gcc-9;make clean;make -j8 core
$ make bootstrap;make bootclean;make -j8 core
$ gsc/gsc -:= -exe test1.scm
$ for i in 1 2 3 4 5 6 7 8 9 10; do ./test1 -:p1; done
(parallelism-level: 1)
(#t)
(parallelism-level: 1)
(#t)
(parallelism-level: 1)
(#t)
(parallelism-level: 1)
(#t)
(parallelism-level: 1)
(#t)
(parallelism-level: 1)
(#t)
(parallelism-level: 1)
(#t)
(parallelism-level: 1)
(#t)
(parallelism-level: 1)
(#t)
(parallelism-level: 1)
(#t)
$ for i in 1 2 3 4 5 6 7 8 9 10; do ./test1 -:p2; done
(parallelism-level: 2)
(#t #t)
(parallelism-level: 2)
(#t #t)
(parallelism-level: 2)
(#t #t)
(parallelism-level: 2)
(#t #t)
(parallelism-level: 2)
(#t #t)
(parallelism-level: 2)
(#t #t)
(parallelism-level: 2)
(#t #t)
Bus error: 10
(parallelism-level: 2)
(#t #t)
(parallelism-level: 2)
(#t #t)
(parallelism-level: 2)
(#t #t)
Segmentation fault: 11
$ for i in 1 2 3 4 5 6 7 8 9 10; do ./test1 -:p3; done
(parallelism-level: 3)
(#t #t #t)
Bus error: 10
(parallelism-level: 3)
(#t #t #t)
Bus error: 10
(parallelism-level: 3)
(#t #t #t)
Bus error: 10
(parallelism-level: 3)
Bus error: 10
(parallelism-level: 3)
(#t #t #t)
Bus error: 10
(parallelism-level: 3)
Bus error: 10
(parallelism-level: 3)
(#t #t #t)
Bus error: 10
(parallelism-level: 3)
(#t #t #t)
Bus error: 10
(parallelism-level: 3)
(#t #t #t)
Bus error: 10
(parallelism-level: 3)
(#t #t #t)
Bus error: 10
```

Note that in the case of N=2 it seems the program fails after printing the results, but with N=3 it can fail before printing the results.

To improve debugging it is useful to add some debugging options to the `./configure` command so that a backtrace will be printed and a history of the control flow is logged in `gambit.log`:

```
$ ./configure --enable-single-host --enable-multiple-threaded-vms --enable-smp --enable-debug --enable-debug-log --enable-debug-c-backtrace --enable-debug-ctrl-flow-history CC=gcc-9;make clean;make -j8 core
$ gsc/gsc -:= -exe test1.scm
$ ./test1 -:p2
(parallelism-level: 2)
(#t #t)
$ ./test1 -:p2
(parallelism-level: 2)
(#t #t)
$ ./test1 -:p2
(parallelism-level: 2)
(#t #t)
$ ./test1 -:p2
(parallelism-level: 2)
(#t #t)
$ ./test1 -:p2
(parallelism-level: 2)
(#t #t)
0   test1                               0x000000010eef6b94 crash_signal_handler + 37
1   libsystem_platform.dylib            0x00007ff811095dfd _sigtramp + 29
2   ???                                 0x0000000000000000 0x0 + 0
3   test1                               0x000000010eee0b50 ___trampoline + 36
4   test1                               0x000000010eee0e80 ___call + 643
5   test1                               0x000000010eee101f ___run + 249
6   test1                               0x000000010eee30e0 ___setup + 577
7   test1                               0x000000010eedd6a5 ___main + 7360
8   test1                               0x000000010eef7328 ___main_char + 139
9   test1                               0x000000010eed946a main + 41
10  dyld                                0x0000000113d5a51e start + 462
*** FATAL ERROR -- Process crashed with SIGBUS
$ cat gambit.log
fd=1 kind=159 direction=2
fd=2 kind=159 direction=2
Most recent control-flow history:

P0:
_kernel.c:12166:
_kernel.c:11777:
_thread.c:18711:
...
_thread.c:18287:
_kernel.c:22713:
_kernel.c:22716:
_thread.c:18304:

P1:
_thread.c:19621:
_thread.c:19621:
_thread.c:19621:
...
_thread.c:19621:
_thread.c:19621:
_thread.c:19621:
_thread.c:19621:
```

So we can see the program crashed in `___trampoline` with what seems to be a NULL pointer dereference.  The file `gambit.log` also reveals the history of the activity of each processor just before the crash.  So the last actions done by P0 and P1 were in `lib/_thread.c` at the lines indicated.  For example, at `_thread.c:18304` we find:

```
___DEF_SLBL(1,___L1__23__23_wait_21_)
   ___SET_R2(___CURRENTVM)
   ___PRIMITIVELOCK(___R2,___FIX(1L),___FIX(9L))
   ___SET_R2(___UNCHECKEDSTRUCTUREREF(___STK(-6),___FIX(17L),___FAL,___FAL))
   ___SET_R3(___UNCHECKEDSTRUCTUREREF(___STK(-6),___FIX(18L),___FAL,___FAL))
   ___UNCHECKEDSTRUCTURESET(___R3,___R2,___FIX(17L),___FAL,___FAL)
   ___UNCHECKEDSTRUCTURESET(___R2,___R3,___FIX(18L),___FAL,___FAL)
   ___SET_R2(___CURRENTVM)
   ___PRIMITIVEUNLOCK(___R2,___FIX(1L),___FIX(9L))
   ___ADJFP(-8)
   ___JUMPPRM(___NOTHING,___STK(1))
```

I suspect some issue with garbage collection that is messing up some code pointer on the stack (possibly the return address used in the instruction `___JUMPPRM(___NOTHING,___STK(1))` that causes `___trampoline` to segfault).  If this can be solved we can move on to more complex scenarios.

Negdayen:
Wow, yeah!  None of my test cases for this issue seem to segfault anymore with that change.

feeley:
Commit d0bf2c36fa0dd6c971cceecf49734fe965b50ae5 fixes the issue with the `___stack_overflow_undo_if_possible` optimization.  I think this fix will give a much better stability and allow easier debugging of the remaining SMP bugs.

On macOS I can now run the above test program with N<=2 with no crash (out of 100 runs), and 2 crashes out of 100 for N=4, and 4 crashes out of 100 for N=8.

Negdayen:
Opened https://github.com/gambit/gambit/pull/788 (couldn't figure out how to re-open with the change here).  Looks like there is still more issues though.

Negdayen:
Seems like there is still a stack issue.

The test program is producing SIGSEGV on line 18310 of _thread.c.  `___STK(-6)`
should be a pointer to a record, but is an invalid memory address
`0xffffffff85000010`.

```
18307	___DEF_SLBL(1,___L1__23__23_wait_21_)
18308	___SET_R2(___CURRENTVM)
18309	___PRIMITIVELOCK(___R2,___FIX(1L),___FIX(9L))
18310	___SET_R2(___UNCHECKEDSTRUCTUREREF(___STK(-6),___FIX(17L),___FAL,___FAL))
18311	___SET_R3(___UNCHECKEDSTRUCTUREREF(___STK(-6),___FIX(18L),___FAL,___FAL))
18312	___UNCHECKEDSTRUCTURESET(___R3,___R2,___FIX(17L),___FAL,___FAL)
18313	___UNCHECKEDSTRUCTURESET(___R2,___R3,___FIX(18L),___FAL,___FAL)
18314	___SET_R2(___CURRENTVM)
18315	___PRIMITIVEUNLOCK(___R2,___FIX(1L),___FIX(9L))
18316	___ADJFP(-8)
18317	___JUMPPRM(___NOTHING,___STK(1))
```

```
(gdb) print/x ___STK(-6)
$53 = 0xffffffff85000010
(gdb) x 0xffffffff85000010
0xffffffff85000010:	Cannot access memory at address 0xffffffff85000010
```

This should correspond to line 2553 in _thread.scm:

```
2540	(define-prim (##wait! devices timeout)
2541	
2542	  (##declare (not interrupts-enabled))
2543	
2544	  (let ((processor (macro-current-processor)))
2545	
2546	    (macro-lock-current-vm!)
2547	    (macro-processor-deq-insert-at-tail! (macro-current-vm) processor)
2548	    (macro-unlock-current-vm!)
2549	
2550	    (let ((code (##os-condvar-select! devices timeout)))
2551	
2552	      (macro-lock-current-vm!)
2553	      (macro-processor-deq-remove! processor)
2554	      (macro-unlock-current-vm!)
2555	
2556	      code)))
```

So, `___STK(-6)` should have been the pointer to the processor scheme object.
I examined the stack to see if any of the elements look like the processor scheme
object:

```
...
(gdb) print/x ___STK(-4)
0x5627a9353761
(gdb) x 0x5627a9353761
0x5627a9353761 <___gstate0+18017>:	0x000000b0
(gdb) print(18017 - sizeof(___processor_state_struct))
441
```

Although it seems like my calculations are one byte off, `___STK(-4)` corresponds to
aligned_pstate[1]->processor_scmobj.

Investigating more into where the value of `___STK(-6)` came from, I notice that
`___R1` has the same value:

```
(gdb) print/x ___R1
$63 = 0xffffffff85000010
```

Tracing backwards through the program to figure out where this value came from, it
seems like it is the Scheme error code object returned by `___device_select`.
This value is equivalent to `___FIX(___ERRNO_ERR(EINTR))`.

os_io.c:
```
1521	#ifdef USE_ASYNC_DEVICE_SELECT_ABORT
1522	
1523	  if (aborted && e == ___FIX(___NO_ERR))
1524	    e = ___FIX(___ERRNO_ERR(EINTR));
1525	
1526	#endif
```

```
(gdb) print/x ___FIX(___ERRNO_ERR(EINTR))
$72 = 0xffffffff85000010
```

I see this is assigned to ___R1 in _kernel.c on line 27721.

```
22716	___DEF_GLBL(___L__23__23_os_2d_condvar_2d_select_21_)
22717	#define ___ARG1 ___R1
22718	#define ___ARG2 ___R2
22719	{ ___SCMOBJ ___RESULT;
22720	___RESULT = ___os_condvar_select (___ARG1, ___ARG2);
22721	   ___SET_R1(___RESULT)
22722	}
22723	#undef ___ARG2
22724	#undef ___ARG1
22725	   ___JUMPPRM(___NOTHING,___R0)
```

So far I have been unable to determine how it got placed on the stack.  It
apparently happens sometime between a call to ___trampoline in setup.c and this
code in _thread.c:

```
#undef ___MD_ALL
#define ___MD_ALL ___D_HEAP ___D_FP ___D_R0 ___D_R1 ___D_R2 ___D_R3 ___D_R4 ___D_F64(___F64V1) ___D_F64( \
___F64V2) ___D_F64(___F64V3) ___D_F64(___F64V4)
#undef ___MR_ALL
#define ___MR_ALL ___R_HEAP ___R_FP ___R_R0 ___R_R1 ___R_R2 ___R_R3 ___R_R4
#undef ___MW_ALL
#define ___MW_ALL ___W_HEAP ___W_FP ___W_R0 ___W_R1 ___W_R2 ___W_R3 ___W_R4
___BEGIN_M_COD
___BEGIN_M_HLBL
___DEF_M_HLBL_INTRO
___DEF_M_HLBL(___L0___thread_23_)
___DEF_M_HLBL(___L1___thread_23_)
...
___END_M_HLBL
```

feeley:
I have found the problem and the analysis is interesting so it is useful to explain the details for the benefit of future debugging.  The procedure `##os-condvar-select!`, which is called when a processor needs to wait for a notification (of an I/O operation becoming possible, or a processor interrupt, or a synchronous operation) is currently written like this:

```
(define-prim (##os-condvar-select! devices timeout)
  (##declare (not interrupts-enabled))
  (##c-code
   "___RESULT = ___os_condvar_select (___ARG1, ___ARG2);"
   devices
   timeout))
```

With SMP (but not with the non-SMP Gambit) it is possible for a processor executing a call to `___os_condvar_select` to be interrupted by another processor to trigger a GC (which is a synchronous operation, i.e. all processors participate in performing the GC).  When a GC occurs, things can be moved around, including the stack (in other words the stack frame pointer may change to point to the new location of the stack).  However, for efficiency, the C code generated by Gambit uses a caching strategy to access critical fields of the processor state structure (pointed to by `___ps`), such as the GVM registers and stack frame pointer.  These fields are copied to local variables when the C host function is entered, and they are copied back to the processor state structure when the C host function returns.

In the current implementation of `##os-condvar-select!` the state of the frame pointer has been copied to the local C variable `___fp` and (when `--enable-single-host` is used) the same value will be written back to the processor state structure when the C host function returns.  But this is wrong when a GC was executed by `___os_condvar_select`.  The stack frame pointer will point to the old location of the stack and clearly everything will go bad from that point on.

The solution is to force a write-back of all the cached fields (using `___W_ALL`) before the call to `___os_condvar_select` and rereading the fields after (using `___R_ALL`).  The return address of `##os-condvar-select!` must also be stored and fetched from the stack frame for the GC to correctly scan the stack.  So this is the correct implementation to support SMP:

```
(define-prim (##os-condvar-select! devices timeout)
  (##declare (not interrupts-enabled))
  (##c-code "

   ___SCMOBJ result;

   ___FRAME_STORE_RA(___R0)
   ___W_ALL

   result = ___os_condvar_select (___ARG1, ___ARG2);

   ___R_ALL
   ___SET_R0(___FRAME_FETCH_RA)

   ___RESULT = result;"

   devices
   timeout))
```

Note that other procedures in `lib/_kernel.scm` already use this strategy when a GC is possible, for example `##sync-op-interrupt!`, `##check-heap`, `##gc`, `##make-vector`, etc.

I have tried this new definition of `##os-condvar-select!` and have not had a single crash of the test program over 100 executions.  The change is commit 4325936c97b189d5d760f734f6cf9288b273ceb6.  Please try it out.

Negdayen:
This changes the nature of the issue for me.  This new issue seems to appear more frequently with `-:p2`.

```
./test1(+0x299f56)[0x55c92bb7af56]
/usr/lib/libpthread.so.0(+0x13900)[0x7f7939bc8900]
./test1(___stack_limit+0x496)[0x55c92bb6f586]
./test1(+0x2bc5fe)[0x55c92bb9d5fe]
./test1(+0x27ef08)[0x55c92bb5ff08]
./test1(___trampoline+0x24)[0x55c92bb63754]
./test1(___call+0x303)[0x55c92bb63b04]
./test1(___run+0x125)[0x55c92bb63cef]
./test1(___setup+0x387)[0x55c92bb66886]
./test1(___main+0x1ad0)[0x55c92bb946d1]
./test1(___main_char+0x92)[0x55c92bb7b75b]
./test1(main+0x29)[0x55c92bb5dadf]
/usr/lib/libc.so.6(__libc_start_main+0xea)[0x7f7939a16e0a]
./test1(_start+0x2a)[0x55c92bb5d9ea]
*** FATAL ERROR -- Process crashed with SIGSEGV
```

After this crash, I need to `reset` my terminal, as what I type no longer appears on the terminal.

```
*** START OF DEBUGGING TRACES
...
P0:
test1.c:379:___DEF_GLBL(___L15_test)
test1.c:337:___DEF_GLBL(___L7_test)
...
_thread.c:16262:___DEF_GLBL(___L13__23__23_thread_2d_join_21_)
test1.c:275:___DEF_SLBL(13,___L13_test1_23_)
test1.c:266:___DEF_GLBL(___L18_test1_23_)
test1.c:293:___DEF_SLBL(15,___L15_test1_23_)
test1.c:297:___DEF_SLBL(16,___L16_test1_23_)
test1.c:293:___DEF_SLBL(15,___L15_test1_23_)
test1.c:297:___DEF_SLBL(16,___L16_test1_23_)
_kernel.c:11569:___DEF_SLBL(12,___L12__23__23_kernel_2d_handlers)
test1.c:269:___DEF_SLBL(11,___L11_test1_23_)
_kernel.c:10698:___DEF_SLBL(2,___L2__23__23_kernel_2d_handlers)

P1:
test1.c:356:___DEF_GLBL(___L11_test)
test1.c:360:___DEF_SLBL(4,___L4_test)
...
_thread.c:19021:___DEF_GLBL(___L19__23__23_thread_2d_schedule_2d_with_2d_acquired_2d_processor_21_)
_thread.c:18290:___DEF_GLBL(___L__23__23_wait_21_)
_kernel.c:22713:___DEF_SLBL(0,___L0__23__23_os_2d_condvar_2d_select_21_)
_kernel.c:22716:___DEF_GLBL(___L__23__23_os_2d_condvar_2d_select_21_)
```

[annotated-gambit.log](https://github.com/gambit/gambit/files/9715575/annotated-gambit.log)

feeley:
Can you confirm that crashes are less frequent with the `___os_condvar_select` fix?  Also, can you give details on your hardware (processor type and number of cores), operating system and C compiler?

For the above test program I observe no crashes anymore on macOS and debian linux with `-:p8` (over 100 executions).  This is not to say that there are no bugs left in SMP... in fact I have a second test that uses mutexes that crashes very easily, but that is probably caused by a separate bug and I am looking into that issue.

By the way, I have been using the following `crash-test.sh` script to execute several tests:

```
$ cat crash-test.sh 
#!/bin/sh

repetitions=$1

shift

runs=0
success=0
fail=0

while test $runs -lt $repetitions; do
    echo $runs
    if $* ; then
        success=$((success + 1))
    else
        fail=$((fail + 1))
    fi
    runs=$((runs + 1))
done

echo runs=$runs success=$success fail=$fail
$ ./crash-test.sh 5 ./test1 -:p8
0
(parallelism-level: 8)
(#t #t #t #t #t #t #t #t)
1
(parallelism-level: 8)
(#t #t #t #t #t #t #t #t)
2
(parallelism-level: 8)
(#t #t #t #t #t #t #t #t)
3
(parallelism-level: 8)
(#t #t #t #t #t #t #t #t)
4
(parallelism-level: 8)
(#t #t #t #t #t #t #t #t)
runs=5 success=5 fail=0
```

Please use that, or something similar, to try many executions.

feeley:
By the way, you may be observing crashes because of the mutex bug I mentioned given that `thread-join!` uses mutexes for the synchronization.  The mutex bug is triggered when two threads try to lock the mutex at the same time.  So in the above test program this is relatively unlikely, although it is still possible and maybe something in your setup (hardware, OS, ...) is making this situation more likely.

I have tried a larger number of executions of the test program with `./crash-test.sh 1000 ./test1 -:p2` and I only get 2 failures out of 1000 executions for 2 processors.  This suggests that those crashes are due to some other reason.

Negdayen:
18/1000 failures before commit, 13/1000 after commit using `-:p4`.  I do think it is an improvement as I haven't seen the original issue since.

This `___stack_limit` issue looks like it is present in earlier commits, I just happened to stumble across it after the fix.  Triggering it is highly dependent on the value of `i`.  Too low or too high and it won't trigger.  100000 is the sweet spot for my machines, where I can trigger it every third run or so.  Adding another zero decreases the rate to every 50 or so runs.

I'm currently using Void Linux across 3 machines.  Kernel version ranges from 5.13 to 5.18.

This machine seems to encounter a wider spectrum of issues, but for the `___stack_limit` issue I was able to reproduce on all three.
```
$ uname -a
Linux void 5.18.17_1 #1 SMP PREEMPT_DYNAMIC Fri Aug 12 13:53:18 UTC 2022 x86_64 GNU/Linux
```

```
$ lscpu
Architecture:            x86_64
  CPU op-mode(s):        32-bit, 64-bit
  Address sizes:         36 bits physical, 48 bits virtual
  Byte Order:            Little Endian
CPU(s):                  8
  On-line CPU(s) list:   0-7
Vendor ID:               GenuineIntel
  Model name:            Intel(R) Core(TM) i7 CPU         870  @ 2.93GHz
    CPU family:          6
    Model:               30
    Thread(s) per core:  2
    Core(s) per socket:  4
    Socket(s):           1
    Stepping:            5
    Frequency boost:     enabled
    CPU(s) scaling MHz:  41%
    CPU max MHz:         2934.0000
    CPU min MHz:         1200.0000
    BogoMIPS:            5851.88
    Flags:               fpu vme de pse tsc msr pae mce cx8 apic sep mtrr pge mca cmov pat pse36 clflush dts acpi mmx fxsr sse sse2 ht
                         tm pbe syscall nx rdtscp lm constant_tsc arch_perfmon pebs bts rep_good nopl xtopology nonstop_tsc cpuid aperf
                         mperf pni dtes64 monitor ds_cpl vmx smx est tm2 ssse3 cx16 xtpr pdcm sse4_1 sse4_2 popcnt lahf_lm pti tpr_shad
                         ow vnmi flexpriority ept vpid dtherm ida
Virtualization features:
  Virtualization:        VT-x
Caches (sum of all):
  L1d:                   128 KiB (4 instances)
  L1i:                   128 KiB (4 instances)
  L2:                    1 MiB (4 instances)
  L3:                    8 MiB (1 instance)
NUMA:
  NUMA node(s):          1
  NUMA node0 CPU(s):     0-7
Vulnerabilities:
  Itlb multihit:         KVM: Mitigation: VMX disabled
  L1tf:                  Mitigation; PTE Inversion; VMX conditional cache flushes, SMT vulnerable
  Mds:                   Vulnerable: Clear CPU buffers attempted, no microcode; SMT vulnerable
  Meltdown:              Mitigation; PTI
  Mmio stale data:       Not affected
  Retbleed:              Not affected
  Spec store bypass:     Vulnerable
  Spectre v1:            Mitigation; usercopy/swapgs barriers and __user pointer sanitization
  Spectre v2:            Mitigation; Retpolines, STIBP disabled, RSB filling, PBRSB-eIBRS Not affected
  Srbds:                 Not affected
  Tsx async abort:       Not affected
```

```
$ gcc -v
Using built-in specs.
COLLECT_GCC=gcc
COLLECT_LTO_WRAPPER=/usr/lib64/gcc/x86_64-unknown-linux-gnu/10.2.1/lto-wrapper
Target: x86_64-unknown-linux-gnu
Configured with: /builddir/gcc-10.2.1_pre1/configure --build=x86_64-unknown-linux-gnu --enable-gnu-unique-object --enable-vtable-verify --prefix=/usr --mandir=/usr/share/man --infodir=/usr/share/info --libexecdir=/usr/lib64 --libdir=/usr/lib64 --enable-threads=posix --enable-__cxa_atexit --disable-multilib --with-system-zlib --enable-shared --enable-lto --enable-plugins --enable-linker-build-id --disable-werror --disable-nls --enable-default-pie --enable-default-ssp --enable-checking=release --disable-libstdcxx-pch --with-isl --with-linker-hash-style=gnu --disable-sjlj-exceptions --disable-target-libiberty --enable-languages=c,c++,objc,obj-c++,fortran,lto,go,ada
Thread model: posix
Supported LTO compression algorithms: zlib
gcc version 10.2.1 20201203 (GCC)
```

Negdayen:
I managed to produce the original symptom again with the new commit, where `___STK(-6) = ___FIX(___ERRNO_ERR(EINTR))`.

At least in this run I am looking at in gdb, just after the call to `___os_condvar_select` and before the call to `___R_ALL`,  `___fp` and `___ps->fp` contain the same address and `___STK(-6)` no longer points to the processor object but to the error code object.

So the timing for this run for when the stack got changed is slightly different.

Negdayen:
Had a run this morning where `___FRAME_FETCH_RA` was setting `___R0` to an invalid address `0xffffffffffffffea`.  Seems like this should have been a pointer into `___lbltbl`.  The frame pointer `___fp` was not equal to `___ps->stack_break`, so it was executing the `___fp[-(-0)]` branch.  When `___FRAME_STORE_RA` ran, the frame pointer and stack break were also not equal, but `f[-(-0)]` was a pointer into `___lbltbl`.  `fp[-(-0)]` seemed to take on this invalid address after the call to `___os_condvar_select`.

```
 22716  ___DEF_GLBL(___L__23__23_os_2d_condvar_2d_select_21_)
 22717  #define ___ARG1 ___R1
 22718  #define ___ARG2 ___R2
 22719  { ___SCMOBJ ___RESULT;
 22720
 22721     ___SCMOBJ result;
 22722
 22723     ___FRAME_STORE_RA(___R0)
 22724     ___W_ALL
 22725
 22726     result = ___os_condvar_select (___ARG1, ___ARG2);
 22727
 22728     ___R_ALL
 22729     ___SET_R0(___FRAME_FETCH_RA)
 22730
 22731     ___RESULT = result;
 22732
 22733     ___SET_R1(___RESULT)
 22734  }
```

In general I think all of my segfault issues were resolved by disabling `___stack_overflow_undo_if_possible`.  Shouldn't that mean the problem is within `___stack_overflow_undo_if_possible`?  Maybe I should do a more thorough testing with it disabled and see if I can still reproduce these issues?

Negdayen:
Digging more into the previous message's session in GDB...  it looks like one thread is modifying the contents of another thread's processor's stack.

```
1       (rr) break _kernel.c:22722
2       Breakpoint 10 at 0x5612b0a57529: file _kernel.c, line 22723.
3       (rr) break _kernel.c:22731
4       Breakpoint 11 at 0x5612b0a5769f: file _kernel.c, line 22731.
5       (rr) c
6       Continuing.
7
8       Thread 4 hit Breakpoint 10, ___H___kernel (___ps=0x5612b19b6688 <___gstate0+17576>) at _kernel.c:22723
9       22723      ___FRAME_STORE_RA(___R0)
10      (rr) watch -location ___ps->fp[0]
11      Hardware watchpoint 12: -location ___ps->fp[0]
12      (rr) print ___fp
13      $43 = (long *) 0x72467f900ed0
14      (rr) print ___ps->fp
15      $44 = (long *) 0x72467f900ed0
16      (rr) set $fpa = $44
17      (rr) print/x ___ps->fp[0]
18      $45 = 0x5612b180a191
19      (rr) x 0x5612b180a191
20      0x5612b180a191 <___lbltbl+37201>:       0x00000000
21      (rr) c
22      Continuing.
23
24      Thread 4 hit Hardware watchpoint 12: -location ___ps->fp[0]
25
26      Old value = 94638287397265
27      New value = 94638288932625
28      ___H___kernel (___ps=0x5612b19b6688 <___gstate0+17576>) at _kernel.c:22724
29      22724      ___W_ALL
30      (rr) print/x ___ps->fp[0]
31      $46 = 0x5612b1980f11
32      (rr) x 0x5612b1980f11
33      0x5612b1980f11 <___lbltbl+21009>:       0x00000000
34      (rr) c
35      Continuing.
36      [Switching to Thread 1216.1216]
37
38      Thread 1 hit Hardware watchpoint 12: -location ___ps->fp[0]
39
40      Old value = 94638288932625
41      New value = -4
42      ___H___system (___ps=0x5612b19b21e0 <___gstate0>) at _system.c:12459
43      12459      ___SET_R0(___LBL(1))
44      (rr) macro expand ___SET_STK(4,___R3)
45      expands to: ___fp[-(4)]=(___r3);
46      (rr) print ___r3
47      $50 = -4
48      (rr) print &___fp[-4]
49      $51 = (long *) 0x72467f900ed0
50      (rr) print $fpa == &___fp[-4]
51      $54 = 1
```

Continuing the debugging session, the invalid address `0xffffffffffffffea` gets written into thread-4's processor's stack from thread-1 in ##kernel-handlers "break handler" (line 11714).  Thread-4 then resumes execution and segfaults.

```
11711              ___FP_ADJFP(fp,-___FRAME_SPACE(fs)); /* get base of frame */
11712
11713              for (i=fs; i>0; i--)
11714                ___SET_STK(i,___FP_STK(fp,i))
11715
11716              ra2 = ___STK(link+1);
```

feeley:
I don't have time to look at this now, but it is a very interesting lead to follow.  Be careful however in "watching" a specific memory location because after a GC the stack may have moved to a different location so the "write" to that location may be for a completely unrelated operation.  Just to be sure this is not happening it would be good to write a message when `___gc` is called (or `move_continuation`).  Another debugging trick is to `./configure --enable-debug --enable-debug-log --enable-debug-ctrl-flow-history` and to look at the last few operations of the two threads in `gambit.log`.  It will be interesting to see what the threads were doing just before the moment of the crash (or when one OS thread modifies another OS thread's stack).  You can force this information to be written to `gambit.log` by calling `___print_ctrl_flow_history()`.

Negdayen:
I don't think the stacks moved during the watchpoint, but there was a garbage collection call just prior to when the problem became observable.  Sorry if this information overload, but here is the full troubleshooting session; also attaching the `gambit.log`.  A little bit stumped of where I could go from here... is there a way I can introspect the label table for what a certain label might refer to?  

Loading up GDB commands to print when garbage collection events occur:

```
$ cat gc.gdb
break ___garbage_collect
commands
silent
printf "\n"
printf "BEG garbage collection: processor %d, os thread %d\n", ___INT(___CURRENTPROCESSORID), $_thread
continue
end

break on_all_processors
commands
silent
printf "\n"
printf "BEG on_all_processors:  processor %d, os thread %d\n", ___INT(___CURRENTPROCESSORID), $_thread
continue
end

break setup.c:1115
commands
silent
printf "\n"
printf "END on_all_processors:  processor %d, os thread %d\n", ___INT(___CURRENTPROCESSORID), $_thread
continue
end

break setup.c:1162
commands
silent
printf "\n"
printf "END garbage collection: processor %d, os thread %d\n", ___INT(___CURRENTPROCESSORID), $_thread
continue
end

break move_continuation
commands
silent
printf "\n"
printf "Move continuation:      processor %d, os thread %d\n", ___INT(___CURRENTPROCESSORID), $_thread
continue
end
```

```
$ rr replay -i=mi minimal-crash-0
...
GNU gdb (GDB) 12.1
...
(rr) source gc.gdb
Breakpoint 1 at 0x5612b09d9919: file setup.c, line 1153.
Breakpoint 2 at 0x5612b09d985d: file setup.c, line 1105.
Breakpoint 3 at 0x5612b09d9895: file setup.c, line 1116.
Breakpoint 4 at 0x5612b09d9947: file setup.c, line 1162.
Breakpoint 5 at 0x5612b09e5fd0: file mem.c, line 6077.
```

Continuing execution, GDB stops when it encounters SIGSEGV.

```
(rr) continue
Continuing.
(rr) 
BEG on_all_processors:  processor 0, os thread 1
[New Thread 1216.1217]
[New Thread 1216.1218]
(rr) 
END on_all_processors:  processor 0, os thread 1
(parallelism-level: 3)
(rr) 
BEG garbage collection: processor 0, os thread 1
(rr) 
BEG on_all_processors:  processor 0, os thread 1
[Switching to Thread 1216.1218]
(rr) 
Move continuation:      processor 2, os thread 3
[Switching to Thread 1216.1217]
(rr) 
Move continuation:      processor 1, os thread 2
[Switching to Thread 1216.1216]
(rr) 
Move continuation:      processor 0, os thread 1
[Switching to Thread 1216.1217]
(rr) 
BEG garbage collection: processor 1, os thread 2
(rr) 
BEG on_all_processors:  processor 1, os thread 2
[Switching to Thread 1216.1216]
(rr) 
END on_all_processors:  processor 0, os thread 1
(rr) 
END garbage collection: processor 0, os thread 1
(rr) 
BEG garbage collection: processor 0, os thread 1
(rr) 
BEG on_all_processors:  processor 0, os thread 1
[Switching to Thread 1216.1218]
(rr) 
Move continuation:      processor 2, os thread 3
[Switching to Thread 1216.1217]
(rr) 
Move continuation:      processor 1, os thread 2
[Switching to Thread 1216.1216]
(rr) 
Move continuation:      processor 0, os thread 1
[Switching to Thread 1216.1217]
(rr) 
END on_all_processors:  processor 1, os thread 2
(rr) 
END garbage collection: processor 1, os thread 2
[Switching to Thread 1216.1216]
(rr) 
END on_all_processors:  processor 0, os thread 1
(rr) 
END garbage collection: processor 0, os thread 1
(rr) 
BEG garbage collection: processor 0, os thread 1
(rr) 
BEG on_all_processors:  processor 0, os thread 1
[Switching to Thread 1216.1218]
(rr) 
Move continuation:      processor 2, os thread 3
[Switching to Thread 1216.1217]
(rr) 
Move continuation:      processor 1, os thread 2
[Switching to Thread 1216.1216]
(rr) 
Move continuation:      processor 0, os thread 1
(rr) 
END on_all_processors:  processor 0, os thread 1
(rr) 
END garbage collection: processor 0, os thread 1
(#t #t #t)

Thread 2 received signal SIGSEGV, Segmentation fault.
[Switching to Thread 1216.1217]
0x00005612b0a88142 in ___H___kernel (___ps=0x5612b19b6688 <___gstate0+17576>) at _kernel.c:29569
29569	___END_M_SW
```

Setting up breakpoints for the points of interest in `##os-condvar-select!`.

```
$ cat crit.gdb
define print_fp_state
printf "\n"
printf "___ps->fp = %x\n", ___ps->fp
printf "___fp     = %x\n", ___fp
printf "___fp[0]  = %x\n", ___fp[0]
end

break _kernel.c:22722
commands
print_fp_state
end

break _kernel.c:22724
commands
print_fp_state
end

break _kernel.c:22726
commands
print_fp_state
end

break _kernel.c:22728
commands
print_fp_state
end

break _kernel.c:22729
commands
print_fp_state
end

break _kernel.c:22737
```

```
(rr) source crit.gdb
Breakpoint 6 at 0x5612b0a57529: file _kernel.c, line 22723.
Breakpoint 7 at 0x5612b0a5754e: file _kernel.c, line 22724.
Breakpoint 8 at 0x5612b0a575cc: file _kernel.c, line 22726.
Breakpoint 9 at 0x5612b0a575ec: file _kernel.c, line 22728.
Breakpoint 10 at 0x5612b0a5766a: file _kernel.c, line 22729.
Breakpoint 11 at 0x5612b0a576bb: file _kernel.c, line 22737.
```

Also setting up breakpoints for every line of `execute_sync_op_loop` for fine-grained observation of garbage collection (omitting uninteresting breakpoint definitions).

```
(rr) source gc-fine-grained.gdb
```

Reversing execution, `___R0` is set to an invalid address `0xffffffffffffffea`.

```
(rr) reverse-continue
Continuing.

Thread 2 hit Breakpoint 11, ___H___kernel (___ps=0x5612b19b6688 <___gstate0+17576>) at _kernel.c:22737
22737	   ___JUMPPRM(___NOTHING,___R0)
print/x ___R0
$1 = 0xffffffffffffffea
(rr) x ___R0
0xffffffffffffffea:	Cannot access memory at address 0xffffffffffffffea
```

`___R0` took on this value from `___fp[-(-0)]` after the call to `___SET_R0(___FRAME_FETCH_RA)` .

```
(rr) reverse-continue
Continuing.

Thread 2 hit Breakpoint 10, ___H___kernel (___ps=0x5612b19b6688 <___gstate0+17576>) at _kernel.c:22729
22729	   ___SET_R0(___FRAME_FETCH_RA)
(rr) 
___ps->fp = 7f900ed0
___fp     = 7f900ed0
___fp[0]  = ffffffea
(rr) print/x ___FRAME_FETCH_RA
$2 = 0xffffffffffffffea
(rr) macro expand ___SET_R0(___FRAME_FETCH_RA)
expands to: ___r0=((___fp != ___ps->stack_break) ? ___fp[-(-0)] : ___ps->handler_break);
(rr) print (___fp != ___ps->stack_break)
$3 = 1
(rr) print ___fp[0] == ___ABSENT
$4 = 1
```

Going back a bit more to a point where `___fp[0]` on thread 2 did not have this value, a few garbage collection points are hit, but I think they don't have impact on the frame pointer.  Looks like thread-1 and thread-3 were finishing up the very tail end of a garbage collection call (origins of which trace back to a `##check-heap` call I think).

```
(rr) reverse-continue
Continuing.

Thread 2 hit Breakpoint 9, ___H___kernel (___ps=0x5612b19b6688 <___gstate0+17576>) at _kernel.c:22728
22728	   ___R_ALL
(rr) 
___ps->fp = 7f900ed0
___fp     = 7f900ed0
___fp[0]  = ffffffea

Continuing.
[Switching to Thread 1216.1218]

Thread 3 hit Breakpoint 8, ___H___kernel (___ps=0x5612b19bab30 <___gstate0+35152>) at _kernel.c:22726
22726	   result = ___os_condvar_select (___ARG1, ___ARG2);
(rr) 
___ps->fp = 7fa0ff70
___fp     = 7fa0ff70
___fp[0]  = b1980f11

Continuing.

Thread 3 hit Breakpoint 7, ___H___kernel (___ps=0x5612b19bab30 <___gstate0+35152>) at _kernel.c:22724
22724	   ___W_ALL
(rr) 
___ps->fp = 7fa0ff70
___fp     = 7fa0ff70
___fp[0]  = b1980f11

Continuing.

Thread 3 hit Breakpoint 6, ___H___kernel (___ps=0x5612b19bab30 <___gstate0+35152>) at _kernel.c:22723
22723	   ___FRAME_STORE_RA(___R0)
(rr) 
___ps->fp = 7fa0ff70
___fp     = 7fa0ff70
___fp[0]  = b180a191

Continuing.

Thread 3 hit Breakpoint 45, execute_sync_op_loop (___ps=0x5612b19bab30 <___gstate0+35152>, sop_ptr=0x72c020f42d90, first_iter=0) at setup.c:1076
1076	  if (id == 0 && ___err != ___FIX(___NO_ERR))
(rr) 
Continuing.

Thread 3 hit Breakpoint 17, execute_sync_op_loop (___ps=0x5612b19bab30 <___gstate0+35152>, sop_ptr=0x72c020f42d90, first_iter=0) at setup.c:1052
1052	          if (!first_iter)
(rr) 
Continuing.

Thread 3 hit Breakpoint 14, execute_sync_op_loop (___ps=0x5612b19bab30 <___gstate0+35152>, sop_ptr=0x72c020f42d90, first_iter=0) at setup.c:1045
1045	      if (sop.op == OP_NOOP)
(rr) 
Continuing.
[Switching to Thread 1216.1216]
(rr) 
END garbage collection: processor 0, os thread 1
(rr) 
END on_all_processors:  processor 0, os thread 1

Thread 1 hit Breakpoint 45, execute_sync_op_loop (___ps=0x5612b19b21e0 <___gstate0>, sop_ptr=0x7ffee6635d10, first_iter=0) at setup.c:1076
1076	  if (id == 0 && ___err != ___FIX(___NO_ERR))
(rr) 
Continuing.

Thread 1 hit Breakpoint 17, execute_sync_op_loop (___ps=0x5612b19b21e0 <___gstate0>, sop_ptr=0x7ffee6635d10, first_iter=0) at setup.c:1052
1052	          if (!first_iter)
(rr) 
Continuing.

Thread 1 hit Breakpoint 14, execute_sync_op_loop (___ps=0x5612b19b21e0 <___gstate0>, sop_ptr=0x7ffee6635d10, first_iter=0) at setup.c:1045
1045	      if (sop.op == OP_NOOP)
(rr) 
Continuing.
[Switching to Thread 1216.1217]

Thread 2 hit Breakpoint 8, ___H___kernel (___ps=0x5612b19b6688 <___gstate0+17576>) at _kernel.c:22726
22726	   result = ___os_condvar_select (___ARG1, ___ARG2);
(rr) 
___ps->fp = 7f900ed0
___fp     = 7f900ed0
___fp[0]  = b1980f11

Continuing.

Thread 2 hit Breakpoint 7, ___H___kernel (___ps=0x5612b19b6688 <___gstate0+17576>) at _kernel.c:22724
22724	   ___W_ALL
(rr) 
___ps->fp = 7f900ed0
___fp     = 7f900ed0
___fp[0]  = b1980f11

Continuing.

Thread 2 hit Breakpoint 6, ___H___kernel (___ps=0x5612b19b6688 <___gstate0+17576>) at _kernel.c:22723
22723	   ___FRAME_STORE_RA(___R0)
(rr) 
___ps->fp = 7f900ed0
___fp     = 7f900ed0
___fp[0]  = b180a191
```

At this point I place a watchpoint on the location of `___ps->fp[0]` and continue execution forward again to see what happens... it modified several times by thread-1 before thread-2 resumes execution.  Thread-1 appears to be performing `pp` related tasks the first time it modifies thread-2's `___ps->fp[0]`; it is in `##table-ref`, and backtracing leads me to procedures involving marktables, and also `(#t #t #t)` is output while continuing through the code.

```
(rr) watch -location ___ps->fp[0]
Hardware watchpoint 49: -location ___ps->fp[0]
(rr) continue
Continuing.

Thread 2 hit Breakpoint 7, ___H___kernel (___ps=0x5612b19b6688 <___gstate0+17576>) at _kernel.c:22724
22724	   ___W_ALL
(rr) 
___ps->fp = 7f900ed0
___fp     = 7f900ed0
___fp[0]  = b1980f11

Continuing.

Thread 2 hit Breakpoint 8, ___H___kernel (___ps=0x5612b19b6688 <___gstate0+17576>) at _kernel.c:22726
22726	   result = ___os_condvar_select (___ARG1, ___ARG2);
(rr) 
___ps->fp = 7f900ed0
___fp     = 7f900ed0
___fp[0]  = b1980f11

Continuing.
[Switching to Thread 1216.1216]

Thread 1 hit Breakpoint 14, execute_sync_op_loop (___ps=0x5612b19b21e0 <___gstate0>, sop_ptr=0x7ffee6635d10, first_iter=0) at setup.c:1045
1045	      if (sop.op == OP_NOOP)
(rr) 
Continuing.

Thread 1 hit Breakpoint 17, execute_sync_op_loop (___ps=0x5612b19b21e0 <___gstate0>, sop_ptr=0x7ffee6635d10, first_iter=0) at setup.c:1052
1052	          if (!first_iter)
(rr) 
Continuing.

Thread 1 hit Breakpoint 45, execute_sync_op_loop (___ps=0x5612b19b21e0 <___gstate0>, sop_ptr=0x7ffee6635d10, first_iter=0) at setup.c:1076
1076	  if (id == 0 && ___err != ___FIX(___NO_ERR))
(rr) 
Continuing.
(rr) 
END on_all_processors:  processor 0, os thread 1
(rr) 
END garbage collection: processor 0, os thread 1

Thread 1 hit Hardware watchpoint 49: -location ___ps->fp[0]

Old value = 94638288932625
New value = -4
___H___system (___ps=0x5612b19b21e0 <___gstate0>) at _system.c:12459
12459	   ___SET_R0(___LBL(1))
(rr) 
Continuing.

Thread 1 hit Hardware watchpoint 49: -location ___ps->fp[0]

Old value = -4
New value = -8
___H___system (___ps=0x5612b19b21e0 <___gstate0>) at _system.c:12689
12689	   ___SET_R0(___LBL(1))
(rr) 
Continuing.

Thread 1 hit Hardware watchpoint 49: -location ___ps->fp[0]

Old value = -8
New value = -6
___H___io (___ps=0x5612b19b21e0 <___gstate0>) at _io.c:36348
36348	   ___SET_R0(___LBL(1))
(rr) 
Continuing.

Thread 1 hit Hardware watchpoint 49: -location ___ps->fp[0]

Old value = -6
New value = -2
___H___io (___ps=0x5612b19b21e0 <___gstate0>) at _io.c:36348
36348	   ___SET_R0(___LBL(1))
(rr) 
Continuing.

Thread 1 hit Hardware watchpoint 49: -location ___ps->fp[0]

Old value = -2
New value = -4
___H___system (___ps=0x5612b19b21e0 <___gstate0>) at _system.c:12459
12459	   ___SET_R0(___LBL(1))
(rr) 
Continuing.

Thread 1 hit Hardware watchpoint 49: -location ___ps->fp[0]

Old value = -4
New value = -12
___H___system (___ps=0x5612b19b21e0 <___gstate0>) at _system.c:12689
12689	   ___SET_R0(___LBL(1))
(rr) 
Continuing.

Thread 1 hit Hardware watchpoint 49: -location ___ps->fp[0]

Old value = -12
New value = -2
___H___io (___ps=0x5612b19b21e0 <___gstate0>) at _io.c:71367
71367	   ___SET_STK(3,___R2)
(rr) 
(#Continuing.
t #t #t)
Thread 1 hit Hardware watchpoint 49: -location ___ps->fp[0]

Old value = -2
New value = 4
___H___kernel (___ps=0x5612b19b21e0 <___gstate0>) at _kernel.c:27375
27375	   ___SET_STK(4,___R0)
(rr) 
Continuing.


Thread 1 hit Hardware watchpoint 49: -location ___ps->fp[0]

Old value = 4
New value = 0
___H___thread (___ps=0x5612b19b21e0 <___gstate0>) at _thread.c:24314
24314	   ___SET_STK(5,___VECTORREF(___R4,___FIX(1L)))
(rr) 
Continuing.

Thread 1 hit Hardware watchpoint 49: -location ___ps->fp[0]

Old value = 0
New value = -22
___H___kernel (___ps=0x5612b19b21e0 <___gstate0>) at _kernel.c:11713
11713	           for (i=fs; i>0; i--)
(rr) 
Continuing.
[Switching to Thread 1216.1218]

Thread 3 hit Breakpoint 14, execute_sync_op_loop (___ps=0x5612b19bab30 <___gstate0+35152>, sop_ptr=0x72c020f42d90, first_iter=0) at setup.c:1045
1045	      if (sop.op == OP_NOOP)
(rr) 
Continuing.

Thread 3 hit Breakpoint 17, execute_sync_op_loop (___ps=0x5612b19bab30 <___gstate0+35152>, sop_ptr=0x72c020f42d90, first_iter=0) at setup.c:1052
1052	          if (!first_iter)
(rr) 
Continuing.

Thread 3 hit Breakpoint 45, execute_sync_op_loop (___ps=0x5612b19bab30 <___gstate0+35152>, sop_ptr=0x72c020f42d90, first_iter=0) at setup.c:1076
1076	  if (id == 0 && ___err != ___FIX(___NO_ERR))
(rr) 
Continuing.

Thread 3 hit Breakpoint 6, ___H___kernel (___ps=0x5612b19bab30 <___gstate0+35152>) at _kernel.c:22723
22723	   ___FRAME_STORE_RA(___R0)
(rr) 
___ps->fp = 7fa0ff70
___fp     = 7fa0ff70
___fp[0]  = b180a191

Continuing.

Thread 3 hit Breakpoint 7, ___H___kernel (___ps=0x5612b19bab30 <___gstate0+35152>) at _kernel.c:22724
22724	   ___W_ALL
(rr) 
___ps->fp = 7fa0ff70
___fp     = 7fa0ff70
___fp[0]  = b1980f11

Continuing.

Thread 3 hit Breakpoint 8, ___H___kernel (___ps=0x5612b19bab30 <___gstate0+35152>) at _kernel.c:22726
22726	   result = ___os_condvar_select (___ARG1, ___ARG2);
(rr) 
___ps->fp = 7fa0ff70
___fp     = 7fa0ff70
___fp[0]  = b1980f11

Continuing.
[Switching to Thread 1216.1217]

Thread 2 hit Breakpoint 9, ___H___kernel (___ps=0x5612b19b6688 <___gstate0+17576>) at _kernel.c:22728
22728	   ___R_ALL
(rr) 
___ps->fp = 7f900ed0
___fp     = 7f900ed0
___fp[0]  = ffffffea

Continuing.

Thread 2 hit Breakpoint 10, ___H___kernel (___ps=0x5612b19b6688 <___gstate0+17576>) at _kernel.c:22729
22729	   ___SET_R0(___FRAME_FETCH_RA)
(rr) 
___ps->fp = 7f900ed0
___fp     = 7f900ed0
___fp[0]  = ffffffea
```

[annotated-gambit.log](https://github.com/gambit/gambit/files/9742222/annotated-gambit.log)

PR #784, opened 2022-09-22
Fix http module

FredericHamel:
- The unit test of the `_http` module was broken. The parameters to `##string-prefix?` were passed in the wrong order.
- Added two tests for invalid URL
- Add URI validation to avoid `http-get` crashing on invalid URL.


feeley:
Thanks for the PR!

PR #787, opened 2022-10-01
Add demo directories for SRFI 231

gambiteer:


feeley:
I have added the file `lib/srfi/0/demo/demo.scm` to be used as a template for other demos.  It can be run with `gsi srfi/0/demo`.  The source code should end with the line

```
(print "Demo source code:\n\n" (read-file-string (this-source-file)))
```

so that the demo's source code is automatically printed when executing the demo, and the source code should be kept to a minimum (~ 20 lines of code so it can fit in a 80x24 terminal).  If there is a need for demos with more lines of code the main demo should explain how to run other demos (e.g. in a comment like`;; Execute "gsi srfi/1234/demoN" with 2<=N<=5 for other demos.`).  The goal of a module's demo is to get a programmer using the module with low effort with a copy-paste of the demo's source code, so the demo should be minimal but still do something interesting and easy to understand.

PR #788, opened 2022-10-02
Fix ##raise-high-level-interrupt call in ##thread-intr-propagate!

Negdayen:


feeley:
Nice of you to catch this typo!

PR #791, opened 2022-10-17
Add tests and demo for SRFI 179

gambiteer:
Marc:  What do you think of this?


gambiteer:
I presume i you don't like some aspect of this you'll tell me.

PR #827, opened 2023-06-16
warn user when symbol is undefined

Terramorpha:
This commit adds in `_env.scm` a new field to the env to accumulates a binding between each symbol and it's origin namespace.
It also adds a phase to warn the user about variables that are probably undefined: those used, but not "declared" in a `(##namespace ("ns#" <var>))` form.

### DIFF
--- a/gsc/_env.scm
+++ b/gsc/_env.scm
@@ -61,11 +61,18 @@
       (lambda () (env-frame #f '())))
 
 (define (env-frame env vars)
-  (vector (cons vars #f)     ; cell containing variables in this frame
-          '()                           ; macro definitions
-          (if env (env-decl-ref env) '()) ; declarations
-          '()                           ; namespace
-          env))                         ; parent env
+  (vector
+   ;; cell containing variables in this frame and an association between each
+   ;; symbol and it's namespace
+   (cons vars (cons '() #f))
+   ;; macro definitions
+   '()
+   ;; declarations
+   (if env (env-decl-ref env) '())
+   ;; namespace
+   '()
+   ;; parent env
+   env))
 
 (define (env-new-var! env name source)
   (let* ((glob (not (env-parent-ref env)))
@@ -106,12 +113,14 @@
           namespace
           (env-parent-ref env)))
 
-(define (env-vars-ref env)       (car (vector-ref env 0)))
-(define (env-vars-set! env vars) (set-car! (vector-ref env 0) vars))
-(define (env-macros-ref env)     (vector-ref env 1))
-(define (env-decl-ref env)       (vector-ref env 2))
-(define (env-namespace-ref env)  (vector-ref env 3))
-(define (env-parent-ref env)     (vector-ref env 4))
+(define (env-var-origins-ref env)       (cadr (vector-ref env 0)))
+(define (env-var-origins-set! env ors)  (set-car! (cdr (vector-ref env 0)) ors))
+(define (env-vars-ref env)              (car (vector-ref env 0)))
+(define (env-vars-set! env vars)        (set-car! (vector-ref env 0) vars))
+(define (env-macros-ref env)            (vector-ref env 1))
+(define (env-decl-ref env)              (vector-ref env 2))
+(define (env-namespace-ref env)         (vector-ref env 3))
+(define (env-parent-ref env)            (vector-ref env 4))
 
 (define (env-namespace-lookup env name)
   (let loop ((lst (env-namespace-ref env)))
@@ -127,14 +136,42 @@
                     (loop (cdr lst))))))
         #f)))
 
+(define (env-namespace-lookup-both env name)
+  (let loop ((lst (env-namespace-ref env)))
+    (if (pair? lst)
+        (let* ((x (car lst))
+               (space (car x))
+               (aliases (cdr x)))
+          (if (null? aliases)
+              (cons (make-full-name space name) #f)
+              (let ((a (assq name aliases)))
+                (if a
+                    (cons (make-full-name space (cdr a))
+                          x)
+                    (loop (cdr lst))))))
+        #f)))
+
+
 (define (env-lookup env name stop-at-first-frame? proc)
+
+  (define (assoc-set! al k v)
+    (let ((cell (assoc k al)))
+      (if cell
+          (begin
+            (set-cdr! cell v)
+            al)
+          (cons (cons k v) al))))
 
   (define (search env name full?)
     (if full?
         (search* env name #t)
-        (let ((full-name (env-namespace-lookup env name)))
-          (if full-name
-              (search* env full-name #t)
+        (let ((full-name-and-ns (env-namespace-lookup-both env name)))
+          (if full-name-and-ns
+              (begin
+                (env-var-origins-set! env (assoc-set! (env-var-origins-ref env)
+                                                      (car full-name-and-ns)
+                                                      (cdr full-name-and-ns)))
+                (search* env (car full-name-and-ns) #t))
               (search* env name #f)))))
 
   (define (search* env name full?)
--- a/gsc/_env.scm
+++ b/gsc/_env.scm
@@ -64,7 +64,7 @@
   (vector
    ;; cell containing variables in this frame and an association between each
    ;; symbol and it's namespace
-   (cons vars (cons '() #f))
+   (cons vars #f)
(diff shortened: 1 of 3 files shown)


feeley:
Thank you for your first contribution to Gambit! This is a much needed feature. There are a few things to change for better integration with the rest of the system.

The warning message should be in the same style as the one that is displayed by the linker for the same issue (an undefined variable). Currently this is what is displayed:

```
$ cat foo.scm
(##namespace ("foo#"))
(##namespace ("bar#" a b))
(##namespace ("" define +))
(define (f x) (+ x a c))
$ gsc -warnings -link foo.scm
*** WARNING IN "foo.scm"@4.22 -- missing definition for `foo#c`
*** WARNING -- "bar#a" is not defined,
***            referenced in: ("foo.c")
*** WARNING -- "foo#c" is not defined,
***            referenced in: ("foo.c")
```

The message should say:

```
*** WARNING IN "foo.scm"@4.22 -- "foo#c" is not defined
```

The use of an association list is not ideal due to the linear lookup time. It is best to use tables. For portability, the file `_host.scm` defines the table data type and operations using association lists, but when bootstrapped with Gambit the tables operations are fast. Moreover, only the global environment needs to maintain an "origins", so the `make-global-environment` could be defined with an extra field like this:

```
(define make-global-environment #f)
(set! make-global-environment
      (lambda ()
        (vector (cons '() #f) ; cell containing variables in this frame
                '() ; macro definitions
                '() ; declarations
                '() ; namespace
                #f  ; no parent env
                (make-table)))) ; origins
```

By the way, I don't particularly like the term `origins`. I think `externals` may be more appropriate (names that come from outside the code unit).

The following code:

```
(let ((full-name-and-ns (env-namespace-lookup-both env name)))
   (if full-name-and-ns
       (begin
         (env-var-origins-set! env (assoc-set! (env-var-origins-ref env)
                                               (car full-name-and-ns)
                                               (cdr full-name-and-ns)))
         (search* env (car full-name-and-ns) #t))
```

is clearer when written like this:

```
(let ((full-name-and-ns (env-namespace-lookup-both env name)))
  (if full-name-and-ns
      (let ((full-name (car full-name-and-ns))
            (ns (cdr full-name-and-ns)))
        (env-var-origins-set!
         env
         (assoc-set! (env-var-origins-ref env) full-name ns))
        (search* env full-name #t))
```

Of course, `assoc-set!` should be replaced by code that uses `table-set!`.

There are a few uses of `filter` in the code that should be replaced by calls to the (portable) `keep` procedure defined in `_utils.scm`.

gsc/_front.scm
```diff
@@ -294,9 +292,10 @@
                          (compiler-user-warning
                           (source-locat (node-source ref))
                           (string-append
-                           "missing definition for `"
+                           "\""
```
feeley, line 295: This could be written better to avoid writing a name that does not match the Scheme string syntax. This is also done in other places in the gsc source code, so this can wait until all places are changed together.

For example:

`(object->string (symbol->string '|abc\"def|))` => `"\"abc\\\"def\""`

gsc/_env.scm
```diff
@@ -153,24 +159,17 @@
...
   (define (search env name full?)
     (if full?
         (search* env name #t)
         (let ((full-name-and-ns (env-namespace-lookup-both env name)))
           (if full-name-and-ns
-              (begin
-                (env-var-origins-set! env (assoc-set! (env-var-origins-ref env)
-                                                      (car full-name-and-ns)
-                                                      (cdr full-name-and-ns)))
+              (let ((full-name (car full-name-and-ns))
```
feeley, line 167: For better error messages, it would seem better to have a table indexed by the namespace that contains for each namespace a set of full-names of the global referred to in that namespace (which could be represented with a table). That way all the undefined names can be grouped according to the namespace, which will make the error messages easier to understand.

feeley (changes requested):

feeley:
Because of how `env-frame` is now defined, `make-global-environment` can be defined as before:

```
(set! make-global-environment
   (lambda () (env-frame #f '())))
```

In `env-namespace-lookup-both` the change from `(cons (make-full-name space name) #f)` to `(cons (make-full-name space name) x)` is an error because it is not possible to distinguish universal namespace declarations from specific namespace declarations. Here's an example:

```
(display "hello!\n")  ;; not in the scope of a universal namespace declaration
(namespace ("foo#") ("bar#" a b) ("" define let +))
(define (f)
  (+ a  ;; is mapped to bar#a by bar# namespace declaration
     b  ;; is mapped to bar#b by bar# namespace declaration
     c  ;; is mapped to foo#c by foo# namespace declaration
     ##other            ;; fully qualified
     fully#qualified))  ;; fully qualified
(f 10)
```

When compiled this gives the warnings that `foo#c`, `##other`, `fully#qualified`, and `display` are not defined. Only `foo#c` should be indicated as undefined because it was resolved using the universal namespace declaration `(namespace ("foo#"))`. The fully qualified names and names like `display` that weren't resolved using a namespace, should not be flagged as not defined.

Concerning `lset-minus`, in other parts of the gsc source code, set difference is called `difference`, not `minus`. So this should be renamed `lset-difference`. The definition of that procedure should be moved to `_utils.scm` where other general purpose set operations are defined.

Terramorpha:
> Because of how `env-frame` is now defined, `make-global-environment` can be defined as before:
> 
> ```
> (set! make-global-environment
>    (lambda () (env-frame #f '())))
> ```
> 
> In `env-namespace-lookup-both` the change from `(cons (make-full-name space name) #f)` to `(cons (make-full-name space name) x)` is an error because it is not possible to distinguish universal namespace declarations from specific namespace declarations.

I don't understand how it is not possible. If a variable, say, `a` is associated to the `(namespace ("foo#"))` declaration, then that namespace can be distinguished from a specific declaration by asking `(null? (cdr ns))`

feeley:
Perhaps I don't understand what the `cdr` of the pair is used for.

Terramorpha:
`env-namespace-lookup-both` returns a pair containing the fully qualified name *and* its namesapce.

```scheme
(##namespace
  ("foo#")
  ("" +))

(a)
(+ 1 2)
```

Here, `env-namespace-lookup-both` would return `(foo#a . ("foo#))` for `a` and `(+ . ("" +))` for `+`

feeley:
OK now I see. You can tell from the `cdr` of `x`.

feeley:
Thanks for the PR which is now merged!

Now that undefined references are detected when the `-warnings` is used, an interesting task is to check if the Gambit runtime system and modules contain undefined references. The `standard` macro must first be changed to avoid introducing what will be detected as undefined references:

```
$ git diff
diff --git a/lib/_gambit#.scm b/lib/_gambit#.scm
index b65eedf9..d604bc41 100644
--- a/lib/_gambit#.scm
+++ b/lib/_gambit#.scm
@@ -173,7 +173,7 @@
        (##source-strip (car (##source-strip src)))))
 
     (define (proc sym)
-      `(##let () (##namespace ("")) ,sym))
+      `(##let () (##namespace ("" ,sym)) ,sym))
 
     (##deconstruct-call
      src
```

The call `(standard car)` is supposed to expand into a reference to `car` in the empty namespace. It is useful in the context where the code is in the scope of a universal namespace declaration and just saying "`car`" would resolve to something in a namespace different from the empty one.

Then Gambit should be recompiled with:

```
make bootstrap
make bootclean
./configure --enable-single-host COMPILATION_OPTIONS=-warnings
make
```

Then the modules should be recompiled with `make clean-modules; make modules` after changing line 227 of `lib/module-common.mk` like this:

```
$(rootfromhere)/gsc/gsc -:~~bin=$(srcdirpfx)$(rootfromhere)/bin,~~lib=$(srcdirpfx)$(rootfromhere)/lib,~~include=$(srcdirpfx)$(rootfromhere)/include -f -e "(let ((x (##build-module \"$$mod_dir$$main_mod\" (quote $$target) (quote ((module-ref $$modref) (warnings)))))) (if (not x) (begin (println \"*** module failed to build\") (exit 1))))"; \
```

so that the `warnings` option is used to build the modules.

I'm sure this will detect undefined references that are not errors (false positives), but possibly a few that are actual errors. Fixing the true positives will be the next step. We should also find a way to reduce or eliminate all the false positives (perhaps by adding appropriate `namespace` declarations, maybe in `foobar#.scm` files).

PR #843, opened 2023-07-28
Fixing the Dockerfile

SamuelYvon:
This commit fixes it as well as adds a CI step to build the docker
image. The step can be configured to publish the docker image to
public container repositories at your discretion so people can try
out gambit by just pulling down the container.


Dockerfile
```diff
@@ -1,6 +1,22 @@
...
+
 WORKDIR gambit_install
+
 COPY . .
-RUN make clean && ./configure --enable-single-host && make -j$(nproc) && make check && make modules && make doc && make install
+
+RUN ./configure $ENABLE_SINGLE_HOST $ENABLE_MARCH $ENABLE_DCLIB
+RUN make -j${NPROC}
+RUN make check
+RUN make modules
```
feeley, line 15: The `make modules` is not needed as it is done during the `make -j${NPROC}`

feeley (changes requested):

SamuelYvon:
@feeley ping ;)

PR #844, opened 2023-07-30
Allow OPENSSL_DIR to be overridable

sevan:
Allows build with OpenSSL support when modern OpenSSL is installed in a different path.


sevan:
ugh, using Makefile syntax for what will be a shell script, only allows it to be overridable by accident.
Will redo when I'm not tired.

PR #861, opened 2023-10-20
Update .gitattributes

lassik:
GitHub Linguist wrongly classified psyntax*.pp as Pascal source files.


lassik:
NB: Since these are the pre-expanded copies of psyntax (`psyntax*.ss` being the non-expanded source files) the `generated` tag is probably fine.

feeley:
These files are no longer needed so they will eventually be removed.

PR #867, opened 2023-11-28
Make the "duplicate pattern variable" message more informative

wolneykien:
Having the `sym` and `pattern-vars` info makes it possible to guess which expression the error concerns.


gambiteer:
I like this proposal. I have some questions.
Do you have a test case to illustrate the change?  I'm not asking that it be added to the commit, just something to see what happens.
How is it related to SRFI 42?  Is this where you first saw a problem?
Thanks.  Brad

PR #883, opened 2024-01-28
Add vector-any, vector-every, string-any, string-every

gambiteer:
I modeled the new code on the code for `vector-for-each`.

### DIFF
--- a/doc/gambit.txi
+++ b/doc/gambit.txi
@@ -4066,6 +4066,51 @@ For example:
 
 @end deffn
 
+@deffn procedure vector-any @var{pred} @var{vec1} @var{vec2} @dots{}
+
+The procedure @code{vector-any} returns the first non-false value that
+the predicate @var{pred} returns when applied to the elements of the vectors
+@var{vec1} @var{vec2} @dots{} in order, beginning with index 0 and
+stopping at the length of the shortest
+vector argument. It returns @code{#f} if there is no such non-false value.
+
+For example:
+
+@smallexample
+> @b{(vector-any = '#(1 2) '#(0 2 4))}
+#t
+> @b{(vector-any number? '#())}
+#f
+> @b{(vector-any (lambda (a b) (and a b)) '#(#f 1 #f 2) '#(1 #f 2 3 5))}
+3
+@end smallexample
+
+@end deffn
+
+@deffn procedure vector-every @var{pred} @var{vec1} @var{vec2} @dots{}
+
+The procedure @code{vector-every} applies
+the predicate @var{pred} to the elements of the vectors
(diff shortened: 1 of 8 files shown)


lassik:
As a user: thank you. This is the kind of unexciting but very important work that makes people start a given Scheme implementation by default.

gambiteer:
Marc:  I just studied what you did for `any` and `every` in `lib/gambit/list/list.scm`; you recently added lines like
```
            (or (macro-auto-force (apply pred args))
                (any-n* rests next-rests))
```
It's not clear to me why you added `macro-auto-force`, but should I add lines like this to `vector-any` and `vector-every`?

feeley:
"auto-forcing" is enabled by a Gambit configuration option. It is intended to make `(force arg)` implicit on all strict operations. For example, `(+ (delay 1) (delay 2))` is equivalent to `(+ (force (delay 1)) (force (delay 2)))`. In the code you quote, the result of the `pred` procedure needs to be tested by the `or`, so the `or` operation is also a strict operation that needs to force the result of the `pred` procedure. Otherwise, if the `pred` procedure is `(lambda (x) (delay (< x 0)))` then the `or` would always think the result is non-false.

So yes this needs to be done for `vector-any` and `vector-every?`.

When Gambit implements `(future <expr>)` with promises and threads, it will allow the elegant expression of parallelism with `(+ (future E1) E2)`.

gambiteer:
I'm looking at this now.  In `_std#.scm` the arguments to `proc` in both `vect-map` and `vect-for-each` are not wrapped in `macro-auto-force`.  Those should be wrapped, too, right?

feeley:
The result of the procedure should not be auto-forced, neither for `vector-for-each` or `vector-map`, because the vector constructor is not strict. It is a different situation for `u8vector-map` because the `u8vector` constructor is strict.

gambiteer:
(a) I guess I don't know what "strict" means.
(b) I can't find `u8vector-map` anywhere in the sources.

feeley:
A function is "strict" on argument x if it needs to know the value of x to compute the result of the function. For example `(sqrt x)` is strict. A function is not strict on an argument x if it does not need the value of x to compute the result, or it only needs to know the identity of the value. For example `(list x)` is not strict. The point is that if x is represented as a placeholder for the result of some other parallel computation, then you can just put a reference to that placeholder in the list created by `(list x)`, i.e. you don't need to wait for the other computation to be done before creating the list.

For `u8vector-map` I was just making an example. If it existed it would need to auto force the result of the function because there's no way to put a reference to a placeholder in a byte (but that's just implementation... it could be designed/implemented differently, for example if u8vectors had a full word of memory for every element of the u8vector).

Issue #884, opened 2024-01-30
Incorrect handling of mismatched vector argument lengths in vector-map, etc.


gambiteer:
`max-arg` never changes  in `lib/_std#.scm` vector routines; should it be incremented each iteration of the loop?

On the other hand, calls to `##raise-length-mismatch-exception` in `lib/_std#.scm` don't seem to match its signature in `lib/_std.scm`.

On the third hand, error handling when `##allow-length-mismatch?` is `#f`seems inconsistent:
```
> (set! ##allow-length-mismatch? #f)
> (vector-map + '#(1) '#(1 2))
*** ERROR IN (stdin)@2.1-2.29 -- (Argument 3) Length does not match other arguments
(vector-map '#<procedure #2 +> #(1) #(1 2))
1> 
> (vector-map + '#(1) '#(1) '#(1 2))
*** ERROR IN (stdin)@3.1-3.35 -- (Argument 4) Length does not match other arguments
(vector-map '#<procedure #2 +> #(1) #(1) #(1 2))
1> 
> (vector-map + '#(1) '#(1) '#(1 2) '#(1))
#(4)   ;; <<<<<<?????????????????????
```
Anyway, I'm going to work on `vector-fold` and `vector-fold-right` because things seem to work when `##allow-length-mismatch?` is `#t`.

PR #892, opened 2024-02-15
fix argument length mismatch checking in vector-map and vector-for-each

afleckenstein:
This regards https://github.com/gambit/gambit/issues/884

I believe this change fixes the issue. It was due to the loop arguments being updated incorrectly. The current implementation changes `min-len` to `len-arg` and changes `min-len` to `max-len` when `len-arg` exceeds `max-len`. I think rather in this case `max-len` should be updated to `len-arg`, and `min-len` should be left alone, which this PR implements. 

Second, since it was also brought up in issue #884, I'm pretty sure that `max-arg` is updated appropriately since it
represents the index of the argument with the maximum length, and so it is changed only when a new max is found.

The issue also questions the number of args passed to `##raise-length-mismatch-exception`. I think passing the empty list as the first argument is intentional, as this is checked for by `##extract-procedure-and-arguments` in `lib/_kernel.scm`. Since it is `x-y` that is passed as args and not each vector individually, it needs to be concatenated rather than appended to correctly report the error, and the empty list seems to be the indicator to do this. Compare:

```
> (##raise-length-mismatch-exception 3 '() map + '((1) (1 2) (1)))
*** ERROR IN (stdin)@4.1-4.65 -- (Argument 3) Length does not match other arguments
(map '#<procedure #2 +> '(1) '(1 2) '(1))
```
and
```
> (##raise-length-mismatch-exception 3 map + '((1) (1 2) (1)))    
*** ERROR IN (stdin)@5.1-5.61 -- (Argument 3) Length does not match other arguments
(map '#<procedure #2 +> '((1) (1 2) (1)))
```
(the list arguments to the map are wrapped in a list in the second example's error, but joined into the entire expression in the first.)

So the current implementation of the error reporting seems right to me.

I have added corresponding tests. I wasn't sure if there was an established protocol to set variables like `##allow-length-mismatch?` for testing.

### DIFF
--- a/lib/_std#.scm
+++ b/lib/_std#.scm
@@ -1262,8 +1262,8 @@ end-of-code
                                                  (if (##fx> len-arg max-len)
                                                      (loop (##cdr lst)
                                                            (##cons arg rev-x-y)
+                                                           min-len
                                                            len-arg
-                                                           max-len
                                                            arg-num
                                                            (##fx+ arg-num 1))
                                                      (loop (##cdr lst)
@@ -1360,8 +1360,8 @@ end-of-code
                                                  (if (##fx> len-arg max-len)
                                                      (loop (##cdr lst)
                                                            (##cons arg rev-x-y)
+                                                           min-len
                                                            len-arg
-                                                           max-len
                                                            arg-num
                                                            (##fx+ arg-num 1))
                                                      (loop (##cdr lst)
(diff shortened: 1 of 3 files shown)


feeley:
Welcome to Gambit and congratulations on your first PR! Fixes #884.

Issue #759, opened 2022-07-26
Get compiler to inline xcons.


gambiteer:

feeley:
Should be trivial. Are you working on it?

gambiteer:
I'm in the middle of other things (mainly arrays).

feeley:
No problem.  I don't think it is too urgent.

PR #893, opened 2024-02-17
inline xcons

afleckenstein:
I'm trying my hand at #759 

I think it "works" in the sense that if I look at the gvm IR produced by compiling `(xcons 1 2)` I now get:

```
#1 fs=0 entry-point nparams=0 ()            [] r0=#ret
  if (##eq? global[xcons] '#<primitive xcons>) jump fs=0 #4 else #2 [] r0=#ret
#2 fs=0   <- #1                             [] r0=#ret
  r2 = '2                                   [] r0=#ret r2=#
  r1 = '1                                   [] r0=#ret r1=# r2=#
  jump/poll fs=0 #3                         [] r0=#ret r1=# r2=#
#3 fs=0   <- #2                             [] r0=#ret r1=# r2=#
  jump/safe fs=0 global[xcons] nargs=2      [] r0=#ret r1=# r2=#
#4 fs=0   <- #1                             [] r0=#ret
  r1 = (##xcons '1 '2)                      [] r0=#ret r1=#
  jump fs=0 r0                              [] r1=#
```

Is this it and did I take the right approach? Was it right to implement in terms of `^cons` and `CONS`, the latter referring to the macro in `include/gambit.h.in`? I also tried adding an XCONS macro in the header and using this instead (i.e. `targ-apply-xcons` would have `(targ-apply-simp-generator #f #f "XCONS")`) , and making an `^xcons` macro, but it seemed like overkill compared to implementing in terms of cons. Also, is there a way to test these things, or what else do I need to do to fill the PR out?

Once I saw what to do I agree it that it's "trivial" but it was definitely found triviality to me in the sense that it was quite the interesting tour, so I'd appreciate any comments on how this is actually going together because I pretty much just implemented by following what was done for cons. Is it right to frame what I've done as 'installing xcons as a primitive'? how does this get picked up, is it by the `define-prim&proc` machinery somehow? and what is the difference between adding the operator to the C backend in `_t-c-2.scm` and defining the primitive in `_t-univ-4.scm`, how do they relate?

Cheers

### DIFF
--- a/gsc/_t-c-2.scm
+++ b/gsc/_t-c-2.scm
@@ -2572,6 +2572,15 @@
     #f ;; flo-result?
     (targ-apply-simp-generator #f #f "CONS")))
 
+(define (targ-apply-xcons)
+  (targ-apply-alloc
+    (lambda (n) targ-pair-space)
+    #t ;; proc-safe?
+    #f ;; side-effects?
+    #f ;; flo-result?
+    (lambda (opnds sn)
+      (cons "CONS" (reverse (map targ-opnd opnds))))))
+
 (define (targ-apply-list)
   (targ-apply-alloc
     (lambda (n) (* n targ-pair-space))
@@ -3502,6 +3511,7 @@
 ;;; - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - -
 
 (targ-op "##cons"             (targ-apply-cons))
+(targ-op "##xcons"            (targ-apply-xcons))
 (targ-op "##set-car!"         (targ-apply-simp-u #f #t 1 "SETCAR"))
 (targ-op "##set-cdr!"         (targ-apply-simp-u #f #t 1 "SETCDR"))
 (targ-op "##car"              (targ-ifjump-apply-u "CAR"))
--- a/gsc/_t-univ-4.scm
+++ b/gsc/_t-univ-4.scm
@@ -1371,6 +1371,11 @@
    (lambda (ctx return arg1 arg2)
      (return (^cons arg1 arg2)))))
 
+(univ-define-prim "##xcons" #t
+  (make-translated-operand-generator
+   (lambda (ctx return arg1 arg2)
+     (return (^cons arg2 arg1)))))
+
 (univ-define-prim "##set-car!" #f
   (make-translated-operand-generator
    (lambda (ctx return arg1 arg2)
--- a/gsc/tests/04-pair/xcons.scm
+++ b/gsc/tests/04-pair/xcons.scm
@@ -1,0 +1,6 @@
+(declare (extended-bindings) (not constant-fold) (not safe))
+
+(define x (##xcons 11 22))
+
+(println (##car x))
+(println (##cdr x))


feeley:
Here is a complete explanation of how to add a primitive to the Gambit Scheme compiler and runtime system.

First the primitive has to be added to the list of primitives in the file `gsc/_prims.scm` (if it is not already there). I will use `##xcons` as an example:

```
(define prim-procs '
(
;;                                                        lift pattern
;;                                                        /
;;                           side-effects?   strictness  /  result
;;                                       \   pattern    /   type
;; name                   call pattern    \  \         /    /     standard
;;  /                                \     \  \       /    /      |
("##xcons"                            (2)   #f ()    0    pair    extended)
```

Then there needs to be a code generator for inlining this primitive for each backend where that primitive is to be inlined: the C backend (one of `gsc/_t-c-[123].scm`), the universal backend (one of `gsc/_t-univ-[1234]`), and the machine code backends (`gsc/_t-cpu.scm`, ...). For instance `gsc/_t-c-2.scm`:

```
(define (targ-apply-xcons)
   (targ-apply-alloc
     (lambda (n) targ-pair-space)
     #t ;; proc-safe?
     #f ;; side-effects?
     #f ;; flo-result?
     (lambda (opnds sn)
       (cons "CONS" (reverse (map targ-opnd opnds))))))

(targ-op "##xcons" (targ-apply-xcons))
```

Note that the `CONS` we see here is connected to the macro `___CONS` defined in `include/gambit.h.in`:

```
#define ___BEGIN_ALLOC_PAIR()___hp[0] = ___MAKE_HD_WORDS(___PAIR_SIZE,___sPAIR)
#define ___ADD_PAIR_ELEM(i,val)___hp[(i)+___PAIR_BODY] = val
#define ___END_ALLOC_PAIR()___ALLOC(___PAIR_SIZE+___PAIR_BODY)
#define ___GET_PAIR()___PAIR_FROM_END(___hp)

#define ___PAIR_SIZE 2
#define ___PAIR_CDR 0
#define ___PAIR_CAR 1

#define ___ALLOC_PAIR_EXPR(x,y)(___BEGIN_ALLOC_PAIR(), \
___ADD_PAIR_ELEM(___PAIR_CAR,x), \
___ADD_PAIR_ELEM(___PAIR_CDR,y), \
___END_ALLOC_PAIR())

#define ___CONS(x,y)(___ALLOC_PAIR_EXPR(x,y),___GET_PAIR())
```

So a call to `##xcons` will generate C code that calls `___CONS(...,...)`.

Finally the runtime system needs a procedure definition for `##xcons`, so that for example it is possible to `(map ##xcons '(1 2) '(3 4))`. In the case of `##xcons` it is defined in `lib/gambit/list/list.scm`:

```
(define-prim&proc (xcons (d object)
                         (a object))
  (cons a d))
```

The `define-prim&proc` defines both `##xcons` and `xcons`. For the procedure `xcons` it will include type checks to verify that the parameters are of the correct type and raise an exception if that is not the case (note that here no type checks are needed because `d` and `a` can be any object).

Note that the definition of `##cons` and `cons` is different:

```
(define-prim&proc (cons (obj1 object)
                        (obj2 object)))
```

This procedure definition has no body. In this case, the `define-prim&proc` macro will check if the backend knows how to inline `##cons`, and if that's the case then it will treat this the same as:

```
(define-prim&proc (cons (obj1 object)
                        (obj2 object))
  (cons obj1 obj2))
```

This is convenient for all the fundamental primitives that can't be written in terms of other primitives.

Once the compiler is bootstrapped, the definition of `xcons` could be replaced with:

```
(define-prim&proc (xcons (d object)
                         (a object)))
```

Because now the compiler knows how to inline `xcons`. However, it is best to leave the original definition which may help someone reading the code understand what `xcons` does. It may also help with bootstrapping if for some reason an old version of the Gambit compiler needs to be used to compile the runtime system.

For testing there are two things to do. First, a unit test that the procedure `xcons` has the right behaviour (this is file `tests/unit-tests/04-list/xcons.scm`). Secondly, a unit test is required for the compiler (a file like `gsc/tests/04-pair/cons.scm`, which is for `cons`).

gambiteer:
> Secondly, a unit test is required for the compiler (a file like `gsc/tests/04-pair/cons.scm`, which is for `cons`, which you can run with `cd gsc;./runtests.scm tests/04-pair/cons.scm`).

Interesting.  I just ran `cd gsc; ./runtests.scm`; how should I interpret the results.

afleckenstein:
Thanks much for the detailed explanation, I learned a lot. I have added a compiler test, which passes when run with default options. Interesting that it is `/usr/bin/env gsi` running `./gsc`, is there any need to ensure the `gsi` is the one just compiled? I am also curious about the full output of `./run-tests.scm`, I see a few I'm not really sure how to proceed.tests fail on my machine and lots of `*** this primitive not inlined: ##...`

Regarding the machine code backends, I have added code in terms of cons for this, again unsure if this was the right approach. Also, I only have gambit running on an x86-64 machine at the moment. Based on my reading of `run-tests.scm`, I think I have to include the `-x86-64` flag to actually test the behavior of this backend, right? I saw this requires a 'gsc64' binary, i'm not sure exactly where the build for this is defined but I assumed it was something like:

```
$ make clean
$ ./configure '--enable-targets=x86-64'
$ make
```
However this fails for me (on master) with
```
...
make[1]: Entering directory '/home/fleck/src/gambit'
make[2]: Entering directory '/home/fleck/src/gambit/lib'
make[3]: Entering directory '/home/fleck/src/gambit/lib/_define-library'
>>>> building module _define-library for C
>>>> building module _define-library/define-library-expand for C
>>>> building module _define-library/debug for C
>>>> building module _define-library for x86-64
>>>> building module _define-library/define-library-expand for x86-64
*** ERROR -- Compiler internal error detected
*** in procedure Delayed actions that should be executed not reachable
*** module failed to build
>>>> building module _define-library/debug for x86-64
make[3]: Leaving directory '/home/fleck/src/gambit/lib/_define-library'
make[3]: Entering directory '/home/fleck/src/gambit/lib'
make[4]: Entering directory '/home/fleck/src/gambit/lib'
make[4]: *** No rule to make target 'GAMBIT_LIB_x86-64'.  Stop.
make[4]: Leaving directory '/home/fleck/src/gambit/lib'
make[3]: *** [makefile:202: univ-libs] Error 2
make[3]: Leaving directory '/home/fleck/src/gambit/lib'
make[2]: *** [makefile:517: modules-recursive] Error 2
make[2]: Leaving directory '/home/fleck/src/gambit/lib'
make[1]: *** [makefile:373: modules] Error 2
make[1]: Leaving directory '/home/fleck/src/gambit'
make: *** [makefile:107: all] Error 2
```
I saw `No rule to make target 'GAMBIT_LIB_x86-64'. Stop.`, I looked in `lib/makefile` and indeed this variable is not present, though corresponding variables for the C and universal backends are . I assume I am missing something?  I noticed x86-64 is given as a default in a few places (e.g. BUILD_TARGETS in `lib/makefile` itself), but `./configure` default is js (according to both source and --help). But then in the root level `makefile` `CONFIGURE_COMMAND` enables x86-64 by default, using the same command I tried above. So I'm not really sure how to proceed.

gambiteer:
I haven't been following well, has this been fixed?

afleckenstein:
I am going to open a new PR with a cleanly formatted history. @feeley 's instructions seem wiki-worthy?

PR #897, opened 2024-03-13
Allow ~ in module refs

vyzo:
Following our discussion on gitter, I need this to unblock my work on integrating the loader.


feeley:
I'm currently working on an extension to the module names. It is too permissive to allow `~` anywhere in the module name. I should have something soon.

PR #903, opened 2024-04-20
SRFI-128 + 162 Comparators

afleckenstein:
Hello,

There are lots of interesting data structure SRFIs that I would find convenient to have in Gambit, many of them (113 Sets and Bags, 146 Mappings, 224 Dictionaries to name a few) depend on having the Comparators SRFI, hence this PR. The core functionality is just a copy of the sample implementation.

The main gambit-specific tweaks are:
- using gambit's built-in hashing functions (for the most part, see below)
- adding a keyword comparator for gambit's keyword data type

It seems to work, and I was just wondering if there were any further gambit specific enhancements to make (I consider numeric vectors at the end)

There is one problem related to some hashing code, which I understand to be used in order to hash pairs/sequences while respecting the comparator on the elements

```
$ gsi/gsi -:= lib/srfi/128/test/test.scm
testing vector/cdr
testing comparators
| testing comparators/predicates
| testing comparators/constructors
| testing comparators/hash
| testing comparators/default
| testing comparators/accessors
| testing comparators/invokers
| testing comparators/comparison
| testing comparators/syntax
| testing comparators/bound-salt
"lib/srfi/128/test/test.scm"@225.5-225.47: FAILED (test-assert (exact-integer? (hash-salt))) GOT *** ERROR IN #<procedure #2>, "lib/srfi/128/128-impl.scm"@182.19-182.25 -- Unbound variable: %salt%
   ┏━━━━━━━━━━━━━━━━━━━━━━━━ lib/srfi/128/128-impl.scm ━━━━━━━━━━━━━━━━━━━━━━━━
   ┃⋯
181┃  (syntax-rules ()
182┃    ((hash-salt) (%salt%))))
   ┃                  ▔▔▔▔▔▔
183┃
   ┃⋯
```

If I export the `%salt%` identifier, the tests pass but this seems counter to the point of the implementation. I know that there is current work to complete the implementation of hygienic macros. I am wondering if I can expect this problem to go away when that is released, or if I am missing something, or even if there is a better gambit-specific way to build these sequence hashers altogether.

Regarding homogenous numeric vectors, SRFI-160 mentions providing comparators for these. I think comparators for homogenous vectors would be useful, but I am just unsure whether to do it in the context of this SRFI, SRFI-160, or just leave it for now.

Cheers


feeley:
Thanks for the PR!

Concerning the hygiene issue, the problem should go away when the hygienic macro expander is integrated. For now there is a simple workaround that consists in using Gambit's `namespaces`. The way `define-library` is implemented in Gambit, the library `(srfi 128)` will generate identifiers in the `srfi/128#` namespace. This means that code outside of the library can access the `%salt%` variable using the fully qualified identifier `srfi/128#%salt%`, without needing to be imported. So it is sufficient to write:

```
(syntax-rules ()
  ((hash-salt) (srfi/128#%salt%))) ;; was: %salt%
```

This is how the hygienic macro expander will work behind the scenes. For consistency, you should also start the file `128#.scm` with `(##namespace "srfi/128#"` .

For this particular issue, another solution to consider is to do `(define hash-salt (make-parameter ...))` and use `hash-salt` (which is exported) wherever `%salt%` is currently used. The library should have been defined this way in the first place, but since it wasn't, I guess that the previous solution is best.

Adding support for homogeneous vectors would be nice of course.

Please fix this and I will accept the PR.

feeley:
One more thing. `test-eq` which you use in the tests should not be used with numbers. Please use `(test-eqv ...)` or `(test-equal ...)` or `(test-assert (= ...))`.

afleckenstein:
I've added the namespace export. I have opted to parameterize `hash-salt` itself because I also thought it would be fine to do it this way. Thanks for catching the `eq` problem.

I've added support for homogenous numeric vectors as well. Should they be installed in the default comparator?

afleckenstein:
> Should they be installed in the default comparator?

This was before I remembered that `comparator-register-default!` offers a convenient way to do this... I have gone ahead and done so

PR #905, opened 2024-04-24
inline xcons

afleckenstein:
Issue #759, this PR is a cleaned up version of #893 (the commit history of that PR is a faff)

Code generation for `xcons` is added to the C and universal backends.

I did nothing regarding the machine code backends, partially because I could not figure out how to build/test the x86-64 backend and partially because of this message from @feeley in the gambit chatroom after I asked about the native backends:

> Yes indeed the native backends are special in that they don’t need to implement any primitives (except the really basic `##not` and `##identity` that are required for the GVM `if` instruction). Of course, to a lesser extent, the other backends can fallback on the primitive procedures (e.g. `xcons`) instead of inlining those primitives.

### DIFF
--- a/gsc/_t-c-2.scm
+++ b/gsc/_t-c-2.scm
@@ -2578,8 +2578,7 @@
     #t ;; proc-safe?
     #f ;; side-effects?
     #f ;; flo-result?
-    (lambda (opnds sn)
-      (cons "CONS" (reverse (map targ-opnd opnds))))))
+    (targ-apply-simp-generator #f #f "XCONS")))
 
 (define (targ-apply-list)
   (targ-apply-alloc
--- a/gsc/_t-c-2.scm
+++ b/gsc/_t-c-2.scm
@@ -2572,6 +2572,15 @@
     #f ;; flo-result?
     (targ-apply-simp-generator #f #f "CONS")))
 
+(define (targ-apply-xcons)
+  (targ-apply-alloc
+    (lambda (n) targ-pair-space)
+    #t ;; proc-safe?
+    #f ;; side-effects?
+    #f ;; flo-result?
+    (lambda (opnds sn)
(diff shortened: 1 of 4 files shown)


gambiteer:
I looked at this thinking I would apply it, but now have some questions about the approach.
The last argument to `targ-apply-alloc` here concerns me:
```
(define (targ-apply-xcons)
  (targ-apply-alloc
    (lambda (n) targ-pair-space)
    #t ;; proc-safe?
    #f ;; side-effects?
    #f ;; flo-result?
    (lambda (opnds sn)
      (cons "CONS" (reverse (map targ-opnd opnds))))))
```
Instead of calling `targ-apply-simp-generator`, which in turn calls `targ-apply-simp-gen` to construct the C backend call, you construct the code explicitly.  If, for some reason, `targ-apply-simp-generator` and `targ-apply-simp-gen`, which are both used for many GVM instructions, were later changed for some systemic reason, then the proposed `targ-apply-xcons` won't see that change.

In Marc's original comments, did he propose a new C macro `XCONS`, similar to `CONS`, but with reversed arguments, to be used in `targ-apply-xcons`?  That approach might be more complicated, but might also avoid later problems.

afleckenstein:
@gambiteer I briefly noted that I did try creating an XCONS macro which worked, but was unsure if XCONS deserved one. I don't think Marc said anything specific on this issue. Perhaps something like

```
(let ((cons-gen (targ-apply-simp-generator #f #f "CONS")))
  (lambda (opnds sn)
    (cons-gen (reverse opnds) sn)))
```

could work without needing the extra C macro?

gambiteer:
I prefer you put in the `XCONS` C macro and have `targ-apply-xcons` follow the pattern of the other code generators.
If you do that and it passes tests, I'll apply it.
Thanks.

Brad

gambiteer:
Merged, thanks.

PR #918, opened 2024-08-28
##os-executable-path on OpenBSD

maconcg:
Hi,

I recently spent some time looking into a test failure and related on quirks on OpenBSD 7.5 and wanted to share (some of) my fixes.

It looks like the behavior I noticed stems from the fact that OpenBSD does not implement a /proc filesystem, but there's an assumption to the contrary in os_files.c.  The first commit directly addresses this by adding a check for `__OpenBSD__` to the relevant ifdef.  It results in more accurate "unimplemented" error messages.

The second commit removes ##create-temporary-directory's dependency on ##executable-path (and thus on ##os-executable-path).  The replaced code is currently used to generate default paths and would usually evaluate to either "gsc-temp" or "gsi-temp"; the replacement is just a hard-coded "gsx-temp".

I've also been using another patch that replaces the application of `(##os-executable-path)` in the definition of ##default-path-expand with `(##string-append (##os-path-gambitdir) "/bin/gambit")`. In most cases, it makes path expansion work as it should on OpenBSD, but for now I'm not including it in this PR; it assumes "execdir" ends with "/bin" and could therefore introduce errors under certain installation configurations.

Please do share any feedback/suggestions, especially re: fixing path expansion.

### DIFF
--- a/lib/_nonstd.scm
+++ b/lib/_nonstd.scm
@@ -3067,16 +3067,19 @@
 (define-prim (executable-path)
   (##executable-path))
 
-(define-prim (##process-mnemonic-identifier)
-  (let ((path (if (##pair? ##processed-command-line)
-                  (##car ##processed-command-line)
-                  (let ((exec-path (##os-executable-path)))
-                    (if (##string? exec-path)
-                        exec-path
-                        #f)))))
-    (if path
-        (##path-strip-extension (##path-strip-directory path))
-        (##string-append "process" (##number->string (##os-getpid))))))
+(define-primitive (##process-mnemonic-identifier)
+  (let* ((pcl ##processed-command-line)
+	 (path (if (##pair? pcl)
+		   (##car pcl)
+		   (let ((exec-path (##os-executable-path)))
+		     (if (##string? exec-path)
+			 exec-path
+			 #f)))))
+    (##string-append (if path
+			 (##path-strip-extension (##path-strip-directory path))
+			 "process")
+		     "-"
+		     (##number->string (##os-getpid)))))
 
 ;;;----------------------------------------------------------------------------
 
@@ -3109,7 +3112,7 @@
            (let* ((prefix
                    (or path
                        (##path-expand
-                        (##string-append (##process-mnemonic-identifier) "-temp")
+                        (##process-mnemonic-identifier)
                         (##os-path-tempdir))))
                   (pid
                    (##os-getpid))
--- a/lib/_nonstd.scm
+++ b/lib/_nonstd.scm
@@ -3067,7 +3067,7 @@
 (define-prim (executable-path)
   (##executable-path))
 
-(define-primitive (##process-mnemonic-identifier)
+(define-primitive (process-mnemonic-identifier)
   (let* ((pcl ##processed-command-line)
 	 (path (if (##pair? pcl)
 		   (##car pcl)
--- a/lib/_nonstd.scm
+++ b/lib/_nonstd.scm
@@ -3067,6 +3067,17 @@
 (define-prim (executable-path)
   (##executable-path))
 
+(define-prim (##process-mnemonic-identifier)
+  (let ((path (if (##pair? ##processed-command-line)
+                  (##car ##processed-command-line)
+                  (let ((exec-path (##os-executable-path)))
+                    (if (##string? exec-path)
+                        exec-path
+                        #f)))))
+    (if path
+        (##path-strip-extension (##path-strip-directory path))
+        (##string-append "process" (##number->string (##os-getpid))))))
+
 ;;;----------------------------------------------------------------------------
 
 ;;; Filesystem operations.
@@ -3098,7 +3109,7 @@
            (let* ((prefix
                    (or path
                        (##path-expand
-                        "gsx-temp"
+                        (##string-append (##process-mnemonic-identifier) "-temp")
                         (##os-path-tempdir))))
                   (pid
                    (##os-getpid))
--- a/lib/_nonstd.scm
+++ b/lib/_nonstd.scm
@@ -2700,7 +2700,11 @@
                        (expand relpath
                                expanded-dir)))
                     ((##string=? instdir-name "execdir")
-                     (expand-in-instdir relpath "bin"))
+                     (let ((exec-path (##os-executable-path)))
+                       (if (##fixnum? exec-path)
+                           (err exec-path)
+                           (expand relpath
+                                   (##path-directory exec-path)))))
                     (else
                      (let ((dir (##os-path-gambitdir)))
                        (if (##fixnum? dir)
--- a/lib/_nonstd.scm
+++ b/lib/_nonstd.scm
@@ -3098,9 +3098,7 @@
            (let* ((prefix
                    (or path
                        (##path-expand
-                        (##string-append
-                         (##path-strip-directory (##executable-path))
-                         "-temp")
+                        "gsx-temp"
                         (##os-path-tempdir))))
                   (pid
                    (##os-getpid))
--- a/lib/_nonstd.scm
+++ b/lib/_nonstd.scm
@@ -2700,11 +2700,7 @@
                        (expand relpath
                                expanded-dir)))
                     ((##string=? instdir-name "execdir")
-                     (let ((exec-path (##os-executable-path)))
-                       (if (##fixnum? exec-path)
-                           (err exec-path)
-                           (expand relpath
-                                   (##path-directory exec-path)))))
+                     (expand-in-instdir relpath "bin"))
                     (else
                      (let ((dir (##os-path-gambitdir)))
                        (if (##fixnum? dir)
--- a/lib/_nonstd.scm
+++ b/lib/_nonstd.scm
@@ -3120,7 +3120,6 @@
                (let* ((resolved-path
                        (##path-resolve
                         (##string-append prefix
-                                         (##number->string pid)
                                          (if (##fx= i 0)
                                              ""
                                              (##number->string i)))))
--- a/lib/_nonstd.scm
+++ b/lib/_nonstd.scm
@@ -3112,10 +3112,8 @@
            (let* ((prefix
                    (or path
                        (##path-expand
-                        (##process-mnemonic-identifier)
+                        (##string-append (##process-mnemonic-identifier) "-temp")
                         (##os-path-tempdir))))
-                  (pid
-                   (##os-getpid))
                   (permissions
                    (##psettings->permissions psettings #o777)))
              (let loop ((i 0))
--- a/lib/_nonstd.scm
+++ b/lib/_nonstd.scm
@@ -3067,7 +3067,7 @@
 (define-prim (executable-path)
   (##executable-path))
 
-(define-primitive (process-mnemonic-identifier)
+(define-prim (##process-mnemonic-identifier)
   (let* ((pcl ##processed-command-line)
 	 (path (if (##pair? pcl)
 		   (##car pcl)
--- a/lib/os_files.c
+++ b/lib/os_files.c
@@ -1740,7 +1740,7 @@ ___SCMOBJ ___os_executable_path ___PVOID
 
 #endif
 
-#if defined (USE_readlink) && defined (USE_getpid) && !(defined (__OpenBSD__))
+#if defined (USE_readlink) && defined (USE_getpid) && !defined (__OpenBSD__)
 
   {
     pid_t pid = getpid ();
--- a/lib/os_files.c
+++ b/lib/os_files.c
@@ -1740,7 +1740,7 @@ ___SCMOBJ ___os_executable_path ___PVOID
 
 #endif
 
-#if defined (USE_readlink) && defined (USE_getpid)
+#if defined (USE_readlink) && defined (USE_getpid) && !(defined (__OpenBSD__))
 
   {
     pid_t pid = getpid ();


lassik:
Generally speaking, `##os-executable-path` is an anti-pattern on Unix-like operating systems. I once put considerable effort into writing a portable version of such a function, and people with more experience pointed out that there are unavoidable gotchas. IIRC, the discussion was on the SRFI 193 mailing list.

maconcg:
> Generally speaking, `##os-executable-path` is an anti-pattern on Unix-like operating systems.
Very much agree.  That's why I took the approach of eliminating calls to procedures that depend on `##os-executable-path`; it seems to me that in all cases, the calling procedures can be made more portable (without sacrificing essential functionality) by removing those calls.

feeley:
`##os-executable-path` is also useful in other contexts. One that is particularly useful in some environments and software installation approaches is to find files relative the the location where the executable program is installed. This is made concrete in the syntax `~~execdir` for paths that refers to the directory containing the program executing:

```
$ gsi -e '(pp (directory-files (path-expand "~~execdir/..")))'
("bin" "include" "info" "lib" "doc" "share")
```

So while I agree `##os-executable-path` can't be implemented on all platforms, it has a moderately high portability that makes it interesting to support.

What I could see is a better API for `(executable-path)`. It should fail when uninmplemented (as it is currently), but it should take a second optional parameter that is the value that is returned when not implemented, or the path can't be determined. So `(executable-path "foo")` would return `"foo"` as a fallback. The pattern `(or (executable-path #f) (find-exec-path-some-other-way))` comes to mind.

lassik:
NB: I originally intended to add a procedure just like `os-executable-path` to SRFI 193, and wrote the code to do so, but the people on the mailing list convinced me to leave it out.

feeley:
I can understand leaving it out of a SRFI that covers several features and is meant to be fully portable. Gambit can and should implement features that are only supported on some platforms, even if Gambit has a tradition to strive to abstract features as much as possible to hide low-level differences. For example when opening a process there is a `show-console:` setting that only applies to Windows and a `pseudo-terminal:` setting that only applies to UNIX. It would be foolish to avoid implementing features just because they are not 100% portable. The programmer using Gambit is well placed to decide whether using a specific non-portable feature is appropriate.

feeley:
> NB: I originally intended to add a procedure just like `os-executable-path` to SRFI 193, and wrote the code to do so, but the people on the mailing list convinced me to leave it out.

I'd be interested to see your code and integrate the heuristics into Gambit's `##os-executable-path`.

feeley:
> > Generally speaking, `##os-executable-path` is an anti-pattern on Unix-like operating systems.
> 
> Very much agree. That's why I took the approach of eliminating calls to procedures that depend on `##os-executable-path`; it seems to me that in all cases, the calling procedures can be made more portable (without sacrificing essential functionality) by removing those calls.

Just to be clear I agree that `##os-executable-path` should be used with care in features that can and should be 100% portable. I think extending it with a second optional "fallback" parameter would achieve this goal.

lassik:
> I'd be interested to see your code and integrate the heuristics into Gambit's `##os-executable-path`.

If memory serves, the nearest code is in [my `exename` repo](https://github.com/lassik/exename).

A detailed rationale, which was to be included in a draft of SRFI 193, is [in this commit](https://github.com/scheme-requests-for-implementation/srfi-193/commit/4381a811088da8d472b3b36b644c6dfa2f381bc3).

vyzo:
linux: /proc/self/exe
darwin: there is libproc afaik, so doable.
bsd: trouble

maconcg:
> `##os-executable-path` is also useful in other contexts. One that is particularly useful in some environments and software installation approaches is to find files relative to the location where the executable program is installed. This is made concrete in the syntax `~~execdir` for paths that refers to the directory containing the program executing:
> 
> ```
> $ gsi -e '(pp (directory-files (path-expand "~~execdir/..")))'
> ("bin" "include" "info" "lib" "doc" "share")
> ```

I think my original wording ("in all cases") was too broad.  That wording (especially in combination with this PR's title) steered the discussion toward the general utility of the procedure and whether Gambit should implement it _at all_.

A better choice of words might have been "in all cases _where ##os-executable-path is called by another primitive procedure_."  This should narrow the scope of the problem to two procedures that currently depend on ##os-executable-path, thereby harming portability, without (imo) good reason:

- One is `##create-temporary-directory`.  My take is that its use of `##os-executable-path` provides a minor aesthetic benefit (by using the executable name as part of the name of the temporary directory) at the cost of portability.  This is a bad tradeoff, imo.
- The other is `##default-path-expand`.  This procedure uses `##os-executable-path` in a way similar to @feeley's example (i.e., to find the "bin" directory).  This use isn't merely aesthetic, but could probably be replaced by something more portable (and by "more portable" I guess I just mean "works on OpenBSD").

> What I could see is a better API for `(executable-path)`. It should fail when uninmplemented (as it is currently), but it should take a second optional parameter that is the value that is returned when not implemented, or the path can't be determined. So `(executable-path "foo")` would return `"foo"` as a fallback. The pattern `(or (executable-path #f) (find-exec-path-some-other-way))` comes to mind.

My understanding is that (at least on OpenBSD) for the foreseeable future there will not be a viable way to implement `find-exec-path-some-other-way` (see [this post](https://github.com/ziglang/zig/issues/6718#issuecomment-711134120).  This tells me the fallback would need to be implemented in the procedure that calls `##os-executable-path` (i.e., `##default-path-expand`).

The alternative to implementing such a fallback is to change the behavior for all systems.  This is what I did with the `##create-temporary-directory` patch.  The benefit to this approach is that we don't introduce per-system differences.

For the case of `##default-path-expand`, all we really need to do is find the directory containing the Gambit executables.  My preferred solution would be to replace the current "by-way-of-##os-executable-path" approach with the use of a constant value set by the `configure` script.  A disadvantage of my preferred solution is that it would no longer work if you copy the executable to a new path after installation.  As far as I can tell, a compromise solution that first tries using `##os-executable-path` and then falls back to a constant value would only benefit the user who copies the executable to a new location after installation.

What are your thoughts?

feeley:
I agree with the principle that an unimplemented `##os-executable-path` should not cause failure of primitive procedures that can and must be portable, such as `create-temporary-directory`. On the other hand, it does provide a useful cue when the temporary directory's name is related to the executable program's name.

Perhaps using `##os-executable-path` is not the best way to get this information reliably. For example, using the command-line may be better since we don't care about the executable path's directory. Something like `(path-strip-extension (path-strip-directory (car ##processed-command-line)))`. When this doesn't work (when `##get-command-line` does not return a pair) then `##os-executable-path` could be tried, and if that fails then using the name `process<pid>` might be a reasonable fallback.

lib/_nonstd.scm
```diff
@@ -3098,9 +3098,7 @@
            (let* ((prefix
                    (or path
                        (##path-expand
-                        (##string-append
-                         (##path-strip-directory (##executable-path))
-                         "-temp")
+                        "gsx-temp"
```
feeley, line 3101: Using a fixed name, and in particular `gsx-temp`, is not a good idea because it will not be related to the program that created the directory.

A primitive procedure `##process-identifier-string` should be added that returns a string related to the program's name and computed in steps, returning the first that doesn't fail :

1) if `##processed-command-line` is a pair, then `(##path-strip-extension (##path-strip-directory (##car ##processed-command-line)))`
2) same for `(##os-executable-path)`
3) `"process"`

and then the pid is added as a suffix with `(##string-append ... "-" (##number->string (##os-getpid)))`.

Then `##create-temporary-directory` can be changed to avoid adding the pid.

maconcg: This makes sense and is definitely the smarter solution.  My thought process in choosing "gsx-temp" was essentially "the smaller the change, the more likely it'll be accepted."  Rookie mistake.

feeley: No problem. By the way the name `##process-identifier-string` is a bit confusing because it could be interpreted as the process PID in string form. So maybe `##process-mnemonic-identifier` is better. Sufficiently unusual that it won't be confused with more common process functionality.

maconcg:
I added a couple commits just now.  Will also take a stab at unit tests/documentation.  A few questions that came up:

1. I had some difficulty deciding which binding form to use for the `##process-mnemonic-identifier` procedure. I ended up going with `define-prim` largely because that's the most common form in `_nonstd.scm` and it seemed to work, but only after considering `define`, `define-primitive`, `define-prim&proc`, etc. Is there a "rule of thumb" I could have used here?
2. If my move to "alias" ~~execdir to ~~bin is valid, I think we might have something that would work as a fallback to try in `##executable-path` before resorting to `##raise-os-exception`.  Basically: it seems like most of the time the value of `(##string-append (##default-path-expand "~~bin") (##process-mnemonic-identifier))` will match the value of `(##os-path-expand)`.  Neither is the _right_ way to get that path (becaulse there isn't one), but it should _usually_ result in the same behavior across systems.  Would adding this fallback be a good idea?

gambiteer:
> I had some difficulty deciding which binding form to use for the `##process-mnemonic-identifier` procedure. I ended up going with `define-prim` largely because that's the most common form in `_nonstd.scm` and it seemed to work, but only after considering `define`, `define-primitive`, `define-prim&proc`, etc. Is there a "rule of thumb" I could have used here?

All these macros are defined in  `_gambit#.scm`.  I don't know all the details, but here is how I understand it.

`define-procedure`: Apply argument checks based on "type" notations added to each argument; the possible "types" are predefined.  Some of these must still be checked inline (like list of bool/fixnum/...).  Declare `(not safe)` and also `(##include "~~lib/gambit/prim/prim#.scm")`, which, e.g.,  turns `fx+` (which normally checks its arguments and results) to `##fx+` (which does not.

`define-primitive`: Do the same without argument checks to define a procedure `##name`, i.e., prepend `##` to the name you gave to `define-primitive`.  These non-error-checking procedures are used internally by the runtime.  (Beginning a name with `##` is not allowed in Scheme, so a user program calling these procedures is technically invalid, but allowed by Gambit.)

`define-prim&proc`: Do both, defining a user-facing version that checks argument types, and an internal version with the same name prepended with `##`.

gambiteer:
I forgot to say, if you use `(define-prim&proc (name ...) ...)`, then calling `(name ...)` just performs argument checks and then tail calls `(##name ...)` with the same arguments.

And `define-prim`, which is also defined in `_gambit#.scm`, is earlier and more primitives that the ones I mentioned.

lib/_nonstd.scm
```diff
@@ -3067,6 +3067,17 @@
 (define-prim (executable-path)
   (##executable-path))
 
+(define-prim (##process-mnemonic-identifier)
+  (let ((path (if (##pair? ##processed-command-line)
```
feeley, line 3067: There's a subtle bug here. The code is not thread safe. Another thread could change `##process-mnemonic-identifier` between the test for pair and the call to car. Also, the PID should be added to the name in all cases. So this definition is better:
```
(define-primitive (##process-mnemonic-identifier)
  (let* ((pcl
          ##processed-command-line)
         (path
          (if (##pair? pcl)
              (##car pcl)
              (let ((exec-path (##os-executable-path)))
                (if (##string? exec-path)
                    exec-path
                    #f)))))
    (##string-append
     (if path
         (##path-strip-extension (##path-strip-directory path))
         "process")
     "-"
     (##number->string (##os-getpid)))))
```

lib/_nonstd.scm
```diff
@@ -3098,7 +3109,7 @@
            (let* ((prefix
                    (or path
                        (##path-expand
-                        "gsx-temp"
+                        (##string-append (##process-mnemonic-identifier) "-temp")
                         (##os-path-tempdir))))
                   (pid
```
feeley, line 3110: The adding of the PID should no longer be done by `##create-temporary-directory`.

lib/_nonstd.scm
```diff
@@ -2700,11 +2700,7 @@
                        (expand relpath
                                expanded-dir)))
                     ((##string=? instdir-name "execdir")
-                     (let ((exec-path (##os-executable-path)))
-                       (if (##fixnum? exec-path)
-                           (err exec-path)
-                           (expand relpath
-                                   (##path-directory exec-path)))))
+                     (expand-in-instdir relpath "bin"))
```
feeley, line 2703: This is plain wrong. It goes against the reason behind `~~execdir`, which is to locate the directory containing the executable when other related files are relative to that.

You have to understand that `gsi` and `gsc` are just 2 special programs that are built with Gambit. In general a program built with Gambit could be installed anywhere, not in the directory `~~bin` which is the directory where Gambit specific programs are installed, when they are installed in the usual way. But even `gsi` and `gsc` could be built in one location (for a "binary" distribution) and then installed on a user's machine to a user preferred location, so the `~~execdir` can be useful to allow Gambit specific files to be found. Specifically, if a Gambit installation has been `mv`ed to a different place and `gsi` is accessible in your `PATH` you can still execute it with `gsi -:~~=~~execdir/..` (which defines `~~` to be the parent directory of the exec dir, i.e. if the exec dir is `/foo/bar/bin` then `~~bin` will resolve to `/foo/bar/bin`, `~~lib` will resolve to `/foo/bar/lib`, etc).

lib/_nonstd.scm
maconcg: Got it -- thanks. (Note for myself: in the fixed version, binding `pcl` lets us write the procedure with only one reference to the free variable `##processed-command-line`.)

lib/_nonstd.scm
maconcg: Ok; where we used to see "`-temp`" at the end of the temporary directory's name, we'll now see "`-pid`" (please let me know If I've misunderstood).

feeley: I was referring to line 3114 which gets the pid and later uses it, which is no longer necessary. Line 3112 is OK, so the general pattern for the directory name is `<executable-name>-<pid>-temp`.

lib/_nonstd.scm
```diff
@@ -3067,7 +3067,7 @@
 (define-prim (executable-path)
   (##executable-path))
 
-(define-primitive (##process-mnemonic-identifier)
+(define-primitive (process-mnemonic-identifier)
```
feeley, line 3070: For consistency with the other definitions you should do `(define-prim (##process-mnemonic-identifier) ...` rather than `(define-primitive (process-mnemonic-identifier) ...`.

The new style of primitive definition is `(define-primitive ...` but I would prefer having a single refactoring of the `_nonstd.scm` file which switches from `define-prim` to `define-primitive` when the time has come to do that.

maconcg:
@gambiteer thanks for the pointers.  I checked the commit history and, indeed, `define-prim` is the oldest of these macros (`define-{procedure,proc,prim&proc}` introduced by `050bdfaa7`).

This implies the definitions that were throwing me off (`##executable-path` and `executable-path`, added by commit d3901a6) use `define-prim` because `define-prim&proc` did not yet exist.

So now I have two more questions:

1. Part of me thinks it would make sense to replace my use of `define-procedure` with `define-prim&proc` so the progammer doesn't need to call a `##` procedure to see how temporary directories will be named.  Is this a good idea?

2. I feel an instinctive urge toward the following course of action:
  - Start reversing the bitrot (if I'm correct to call it bitrot) by replacing uses of `define-prim` with their newer counterparts.  Document `define-prim&proc` and maybe add some examples.
Another course of action, preferred by some, would be:
  - Document `define-prim`, `define-prim&proc`, etc.  Avoid "fixing that which is not broken" by leaving the `define-prims` in place.
What is Gambit's philosophy here?

lib/_nonstd.scm
maconcg: Ah, okay.  You posted this while I was typing up another (related) comment, so we are a bit off sync.  In a code snippet you shared earlier, you used `(define-primitive (##process-mnenomic-identifier)`.  After I saw that, I figured out the "new style" vs. "old style" distinction, and then assumed you were trying to steer me to the "new style" but had made a typo.

lib/_nonstd.scm
maconcg: Reverted this one, so for now that clause will just raise an exception if `##os-executable-path` doesn't work.

Thank you for the clear explanation.  Especially the first sentence of the 2nd paragraph.  On some level I _knew_ that, but that wasn't how I was _thinking about_ the problem...if that makes sense?

feeley:
Thanks for your contribution! It took quite a bit of back-and-forth, but it is all a good investment in getting you accustomed to the "Gambit philosophy" of writing the libraries.

maconcg:
> Thanks for your contribution! It took quite a bit of back-and-forth, but it is all a good investment in getting you accustomed to the "Gambit philosophy" of writing the libraries.

I very much appreciate you taking the time for that back-and-forth.

gambiteer:
There were 11 commits packaged into this pull request, to add one small procedure and change two single lines. That's not a criticism, just an observation.
But it makes the git history somewhat hard to read.  I'm not a Git master, but I've heard of "squashing" a pull request, or something, maybe something like this should be applied in situations like this with a meandering and somewhat complicated commit history.

lassik:
I also recommend "squashing" the original commits into one commit unless the PR is really complicated.

maconcg:
Totally fair criticism.  I actually _did_ squash some commits in there, or else it would have been even higher.  My concern was that the conversation would be hard to follow if they weren't there.  Arguably erred too far on the side of caution.

Maybe someone with the necessary privileges could take a look at [enabling this through GitHub](https://docs.github.com/en/repositories/configuring-branches-and-merges-in-your-repository/configuring-pull-request-merges/configuring-commit-squashing-for-pull-requests)?

feeley:
All of the options are currently enabled (and were that way):

image

lassik:
GitHub's "merge" button has a "squash" option in the drop-down menu. That's the usual way to do it. This implies the squashing is done by the guy merging the PR, not the guy submitting the PR.

lassik:
All of "merge", "squash", and "rebase" are enabled by default on GitHub. The above settings screen lets admins disable them. This is rarely needed. The "merge" button on each PR lets you select any one of the three strategies separately for each PR if you like. By default, the button remembers the strategy last used.

feeley:
Good to know!

PR #919, opened 2024-08-28
Don't automatically include $PWD/.gambini

maconcg:
While going over the manual today, I noticed a feature I had previously overlooked in §2.4:
>  In all modes, the interpreter first tries to locate the initialization file by searching the following locations: ‘.gambini’ and ‘~/.gambini’...The first file that is found is examined as though the expression (include initialization-file) had been entered at the read-eval-print loop where initialization-file is the file that was found.

This reminded me of a [a recent issue related to Fedora's Lua package](https://bugzilla.redhat.com/show_bug.cgi?id=2177239).  The situation described there is similar to Gambit's present behavior: a user might unknowingly execute code by running gsi in a certain directory.

Here's a Gambit-specific example of how this works:
```
$ cd $(mktemp -d)
$ echo '(create-directory "could-be-worse")' >.gambini
$ ls
$ gsi -e '(pp (features))'
```
The user, who just wanted to see a list of supported features, will have silently created a new directory.  Could be worse.

For greater effect, you might mentally replace that echo command with `tar -xzf nefarious.tgz` or something similar.  Also note that because ".gambini" begins with a period, it's not included in the default `ls` output.

In the Lua bug I shared earlier, no patch was applied because it could break things for users who depend on this behavior.  Which is a legitimate objection that may also apply to Gambit. _My_ opinion, however, is that the feature is at best annoying (I'd rather not have a dotfile in my working directory inhibit the inclusion of my customizations in `~/.gambini`) and at worst...well, you know.

Patching out this behavior looks trivial -- unless I'm missing something, this PR is all it would take.

### DIFF
--- a/gsi/main.scm
+++ b/gsi/main.scm
@@ -163,8 +163,7 @@ usage-end
                #f
                #f)))
 
-    (or (try (macro-initialization-file))
-        (try (in-homedir (macro-initialization-file)))))
+    (try (in-homedir (macro-initialization-file))))
 
   (define (read-source-from-string str name)
     (let ((port
--- a/doc/gambit.txi
+++ b/doc/gambit.txi
@@ -499,11 +499,11 @@ An error is not signaled when the file does not exist.  Interpreter
 extensions and patches that are meant to apply to all users and all
 modes should go in that file.
 
-Extensions which are meant to apply to a single user or to a specific
-working directory are best placed in the @dfn{initialization file},
+Extensions which are meant to apply to a single user
+are best placed in the @dfn{initialization file},
 which is a file containing Scheme code.  In all modes, the interpreter
 first tries to locate the initialization file by searching the
-following locations: @file{.gambini} and @file{~/.gambini} (with no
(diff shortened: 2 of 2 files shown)


gsi/main.scm
```diff
@@ -163,8 +163,7 @@ usage-end
                #f
                #f)))
 
-    (or (try (macro-initialization-file))
-        (try (in-homedir (macro-initialization-file)))))
+    (try (in-homedir (macro-initialization-file))))
```
feeley, line 166: The local definition of the procedure `in-homedir` should also be removed.

feeley (commented):
I agree that the init file should not be searched in the current directory. This is a security issue that dates back to Gambit's early days (circa 1990) when security issues where not on the radar.

gsi/main.scm
feeley: Sorry... it should be kept!

feeley:
Thank you for your first approved contribution to Gambit!

Issue #921, opened 2024-09-03
"Double echo" in line editor mode under Emacs


maconcg:
Hi,

I've noticed a line editor mode quirk that I believe was introduced with commit 556a884d.  When I evaluate an expression at the REPL, both the expression and the value are printed where only the value should be printed.  Example:

```
> (+ 1 1)
(+ 1 1)
2
```

This happens with `emacs -q` and in several modes (including term mode, shell mode, CMU scheme mode, and geiser).

I've been able to semi-fix the quirk by modifying [these lines of os_tty.c](https://github.com/gambit/gambit/blob/898028d1a5cdf9ffe9a9dc6d9b2257d00e18fe3d/lib/os_tty.c#L5320-L5329).  My semi-fix is to just prevent the execution of `d->lineeditor_mode = LINEEDITOR_MODE_SCHEME;` under Emacs (e.g., by wrapping the #ifdef in `else { ... }`).  It's only a _semi_-fix because this breaks any REPL run from Emacs' term mode.

To implement a "real" fix, I considered adding some granularity to `lineeditor_under_emacs` by checking for the presence of `comint` in the value of `$INSIDE_EMACS` (rather than just checking whether that environment variable is set). That way, lineediting_mode could be disabled only in comint-derived modes.  This would probably require the definition of a new `___env_var_contains_UCS_2` function(?).

My more immediate problem, however, is that I don't understand the intent behind `lineeditor_under_emacs` in its current form.  Under what Emacs mode would one get the correct result by running a REPL where lineeditor_mode = 1 and input_echo = 0?

Also: there's a good chance I've misunderstood some things.  I may not even be asking the right questions.  If that's the case, just prod me in the right direction.

lassik:
#488 may be related

maconcg:
> #488 may be related

Could be.  Typescript outputs (created via `script -c <gsi>`) are attached, plus visual-format versions for people who don't have `vis(1)` installed.  These were run with a gsi built from git with no local modifications.

`term-mode` output is identical to the output from an actual xterm terminal window, and each line ends with a carriage return. In `shell-mode`, the carriage returns are not output.

[repl-typescripts.tgz](https://github.com/user-attachments/files/16848041/repl-typescripts.tgz)

feeley:
I had not noticed the double echo when running `emacs -q`.

[...]

PR #922, opened 2024-09-03
(Hopefully) fix the Emacs "double echo" problem

maconcg:
This is my attempt at a fix for issue #921.  With this patch applied, all the cases described in that issue work as they should (on my machine).  It doesn't rely on the use of Emacs' `comint-process-echoes`, so it avoids the issue described in [this comment](https://github.com/gambit/gambit/issues/921#issuecomment-2326336832) re: entering an expression whose value is then printed in the same format.

As far as I'm aware, most of the Emacs modes under which the sane might want to run a Gambit REPL are derived from `comint-mode`.  When you're running Gambit under a comint-derived mode, you really don't have any use for Gambit's own high-quality, Emacs-like line editing facilities.  If Gambit is run from a comint-derived mode, the value of the environment variable `INSIDE_EMACS` should end with `,comint`.

This approach is a bit more granular than the one currently in place, which checks whether `INSIDE_EMACS` is defined in the environment.  The added granularity ensures Gambit's own line editing facilities are available in non-comint modes (e.g., `term-mode`).

In [one of my comments on that issue](https://github.com/gambit/gambit/issues/921#issuecomment-2326518647), I mentioned that I hadn't been able to identify any behavior that depends on the value of `input_echo`.  I eventually did identify one case: if `input_echo` is 1, the "double echo" bug persists even when line editing mode is turned off.

This is my first substantial contribution in C, so while I think everything is above-board I also expect this will need some changes.  At the least, it can serve as a starting point.

Please share any feedback, especially if this patch breaks a setup that used to work.  And don't be gentle.


feeley:
Close, but not quite ideal. When TERM is set to something else than `dumb` I would expect some color highlighting but none happens. Here is a screenshot of the old and new behavior:
image
Note the setting of `TERM=xterm-256color` (which is my default `TERM` under emacs).

maconcg:
Do you see highlighting when you run the 4.9.5 release version? That's what I used as my point of reference for regressions, and on my machine there's no highlighting in comint-derived modes with that version.

I agree that it would be better to have some escape sequences.  Comint does a good job with colorization, bolding, and even directory tracking via OSC codes.  I'm thinking this functionality should be decoupled from line editing, which would mean some more complicated changes.  What do think?

maconcg:
Thought that would be a lot worse than it was.  I think this should permit only appropriate escape sequencences in a non-dumb comint mode.

Elsewhere in the file, `___getenv` and `strcmp` are used to check for "dumbness."  I thought about trying to update that code (because I added a `tty_is_dumb` function), but for now I'm holding off in the name of changing as few things as possible at a time.

feeley:
Your most recent patch causes output to be discarded when it is the same as the input, and colors are gone:
image

lib/os_tty.c
```diff
@@ -5311,8 +5317,18 @@ ___device_tty *self;)
   ___SCMOBJ default_options =
     ___INT(___device_tty_default_options_virt (&d->base));
 
-  if (___device_kind (&d->base.base) == ___TTY_DEVICE_KIND ||
-      d->stage == TTY_STAGE_NOT_OPENED)
+  if (comint_under_emacs ())
```
feeley, line 5320: I don't like the name `comint_under_emacs`... perhaps `lineeditor_under_emacs_comint` ? My reasoning is that the function checks for the situation where Gambit's line editor is running under emacs' comint, and the behavioural consequences of that on the line editor.

feeley (commented):
Just so you know, a lot of os_tty.c was written 25 years ago when I was designing the "new" (at the time) I/O system for Gambit, and every port had to be nonblocking to correctly implement the green threads. I now think it was an error to implement the line editor in C, in particular because it would be nice to have the same line editor features on ports other than TTYs, for example offering a REPL on a socket for remote debugging. It is far from the top of my TODO, but someday all of this should be rewritten in Scheme. It probably represents a few months of work to get to the same level of supported features.

lib/os_tty.c
maconcg: That'd be a fair name. I had initially changed that name's `lineeditor` part to `comint` because I was using that value to turn _off_ line editing mode, in which case the name `lineeditor_under_emacs` would have been misleading.

I kept the name with the latest commit because, conceptually, I think of "line editing" as a subset of "escape sequences" comprising actions like cursor movement, character deletion, etc.  The function's value is now intended to enable only the escape sequences not in my conceptual "line editing" set.  I also wanted to distinguish running Gambit from `term-mode` under Emacs as a different case.

From the perspective of the code as it exists right now, however, all features implemented via escape sequences, including stuff like colorization, are grouped under "line editor mode".  And so, from that perspective, `lineeditor_under_emacs_comint` would be the more appropriate name.  I think.

maconcg:
> > Your most recent patch causes output to be discarded when it is the same as the input,
> 
> It looks like you may have had `comint-process-echoes` set to a non-nil value in that screenshot. I can replicate this by evaluating `(setq-local comint-process-echos t)` in my shell-mode window. This variable's default value as of Emacs 29.4 should be `nil`, but it could maybe be different if you're running a different version of Emacs. Was this screenshot taken from an `emacs -q` session?
> 
> > and colors are gone:
> 
> This part is tricker. Haven't yet managed to replicate on my end.

Brought out the old Mac and gave it a shot -- results are the same as on OpenBSD.  Here's what I see:
Screenshot 2024-09-05 at 16 38 36

And here's the command line I used to start a super-default emacs:
`env -i $(command -v emacs) -Q -nw --eval "(setq-default explicit-shell-file-name \"/bin/bash\" explicit-bash-args '(\"--norc\" \"-i\"))"`

maconcg:
Realized the change I had made to `lineeditor_under_emacs_comint` had changed the function's behavior in an unintended way: before, when compiling with `USE_OLD_INSIDE_EMACS_DETECTION` defined, checking `INSIDE_EMACS` was a fallback.  My change meant `USE_OLD_INSIDE_EMACS_DETECTION` would cause `INSIDE_EMACS` to be skipped entirely.

I then asked myself, "how old would Emacs have to be for this to be useful?" and checked some old tarballs on [ftp.gnu.org](https://ftp.gnu.org/gnu/emacs/).  Looks like the most recent version for which the `INSIDE_EMACS` check would not have worked is 21.4, released in February 2005.  So I made the function simpler and easier to read by taking that logic out.

In lieu of asking permission, I cite [this comment](https://github.com/gambit/gambit/pull/710#issuecomment-882049585) and a willingness to revert upon request.

Negdayen:
Hmm, in st (suckless terminal), none of my input is echoed back, and I can't exit with Ctrl-D or Ctrl-C.

```
*0 jacob@void:gambit $ ./gsi/gsi
Gambit bc34d8ea

> 10
> 21
> *** ERROR IN (stdin)@1.13 -- Illegal character: #\x4
> *** ERROR IN (stdin)@1.14 -- Illegal character: #\x4
> *** ERROR IN (stdin)@1.15 -- Illegal character: #\escape
> *** ERROR IN (stdin)@1.18 -- Illegal character: #\x3
> *** ERROR IN (stdin)@1.19 -- Illegal character: #\x3
> *** ERROR IN (stdin)@1.20 -- Illegal character: #\escape
```

Negdayen:
It might have something to do with the fact that I use EXWM as my window manager, which means I have INSIDE_EMACS defined even in an xterm?

maconcg:
> It might have something to do with the fact that I use EXWM as my window manager, which means I have INSIDE_EMACS defined even in an xterm?

Are you starting the terminal emulator from a comint window?  I'm able to reproduce what you've shown that way.  It looks like the environment variable gets inherited by the terminal emulator (and then by Gambit, by way of the terminal emulator).  And because the terminal emulator sets the value of TERM (to, for example, `xterm`), things get wonky.

I don't currently use `exwm` (maybe I'll try someday), so I don't know much about how it works.  Is there a way you can  start the terminal emulator without going through comint (e.g., `M-!`/`shell-command` or `M-&`/`async-shell-command`)?

Also, things should work correctly when you run gsi from `term-mode`.  `term-mode` also sets `INSIDE_EMACS`, but because it sets `INSIDE_EMACS` to a value that doesn't end with `,comint` it should be ignored by Gambit.

Negdayen:
Yeah, I think I need to investigate more on the EXWM side of things, for some reason INSIDE_EMACS is set to "29.4,comint" as soon as I startup EXWM, which seems like unexpected behavior.

When I use cwm as my window manager, the only odd behavior I notice is comint-dechar-or-maybe-eof (Ctrl-D) produces that illegal char message.

maconcg:
I hadn't noticed the `comint-delchar-or-maybe-eof` problem, but that also happens on my end.  Many thanks for helping catch that.

Glad the `INSIDE_EMACS` thing (sort of) worked itself out.  Weird one.

feeley:
> Hmm, in st (suckless terminal) and xterm, none of my input is echoed back, and I can't exit with Ctrl-D or Ctrl-C.
> 
> ```
> *0 jacob@void:gambit $ ./gsi/gsi
> Gambit bc34d8ea
> 
> > 10
> > 21
> > *** ERROR IN (stdin)@1.13 -- Illegal character: #\x4
> > *** ERROR IN (stdin)@1.14 -- Illegal character: #\x4
> > *** ERROR IN (stdin)@1.15 -- Illegal character: #\escape
> > *** ERROR IN (stdin)@1.18 -- Illegal character: #\x3
> > *** ERROR IN (stdin)@1.19 -- Illegal character: #\x3
> > *** ERROR IN (stdin)@1.20 -- Illegal character: #\escape
> ```
> 
> I do see my input in `emacs -q` using `run-scheme` without duplicates, but I also get these same illegal character messages when trying to quit using control characters.

It seems your terminal is in raw mode (there is no line editing and buffering and characters pass through as is). This mode can be entered with `tty-mode-set!` like this:

```
$ gsi
Gambit v4.9.5-140-g37f022c0

> (tty-mode-set! (current-input-port) #f #f #t #f) ;; don't handle special chars, no echo, raw input, cooked output
> 12
> *** ERROR IN (stdin)@2.4-2.5 -- Illegal character: #\x4
> *** ERROR IN (stdin)@2.5-2.6 -- Illegal character: #\x3
```

So something has put the tty in that mode. In `os_tty.c` it could be these lines:

```
      d->input_allow_special = 0;
      d->input_echo = 0;
      d->input_raw = 1;
      d->output_raw = 1;
      d->speed = 0;
```

maconcg:
Setting `input_raw` to `0` fixes the `comint-delchar-or-maybe-eof` problem.  I'll test it some more and if I can't find anything else that breaks, I'll add a commit here that does that.

Another issue @Negdayen highlighted is that because we are using the values of two environment variables that can be set independently, `INSIDE_EMACS` and `TERM`, we have to trust that the former's value is correct.  It might be possible to address this with more logic (for example, checking that `INSIDE_EMACS` ends with `,comint` _and_ that `TERM` begins with `dumb`), but it also might be sufficient to say "don't do that" (which keep things simpler).

What do you think?

feeley:
I could be wrong, but I think `INSIDE_EMACS` and `TERM` provide completely independent information. `TERM` indicates terminal capabilities and in particular if escape sequences and colors are allowed. The case of `TERM=dumb` means that no tty capabilities are supported for that terminal (no escape sequence processing and no colors). `INSIDE_EMACS` indicates that the standard input is provided by emacs, and that "line editing" is taken care of by emacs (so Gambit should disable its builtin line editing). With the `,comint` suffix, it indicates that the standard output is also filtered by emacs' comint, which eats output lines that are equal to input lines to avoid double echo.

So it seems to me the setting of `TERM` should never affect echo. It is `INSIDE_EMACS` that should affect echo. If `,comint` is present then echo should be forced, because comint expects that, but otherwise echo should be disabled, because emacs' line editing is already echoing characters that are typed.

feeley:
I found this in the bash manual, so maybe they know something I don't:

INSIDE_EMACS
If Bash finds this variable in the environment when the shell starts, it assumes that the shell is running in an Emacs shell buffer and may disable line editing depending on the value of TERM.

lassik:
https://github.com/search?q=repo%3Abminor%2Fbash%20inside_emacs&type=code

feeley:
> https://github.com/search?q=repo%3Abminor%2Fbash%20inside_emacs&type=code

Thanks. I see this in the bash sources and it explains why bash is sensitive to the combination of `INSIDE_EMACS` and `TERM`:

```
      no_line_editing |= in_emacs && STREQ (term, "dumb");
```

However, I think this is an error because it is still possible to implement a basic form of line editing on a dumb terminal (as long as it correctly obeys "backspace" to go backward so that forward and backward arrows can be implemented and parenthesis balancing).

maconcg:
This may not be the final answer, and I'm still wrapping my head around why this works (other than "the debug log led me here").  I'm on my way to out-of-my-depth if I'm not already there.

Even so, this seems to fix EOF/`Ctrl-D` behavior without introducing any new bugs.  If I achieve clarity in the near future, I'll report back.

The only case I _know_ is broken is when you launch a REPL in a normal terminal emulator with `INSIDE_EMACS` set to something that ends with `,comint`.  Of course, this is a weird case in which Gambit is essentially being misled by an environment variable, so without checking `INSIDE_EMACS` and `TERM` in combination, there may not be a way to fix it?

maconcg:
> > https://github.com/search?q=repo%3Abminor%2Fbash%20inside_emacs&type=code
> 
> Thanks. I see this in the bash sources and it explains why bash is sensitive to the combination of `INSIDE_EMACS` and `TERM`:
> 
> ```
>       no_line_editing |= in_emacs && STREQ (term, "dumb");
> ```
> 
> However, I think this is an error because it is still possible to implement a basic form of line editing on a dumb terminal (as long as it correctly obeys "backspace" to go backward so that cursor movement can be implemented and parenthesis balancing).

I'd also like to say thanks for clearing up a head-scratcher I'd shrugged off.  I've been using `TERM=dumb-emacs-ansi` only on OpenBSD with `ksh` and have had a line in my `~/.kshrc` that executes `set +o emacs` when `INSIDE_EMACS` ends with `,comint`.  Echoing behavior was wonky when I started an interactive `bash` session from my `shell-mode` `ksh` session.

Turns out this is why -- bash isn't quite doing the right thing here.  Instead, it should examine the value of `INSIDE_EMACS` (like Gambit will soon do, knock on wood).

Negdayen:
One more difference in behavior:

In gambit master I can execute `comint-interrupt-subjob` to abort a long running computation and bring up a debug REPL.

```
Gambit v4.9.5-83-g7d462916

> (thread-sleep! 10)
(thread-sleep! 10)
  C-c C-c*** INTERRUPTED IN (stdin)@1.1-1.19
1> 
```

But with the latest patch it just kind of does nothing:

```
Gambit 01d3b9d1

> (thread-sleep! 10)
  C-c C-c  C-c C-c  C-c C-c  C-c C-c> 
```

If I first run `comint-interrupt-subjob` and then sleep the thread, I get this:
```
Gambit 01d3b9d1

>   C-c C-c (thread-sleep! 10)
*** ERROR IN (stdin)@1.1 -- Illegal character: #\x3
> 
```

maconcg:
This interrupt issue is pretty weird.  It looks like Gambit still responds to an INT signal as it should when I run `kill -s INT` on the process, at least.  I'll have to look into why that C-c C-c output pops up.  If I evaluate `M-x comint-interrupt-subjob` instead of using the keybinding, this bizarre output pops up:

```
$ gsi/gsi
Gambit v4.9.5-182-g01d3b9d1

>   M-x c o m i n t - i n t e r r u p t - s u b j o b RET
```

feeley:
When a TTY is the "controlling terminal" and the TTY is in the normal "canonical" mode, then ctrl-c and ctrl-z typed at the terminal respectively send SIGINT and SIGTSTP. But this is not something you want if the terminal is connected to a serial port and you want to transfer raw bytes over that tty without them causing the program to stop or be suspended. That is why there is the `d->input_raw` boolean and the call `tcsetattr (fd, TCSANOW, &new_termios)` to change the terminal mode. More info on the POSIX terminal interface is available here: https://en.wikipedia.org/wiki/POSIX_terminal_interface .

maconcg:
Looks like I have a nice Sunday of reading cut out for me.  Thanks for the pointer.

I'll also need to figure out the meaning of this bit (i.e., what bug?) from `comint.el`:

```
;; Fixme: Is this still relevant?
(defvar comint-ptyp t
  "Non-nil if communications via pty; false if by pipe.  Buffer local.
This is to work around a bug in Emacs process signaling.")
```

In addition to the `exec` example I shared above, setting `comint-ptyp` to `nil` makes things work in `geiser`, which is the main way I interact with the Gambit REPL.

Negdayen:
Setting Gambit's `struct ___device_tty_struct` `input_allow_special` field to true seems to resolve to the `comint-interrupt-subjob` issue for me, still poking around to understand things though.

```
Gambit 01d3b9d1

> (tty-mode-set! (current-input-port) #t #f #f #f #f)
> (thread-sleep! 10)
  C-c C-c*** INTERRUPTED IN (stdin)@2.1
1> 
```

maconcg:
> Setting Gambit's `struct ___device_tty_struct` `input_allow_special` field to true seems to resolve to the `comint-interrupt-subjob` issue for me, still poking around to understand things though.
> 
> ```
> Gambit 01d3b9d1
> 
> > (tty-mode-set! (current-input-port) #t #f #f #f #f)
> > (thread-sleep! 10)
>   C-c C-c*** INTERRUPTED IN (stdin)@2.1
> 1> 
> ```

That's very helpful, thanks.  This means we know, roughly, how to approach fixing this issue from either the Gambit side or the Emacs side.  I'd like to figure out which is the "right" way to do it, though.  Like, the way Emacs fixes this "bug" to set `comint-ptyp` to `t`? Or was `comint-ptyp` just made available so that you can set it to `nil` when you need to work around the "bug"?

Negdayen:
My current guesstimate is you can control whether Emacs launches the process with a pty or pipe by binding `process-connection-type` before comint executes `start-process` (through `start-file-process`), in which case you would probably also want to bind `comint-ptyp` to `nil`.

When we set Gambit's `input_allow_special` to true, it enables the `ISIG` terminal input flag (`man termios`).  It also enables a few other flags (`IXON`, `IEXTEN`).  Might need to check Windows + Emacs + Gambit integration, as this also touches a Windows console setting, although I think historically I have had troubles there, so maybe this will make that work better. :-)

maconcg:
Looks like:

- `IXON` gets turned off either way because `d->lineeditor_mode` is 1.
- `IEXTEN`'s behavior is implementation-defined, but it's on by default on the systems I've tested, meaning the effect of `d->input_allow_special = 1` is just to skip turning it off.

So I think setting `d->input_allow_special = 1` works with default Emacs settings and shouldn't introduce any other problems, at least on unix.  I don't understand what you're saying about Windows/the Windows console, however, and I don't have a machine that runs Windows.

The remaining known-broken case is again running a REPL with `INSIDE_EMACS` inaccurately set.  On one hand, it's a weird corner case that I think can be fixed only by (erroneously? unprincipled-ly?) checking `TERM`'s value.  On the other hand, it's a case that actually came up.  I'm open to direction.

Negdayen:
`input_allow_special` will set `ENABLE_PROCESSED_INPUT` on Windows, which looks similar in spirit to ISIG.  I will check that just to make sure there's no regressions there.

Personally I am fine with saying its the responsibility of Elisp applications spawning processes to set INSIDE_EMACS appropriately.

Negdayen:
Looks good to me now, tested with Emacs 29.

- I can see colors in `shell-mode`when I `export TERM=xterm-256color`.
- I don't see any duplicate input with the default `nil` value of `comint-process-echoes`.
- Default bindings of `C-c` and `C-d` work as expected.

There is an unfortunate behavior on Windows where `comint-delchar-or-maybe-eof` immediately terminates Gambit, but it looks like a pre-existing behavior not introduced by your work and seems to be a known problem with Emacs on Windows (https://www.gnu.org/software/emacs/manual/html_mono/efaq-w32.html#Subprocess-EOF).

maconcg:
That comint bug is annoying, but otherwise I'm glad to read things are working well.  Thanks a ton for your help, and if you run into anything down the road please let me know.

monnier:
I'm not super familiar with the comint side of Emacs but I remember when we had to change `INSIDE_EMACS`, it was pretty clear that it was not intended for such uses and the idea was that something like `gsi` should check `TERM` instead.  What were the reasons that made you use `INSIDE_EMACS` rather than `TERM` to try and detect the "comint-style" situation?

maconcg:
I'll try to answer your question, but I can't claim to have decided to use `INSIDE_EMACS` rather than `TERM`; that decision predates me.  I just decided to use the value of `INSIDE_EMACS` rather than to check whether it's set.  Also note that other projects, including Bash, use this environment variable for similar ends.

On a technical level, I don't think there's a reason `TERM` couldn't be used; if there were a widely-accepted, standardized `TERM` value that corresponds to `comint`, we could use that.  There isn't one (the closest I've found is `dumb-emacs-ansi`), but even if there were, you'd really need every individual program that uses the value of `TERM` to correctly interpret that value.  This is a big coordination problem. For example, currently, `git` produces colorized output for `dumb-emacs-ansi`, but GNU `ls --color=always` does not.

That leads us to the first reason: `comint` users typically end up picking a common `TERM` value that works for most of the programs they use (e.g., `xterm`, `ansi`, `vt100`, `xterm-256color`) and automatically setting that for all `shell-mode` sessions.

The second reason is related: With `comint`'s default `TERM` value, `dumb`, you want echoing turned off (otherwise you'll end up with an extra output line each time you press return).  But if you're in a "traditionally dumb" terminal, you need echoing turned on or else your typed input is invisible.  The same is true for any other common `TERM` value a `comint` user might choose.

I'd also like to ask if you could elaborate on what you mean when you say "it was pretty clear that it was not intended for such uses".  If it isn't intended for uses like this, what _is_ it intended for?

monnier:
`INSIDE_EMACS` is not overridden by other applications, so an `xterm` started from an Emacs subprocess will have that variable set, even tho programs running inside of it aren't running "inside Emacs" even less inside comint.
The reverse problem appears when ssh-ing into a host, where `INSIDE_EMACS` is usually not forwarded to the remote host, even though their input/output may indeed end going from/to comint.

Using `TERM` addresses these issues.  The standard way to detect "running inside comint" is to check if `TERM` is set to `dumb`.  The problem with ANSI color escapes is much more general than Gambit: Emacs has started supporting them without announcing it with a new `TERM` setting.  As you say, the use of colors in "batch" process output is a big mess currently, with different programs using different heuristics.

Maybe you want to `M-x report-emacs-bug` asking for Comint to use `dumb-emacs-ansi`.
Or maybe you can just as well assume that `dumb` terminals support ANSI color sequences (even tho it goes against the very notion of `dumb`)?  This said, the use of colors is somewhere separate from the double-echo problem.

Coming back to the double-echo, rather than comparing `TERM` against `dumb`, I believe The Right Thing™ would be to ask ncurses if the current terminal supports line-editing.  Does ncurses provide that info?  If not, maybe a bug report against ncurses is in order?
After all, the problem is not exclusive to comint, it probably affects also things like [rlwrap](https://github.com/hanslub42/rlwrap) or [ledit](https://github.com/chetmurthy/ledit) for example.

maconcg:
I realize this thread has become long, but this all might make more sense if you read through this PR's comments chronologically.  Not that you need to, just that it's somewhat interesting.  Feels like we've just about come full-circle.

I don't disagree with anything you've said, except for a couple tangential things.  One is that I'm not sure Gambit uses `ncures` (it definitely doesn't link against it).  Also, submitting a bug report for `comint` to use `dumb-emacs-ansi` wouldn't do anything; `comint` itself works either way, and `dumb` is probably the more portable default choice.

Aside from very brief test sessions, I don't run `gsi` from `shell-mode`.  And those "test sessions" are on the order of "start a REPL, make sure I can evaluate `(+ 1 1)` without a segfault.  I hope no one runs `gsi` this way for anything more extensive.

I'm happy to drop the use of `INSIDE_EMACS` entirely and just use `dumb`, as long as others are okay with that.  The easiest thing is definitely to just turn off line editing when `TERM=dumb` -- in fact, if you swap out that `TERM` test for a test against the value of `INSIDE_EMACS`, this is essentially what the PR's initial commit did.  Because of the way Gambit works internally, however, this means the REPL's welcome header won't be colorized in `shell-mode`.

monnier:
BTW, I just test `bash` and `gsi` under `rlwrap` and `ledit` and my conclusion seems to be that it's a big mess out there.
(FWIW, `gsi` seems to work well under `rlwrap` but not at all under `ledit`).

maconcg:
Ok, so my summary of the status of the PR at this point:

Given _the mess out there_ as it is today, we've covered our bases pretty well.  To cover those bases, we've done some arguably contrived things, including checking the value of `INSIDE_EMACS`.  (Not that either is The Right Thing, but I'd still argue that checking _the value of_ `INSIDE_EMACS` is less wrong the previous behavior of just checking whether it's set).

These contrivances can be explained, in part, by an attempt to accommodate use cases that may or may not matter to actual users -- for example, whether REPL's version header is colorized when started from `shell-mode`.

We could simplify things by ignoring the value of `INSIDE_EMACS` entirely.  This means that running `gsi` from a default-configuration `shell-mode` session will not work correctly unless `TERM=dumb` (or maybe `TERM=dumb-emacs-ansi`, if I have my way).  An advantage of this path would be that Gambit's behavior is more likely to remain the same if/when _the mess out there_ takes on a new form.

feeley:
I think testing `INSIDE_EMACS` is the right path, because the most common use of gsi (for me and others I know) is to start it from emacs, so the REPL under emacs must work for all (or most) ways to start gsi. I normally use M-x shell, then start gsi or M-x run-scheme. Term mode and geiser should work too. I don't really care about the edge case of starting gsi in an xterm started from a shell in emacs.

Perhaps the cleanest approach would be to convey the information through TERM, but that's not how emacs works right now. And even if some heroic effort made that happen in the next version of emacs, I would worry about all the users that can't replace the old emacs they have on an old machine.

maconcg:
I think I agree that testing `INSIDE_EMACS` is the right path, especially with confirmation that people do, in fact, run `gsi` from `shell-mode`.

Basically, I want to be able to recommend that <generic Emacs user> try Gambit without having to explain a bunch of caveats and/or workarounds needed to get their REPL working as well as <some other Scheme>.

feeley:
> Basically, I want to be able to recommend that <generic Emacs user> try Gambit without having to explain a bunch of caveats and/or workarounds needed to get their REPL working as well as <some other Scheme>.

I agree with that goal. For the simple case, it should just work. Otherwise new users will just go away when they can't even use the REPL to evaluate `42`.

feeley:
Perhaps we are just trying too hard. Maybe just implementing that when `INSIDE_EMACS` is defined then Gambit will disable its line editor completely and just use the TTY canonical mode. So the echo will happen and emacs can strip it away. And if you start gsi from an xterm started from a shell in emacs, then that gsi will also use TTY canonical mode, so at least the basic line editing of the TTY canonical mode will happen, which is what most other Schemes do anyway. There could be an exception for term mode in emacs, which would not disable the Gambit line editor.

maconcg:
One point: I don't think we can treat `TERM=dumb` as a synonym for "in a comint-derived mode".  This value is documented and supported by other programs, including [`xterm`](https://invisible-island.net/xterm/manpage/xterm.html).  We want the REPL to behave differently (with regards to echoing, specifically) in a `dumb` xterm session vs. a `dumb` `shell-mode` session.

Another point: In beginning to put together a matrix, I looked through the Emacs 29.4 sources for modes that may assign a value to `INSIDE_EMACS` and what that value might look like.  Below is a possibly-incomplete set of such cases, plus (perhaps not _the_, but at least _a_) default `TERM` values for each case:

|Emacs mode(s)     |      `INSIDE_EMACS` example|`TERM` default|
|------------------|----------------------------|--------------|
|`comint`          |               `29.4,comint`|        `dumb`|
|------------------|----------------------------|--------------|
|`compile`         |              `29.4,compile`|        `dumb`|
|------------------|----------------------------|--------------|
|`epg`             |                  `29.4,epg`|`xterm` (huh?)|
|------------------|----------------------------|--------------|
|`eshell`          |               `29.4,eshell`|        `dumb`|
|------------------|----------------------------|--------------|
|`term`            |            `29.4,term:0.96`| `eterm-color`|
|------------------|----------------------------|--------------|
|`tramp`           |                `29.4,tramp`|        `dumb`|
|------------------|----------------------------|--------------|
|`tramp` + `comint`|   `29.4,comint,tramp:2.6.3`|        `dumb`|
|------------------|----------------------------|--------------|
|`tramp` + `term`  |`29.4,term:0.96,tramp:2.6.3`| `eterm-color`|
|------------------|----------------------------|--------------|

This PR's patches currently don't account for `compile` `epg`, `eshell`, or `tramp`.  Neither `compile` or `epg` seems relevant, but `eshell` and `tramp` do.  This would add some complexity, but at least until the next Emacs release, the complexity should stop there.

Due to what I said above about `TERM=dumb`, we have to at least check for whether `INSIDE_EMACS` is set.  This means we must consider both `INSIDE_EMACS` and `TERM`.

If we only consider whether `INSIDE_EMACS` is set and the value of `TERM`, we have to pick which scenario we'd like to break:
- Disable line editing when `INSIDE_EMACS` is set and `$TERM != dumb` => break Gambit REPL in `shell-mode` when `TERM` has a non-dumb value
- Disable line editing when `INSIDE_EMACS` is set, regardless of `$TERM` => break Gambit REPL in `term-mode`

Because each of the two preceding scenarios represents a common use case, neither seems appealing.

Given that most of the necessary logic has already been implemented, I say we add accommodation for `INSIDE_EMACS` values set by `eshell` and `tramp`, then call it a day (week? fortnight?).

This will produce the right behavior for the maximum number of users.  This is solution is inelegant, sure, but so is untangling the cables behind your desktop PC.  I think it's the _most_ right we can currently hope for.

monnier:
> We could simplify things by ignoring the value of `INSIDE_EMACS` entirely.
> This means that running `gsi` from a default-configuration `shell-mode`
> session will not work correctly unless `TERM=dumb` (or maybe

Not sure I understand what you're saying here: `TERM=dumb` *is* what you
get by default in `M-x shell`.

maconcg:
> We could simplify things by ignoring the value of `INSIDE_EMACS` entirely. This means that running `gsi` from a default-configuration `shell-mode` session will not work correctly unless `TERM=dumb` (or maybe
> Not sure I understand what you're saying here: `TERM=dumb` *is* what you get by default in `M-x shell`.

Right -- I'm just saying this is a common case (due to reasons I described in a previous comment).  And if our goal is to have the REPL "just work", we should account for that common case.

feeley:
Do I understand correctly that this would mean that in shell-mode by default TERM will be dumb and thus there will be no colors but all interaction will work well, and that if I force in that shell TERM to dumb-emacs-ansi or xterm-256colors things will break?

maconcg:
> Do I understand correctly that this would mean that in shell-mode by default TERM will be dumb and thus there will be no colors but all interaction will work well, and that if I force in that shell TERM to dumb-emacs-ansi or xterm-256colors things will break?

Yes, I think.  To be doubly sure, let me rephrase that back to you:

By default, when you start a `shell-mode` shell, the value of `TERM` is set to `dumb`.  Under this default configuration, things would work.  But if you change the value of `TERM` to something else (so you can get colorized `ls` output, for example), the Gambit REPL would break.

feeley:
It is unfortunate that we lose colors. One of the recent additions to Gambit is the error highlighter that is much more communicative when colors are available:

image

maconcg:
Agreed.  To clarify again -- we're talking about a hypothetical situation in which we check only whether `INSIDE_EMACS` is set and `TERM == dumb`.

With the current iteration patch, you can run a REPL from `shell-mode` with whatever `TERM` value you like (other than "dumb"), and colorization (as well as everything else you'd expect) should work fine.

feeley:
Could you please add the information of what works and not to the matrix (echo and colors)?

maconcg:
Yes, I'm working on that now.  Requires some testing for each test to double-check :)

monnier:
> whatever `TERM` value you like

Note that in my book, setting `TERM` manually is either a pilot error or
just a (temporary?) workaround for an error in the program that provides the
"terminal emulator".

Which is why I think Comint *should* set `TERM` to something like
`dumb-emacs-ansi` rather than `dumb`: so as to announce that it's not
just a dumb terminal but one that supports ANSI color escape sequences.

feeley:
Yes I agree. My .profile contains

```
if [ "$INSIDE_EMACS" != "" ] ; then
    TERM=xterm-256color
fi
```

but that is when I didn't know about `dumb-emacs-ansi` which would be the correct thing to set `TERM` to. I just checked and it gives the same colorized errors.

maconcg:
> Note that in my book, setting `TERM` manually is either a pilot error or just a (temporary?) workaround for an error in the program that provides the "terminal emulator".

I agree -- it's a workaround, but it's a workaround that is needed for virtually all cases to have a good general-purpose `shell-mode` experience (with "batch mode" colorization, etc.).  But to fix the root cause would require the Herculean feat of coordinating how various programs handle the value of `TERM`.  With `TERM`, dishonestly is the term.  For example: everyone and their grandmother claims to be `xterm` (including Apple's Terminal.app), but they aren't `xterm`.  Changing the value of `TERM` for `shell-mode` is roughly as much an error as Apple claiming to be `xterm`.

monnier:
> ```
> if [ "$INSIDE_EMACS" != "" ] ; then
>     TERM=xterm-256color
> fi
> ```

A better fix is to `(setq comint-terminfo-terminal "...")` so that you
don't mistakenly change `TERM` when running a terminal emulator from
`M-x shell` and so that `TERM` is set properly even if your shell is
not involved.

maconcg:
In both my `.bashrc` and `.kshrc` files, I have this (which is needed to correct echo behavior for `dumb-emacs-ansi`):
```
if [[ "${INSIDE_EMACS:-}" == *,comint ]]; then
    set +o emacs
fi
```

And then in my `custom-set-variables` block of `init.el`, I set `comint-terminfo-terminal` to `"dumb-emacs-ansi"`.  You need both to get the right behavior.

maconcg:
Here's a table that summarizes relevant behaviors as of commit `bc6008c`:

|Mode    |  `INSIDE_EMACS`|           `TERM`|colors?|doubling?|interrupt shortcut?|EOF shortcut?|
|--------|----------------|-----------------|-------|---------|-------------------|-------------|
|`eshell`|   `29.4,eshell`|`dumb-emacs-ansi`|      Y|        Y|                  Y|          N/A|
|`eshell`|   `29.4,eshell`|           `dumb`|      N|        Y|                  Y|          N/A|
|`geiser`|   `29.4,comint`|`dumb-emacs-ansi`|      Y|        N|                  Y|          N/A|
|`geiser`|   `29.4,comint`|           `dumb`|      N|        N|                  Y|          N/A|
|`shell` |   `29.4,comint`|`dumb-emacs-ansi`|      Y|        N|                  Y|            Y|
|`shell` |   `29.4,comint`|           `dumb`|      N|        N|                  Y|            Y|
|`term`  |`29.4,term:0.96`|    `eterm-color`|      Y|        N|                  Y|            Y|
|`term`  |`29.4,term:0.96`|           `dumb`|      N|        N|                  Y|            Y|

Leaving off `tramp` from this because `comint` detection will currently fail.

Negdayen:
https://github.com/akermu/emacs-libvterm has become a popular alternative to `term-mode` and `ansi-term-mode`.  Interestingly, they set `INSIDE_EMACS=vterm` and default to `TERM=xterm-256color`.  Still catching up on the thread, but on commit abd850b there's colors, no doubling or missing input, eof, sigint as I would expect.

maconcg:
Ok, yeah, I put `N/A` for `geiser` and `eshell` because I thought they weren't bound by default.

However, as you've shown, in `eshell` the binding is just `C-c C-d`, and in `geiser` I'm able to produce the correct behavior by invoking `M-x comint-send-eof`.  So I'll mark the relevant `eshell` values with a `Y` and leave the `geiser` values as `N/A` only because there isn't an EOF keyboard shortcut in the default configuration.  But depending on how you look at it, the table should maybe have `Y`s for geiser's EOF shortcut.

maconcg:
As of `1a49ccf`, the early return in `lineeditor_output_terminal_emulate` just uses existing members of the `___device_tty` struct.  So we avoid the inefficient environment variable check, and you should be able to control echoing without any modifications to `tty-mode-set!`.

maconcg:
Okay, head is on a bit straighter now.  Today I found that running `gsi -:dc` exposed a leak in earlier versions of my echo dam.  With commit 5c2904d, I hope to have plugged these leaks at their source (`lineeditor_output_chars`).

feeley:
@maconcg Thank you for your contribution!

PR #923, opened 2024-09-04
Improve unit test behavior for "Unimplemented operation" os-exception

maconcg:
This is a follow-up to #918.  It reuses `exit0-when-unimplemented-operation-os-exception` (currently used for network unit tests) for the filesystem unit tests.  It reflects an understanding that "the OS does not implement this operation" an acceptable answer for these tests.

Something I don't love about this is that for a given test in this file, "unimplemented" will be treated as an acceptable answer on platforms where we'd normally expect the tested feature to be implemented (e.g., `executable-path` on Linux).

In the future, it might make sense to adopt elements of either of the following approaches:

1. Distinguish between features that are mandatory on all platforms (e.g., the `file-exists?` procedure, which is explicitly mentioned in R7RS) from features that may or may not be supported on a given platform (e.g., the `executable-path` procedure).

2. Choose `capture-behavior` based on platform.  For example: `capture-behavior` is re-bound only on platforms where certain features are known not to be implemented.

I'm partial to the first approach.  The second would be a move toward Python-style "support tiers", which I don't like and wouldn't really comport with Gambit's philosophy as I understand it.

### DIFF
--- a/tests/unit-tests/#.scm
+++ b/tests/unit-tests/#.scm
@@ -202,12 +202,16 @@
 (##define-syntax check-exn        (lambda (src) (##expand-check src)))
 (##define-syntax check-tail-exn   (lambda (src) (##expand-check src)))
 
+(define (os-unimplemented-exception-string e)
+  (and (os-exception? e)
+       (let ((err-code (##os-err-code->string (os-exception-code e))))
+         (and (string=? err-code "Unimplemented operation")
+              err-code))))
+
 (define (exit0-when-unimplemented-operation-os-exception thunk)
   (with-exception-catcher
    (lambda (e)
-     (if (and (os-exception? e)
-              (equal? (##os-err-code->string (os-exception-code e))
-                      "Unimplemented operation"))
+     (if (os-unimplemented-exception-string e)
          (exit 0)
          (raise e)))
    thunk))
--- a/tests/unit-tests/#.scm
+++ b/tests/unit-tests/#.scm
@@ -204,9 +204,9 @@
 
 (define (os-unimplemented-exception-string e)
   (and (os-exception? e)
-       (let ((err-code (##os-err-code->string (os-exception-code e))))
-         (and (string=? err-code "Unimplemented operation")
-              err-code))))
+       (let ((err-string (##os-err-code->string (os-exception-code e))))
+         (and (string=? err-string "Unimplemented operation")
+              err-string))))
 
 (define (exit0-when-unimplemented-operation-os-exception thunk)
   (with-exception-catcher
--- a/tests/unit-tests/13-modules/#.scm
+++ b/tests/unit-tests/13-modules/#.scm
@@ -19,7 +19,7 @@
 
 (define (capture-behavior thunk)
   (with-exception-catcher
-   (lambda (e) e)
+   (lambda (e) (or (os-unimplemented-exception-string e) e))
    thunk))
 
 (define (compare-behavior namespaces calls . behaviors)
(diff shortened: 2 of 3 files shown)


feeley:
I agree that the unit tests should distinguish procedures that must be implemented (and work on all platforms) from procedures that are in a sense optional. The classification could be more fine grained by defining requirements for specific "features", for example for "UDP networking" the following group of procedures must all be implemented: X, Y, Z...

However, I don't understand how your patch helps in any way with the issue.

Also be careful with the meaning of "unimplemented". There are two meanings in the Gambit source code:

1) A primitive that relies on a function that is not provided by the OS (or that hasn't been implemented in the runtime library)... such as `##executable-path` on OpenBSD.

2) A procedure exported by the Gambit runtime library that has no corresponding `##` (primitive) procedure, which typically has no type checking and error checking. So `car`, `list`, `vector-length` all have primitive implementations `##car`, `##list`, `##vector-length`, etc. But `pp`, `write-string`, and `vector-map` don't have a primitive version, so we can say the primitive variant is "unimplemented".

This second meaning is important in the context of the `(define-procedure ...` form, because the body of the definition is implicitly in the scope of a `(include "~~lib/gambit/prim/prim#.scm")` which causes calls to standard names to be replaced by the primitive variant. For example the code:

```
(define-procedure (foo (vect vector))
  (list (vector-length vect)
        (vector-map square vect)))
```

is equivalent (i.e. it expands into):

```
(define foo
  (lambda (%vect)
    (if (##vector? %vect)
        (##list
         (##vector-length %vect)
         (##unimplemented#vector-map ##square %vect))
        (##fail-check-vector '(1 . vect) foo %vect))))
```

Notice the call to `##unimplemented#vector-map` which will lead to a warning when compiling the code (and a run time error if the code is executed). To actually call the non-primitive `vector-map`, the macro `(standard <name>)` can be used:

```
(define-procedure (foo (vect vector))
  (list (vector-length vect)
        ((standard vector-map) square vect)))
```

It is done this way so that fast libraries can be written (fast because they only check for errors on entry to the procedure by checking that the parameters are of the correct type, expected range, etc). It also avoids having `##` all over the place, which is an eyesore.

maconcg:
I'm gonna think on this some more.  In the meantime, here's some context that I should have included about why I wrote this:

As a result of the previous PR, evaluating `(executable-path)` on OpenBSD raises an os-exception that I referred to as "unimplemented" because `(##os-err-code->string (os-exception-code <the exception in question>))` returns "Unimplemented exception". Since each evaluation produces a unique exception object, `check-same-behavior` was showing a failure for the `(executable-path)` test.

I first tried wrapping the `(executable-path)` call in prim_filesystem.scm inside `with-exception-catcher`/`with-exception-handler` to make the comparison between the exception objects' respective error codes (as opposed to comparing the exceptions themselves), but that resulted in the test suite producing a failure message about executable-path not having a test.

I then came across `exit0-when-unimplemented-operation-os-exception`, which for the purpose of the unit test produces the "same" result (i.e., `(exit 0)`) for `(executable-path)`.  I (questionably) reasoned that kind of "sameness" is about as good as the sameness that would result from comparing exception codes, and that by reusing this already-existing procedure I could avoid adding that code.

The only test producing such an "unimplemented" exception in `prim_filesystem.c` was the call to `(executable-path)`.  And wrapping _just_ the `(executable-path)` call in `exit0-when-unimplemented-operation-os-exception` produced that "there's no test for executable-path!" message, so I wrapped the whole thing in `exit0-when-unimplemented-operation-os-exception`.

So, yeah...gonna sleep on this one.

maconcg:
With a forehead-smack and a "D'oh!", I reverted the first commit and instead took an approach that effectively just results in a comparison between strings (and therefore a true result) instead of a comparison between two distinct "unimplemented operation" os-exception objects.

PR #924, opened 2024-09-04
configure.ac: Check for backtrace_symbols_fd in libexecinfo

maconcg:
I noticed that on my machine, the `configure` script wasn't finding `backtrace_symbols_fd`.  Looks like that's just because it's split out into a separate library on OpenBSD.  This PR should allow this check to succeed on OpenBSD and have no effect on other systems.

I believe I've followed the relevant conventions by including the change to `configure.ac` as well as the resulting code (generated by autoconf 2.69) in `configure`.  But if not, I'll fix it.

### DIFF
--- a/configure
+++ b/configure
@@ -8400,6 +8400,63 @@ if test "$ac_res" != no; then :
 fi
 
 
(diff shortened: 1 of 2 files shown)


feeley:
Thanks for the PR!

feeley:
Thanks!

PR #926, opened 2024-09-07
Add support for loongarch

wuruilong01:
Add support for loongarch


feeley:
Only the file `include/gambit.h.in` should change. Please change your PR to only patch this file. If other files need to change please explain why.

wuruilong01:
The config.guess config.sub file is too old and needs to be updated to support new architectures, e.g. riscv, loongarch.

feeley:
Understood. Thanks for your first contribution to Gambit!

PR #927, opened 2024-09-10
document executable-path under "Host environment"

maconcg:
Starting with this one per #920.

I had a think on the ways I had misunderstood this procedure when posting an [earlier comment](https://github.com/gambit/gambit/pull/918#issuecomment-2322895044), then tried to document it in a way that would have helped me avoid that.

I added its description adjacent to `command-line` because that contrast helped with my understanding.

I also tried to use an example that illustrates the point made in [this comment](https://github.com/gambit/gambit/pull/918#discussion_r1740264242). Hopefully people aren't thrown off by the `$_` -- it's pretty universal nowadays.

https://github.com/gambit/gambit/pull/918#issuecomment-2322895044

### DIFF
--- a/doc/gambit.txi
+++ b/doc/gambit.txi
@@ -5459,9 +5459,6 @@ The procedures in this section are not yet documented.
 @deffn procedure get-environment-variables
 @end deffn
 
-@deffn procedure executable-path
-@end deffn
-
 @deffn procedure command-name
 @deffnx procedure command-args
 @end deffn
@@ -12905,7 +12902,7 @@ operating systems supported by Gambit (e.g. MSDOS).
 * Filesystem operations::                   Filesystem operations
 * Shell command execution::                 Shell command execution
 * Process termination::                     Process termination
-* Command line arguments::                  Command line arguments
+* Invocation information::                  Invocation information
 * Environment variables::                   Environment variables
 * Measuring time::                          Measuring time
 * File information::                        File information
@@ -13378,7 +13375,7 @@ For example under Windows:
 
 @end deffn
 
-@node Process termination, Command line arguments, Shell command execution, Host environment
+@node Process termination, Invocation information, Shell command execution, Host environment
 @section Process termination
 
 @deffn procedure exit @r{[}@var{status}@r{]}
@@ -13401,8 +13398,8 @@ $ @b{echo $?}
 
 @end deffn
 
-@node Command line arguments, Environment variables, Process termination, Host environment
-@section Command line arguments
+@node Invocation information
+@section Invocation information
 
 @deffn procedure command-line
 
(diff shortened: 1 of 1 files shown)


doc/gambit.txi
```diff
@@ -13426,7 +13423,26 @@ $ @b{./foo 1 2 "3 4"}
 
 @end deffn
 
-@node Environment variables, Measuring time, Command line arguments, Host environment
+@deffn procedure executable-path
+
+This procedure returns a newly allocated string representing a
```
feeley, line 13428: I think we should avoid saying that the returned string is newly allocated, because I can imagine a different implementation (possibly a "better" implementation in terms of memory usage) where the string is not newly allocated.

doc/gambit.txi
```diff
@@ -13426,7 +13423,26 @@ $ @b{./foo 1 2 "3 4"}
...
+@dfn{normalized path} to the executable file from which the running process
+spawned.  The path is determined dynamically using a mechanism specific
+to the host operating system.
+
+For example under UNIX:
+
+@smallexample
+$ @b{echo '(pp (executable-path))' > ep.scm}
+$ @b{gsc -exe ep.scm}
+$ @b{mv ep $(mktemp -d)}
```
feeley, line 13438: If possible, examples should be executable so that they can be used as unit tests. So it would be good to write this example so that its execution is deterministic. Unfortunately using `mktemp`, and `/tmp` in general, is probably not deterministic. One way would be to reveal the CWD with a `pwd`, and the unit testing framework could enforce that the CWD for examples is always the same, say `/gambit/example`. This would give:

```
$ pwd
/gambit/example
$ ./ep
/gambit/example/ep
```

maconcg: Makes sense, thank you for clearing that up. Had gone out of my way there to _not_ use a determinisitc example.

maconcg:
Something that just occurred to me, which may or may not ultimately matter, is that `/tmp` is one of the few (the only?) predictably word-writable paths on a generic unix-like system.  So we could potentially avoid the need to manage $PWD via the unit testing framework if we instead used `/tmp/<something>/` for non-deterministic examples.

I'm thinking more about a user running `make check` than, say, the automated GitHub actions.

feeley:
Yes, but one has to be careful that the normalized path doesn't show up in the results:

```
$ gsi
Gambit v4.9.5-140-g37f022c0

> (with-output-to-file "/tmp/test.scm" (lambda () (pretty-print '(pp (/ 1 0)))))
> (load "/tmp/test.scm")
*** ERROR IN "/private/tmp/test.scm"@1.5-1.12 -- Divide by zero
(/ 1 0)
 ┏━━━━━━━━━━━━━━━━━━━━━━━━━━━ /private/tmp/test.scm ━━━━━━━━━━━━━━━━━━━━━━━━━━━
1┃(pp (/ 1 0))
1> 
```

Because that would not be portable between operating systems.

maconcg:
Ah, good catch.  I missed the macOS case.  On the Linux/BSD platforms I looked at, you might have `/var/tmp` as a symlink to `/tmp`, but `/tmp` would _normally_ be the normalized path.  Just an idea.  Might be better to avoid having different tests for different operating systems.

feeley:
Thanks!

PR #928, opened 2024-09-11
Correct use of script line to determine program behavior

maconcg:
I noticed I wasn't able to affect Gambit's case sensitivity using the executable name and spent some time looking into why.  Runtime options (e.g., `-:r5rs`) and the `#!fold-case` directive worked as expected.

After a pleasant safari through `_eval.scm` and `_io.scm`, I managed to track the problem back to the point where `##readtable-setup-for-language!` calls `##readtable-setup-for-standard-level!` (importantly, that point is at the end of `##readtable-setup-for-language!`).

The behavior of `##readtable-setup-for-standard-level!` depends on the value of `(##get-standard-level)`, which evaluates to `0` when Gambit mode has been selected explicitly and defaults to `-1`.

`##readtable-setup-for-standard-level!` interpreted these values in the same way.  By distinguishing "explicit" Gambit mode from "implicit" Gambit mode, we get what I believe is the intended behavior: a command-line runtime argument will override behavior specified by the executable name, but in the absence of such an argument the executable name will determine relevant behaviors.

Before:
```
$ cat ex0.scm
#!/usr/local/Gambit/bin/scheme-r5rs
(pp (eq? 'sym 'SYM))
$ ./ex0.scm
#f
$ cat exA.scm
#!/usr/local/Gambit/bin/scheme-r5rs -:r5rs
(pp (eq? 'sym 'SYM))
$ ./exA.scm
#t
```

After:
```
$ cat ex0.scm
#!/usr/local/Gambit/bin/scheme-r5rs
(pp (eq? 'sym 'SYM))
$ ./ex0.scm
#t
$ cat exB.scm
#!/usr/local/Gambit/bin/scheme-r5rs -:r7rs
(pp (eq? 'sym 'SYM))
$ ./exB.scm
#f
```

### DIFF
--- a/lib/_io.scm
+++ b/lib/_io.scm
@@ -10570,19 +10570,20 @@
 (define-prim (##readtable-setup-for-standard-level! rt)
   (let ((standard-level (##get-standard-level)))
 
-    (cond ((or (##fx= standard-level 4)  ;; R4RS language
+    (cond ((##fxzero? standard-level) ;; Gambit language, explicit
+           (macro-readtable-case-conversion?-set! rt #f)
+           (macro-readtable-keywords-allowed?-set! rt #t)
+           (macro-readtable-bracket-handler-set! rt '|[...]|)
+           (macro-readtable-brace-handler-set! rt '|{...}|))
+          ((or (##fx= standard-level 4)  ;; R4RS language
                (##fx= standard-level 5)) ;; R5RS language
            (macro-readtable-case-conversion?-set! rt #t)
            (macro-readtable-keywords-allowed?-set! rt #f))
-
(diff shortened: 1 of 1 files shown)


lib/_io.scm
```diff
@@ -10570,19 +10570,20 @@
 (define-prim (##readtable-setup-for-standard-level! rt)
   (let ((standard-level (##get-standard-level)))
 
-    (cond ((or (##fx= standard-level 4)  ;; R4RS language
+    (cond ((##fxzero? standard-level) ;; Gambit language, explicit
```
feeley, line 10573: Should keep the same style `(##fx= standard-level 0)`.

maconcg: Noted and fixed.

PR #969, opened 2025-08-04
Miscellaneous improvements in configure/makefiles

abdelq:
This comes from an effort to package 4.9.7 for Debian/Ubuntu. Most of this focuses on cleaning up after a build with small change for pkg-config and other for reproducibility of documentation by setting a seed.


abdelq:
Not fixed here, but `stat -s` used in `doc/makefile` isn't valid on Linux

feeley:
Thanks Abdel! Sorry for taking so long to review your PR... it slipped under the radar.

PR #974, opened 2025-08-28
Allow strings and u8vectors in open-process's arguments setting.

Negdayen:
Hi Marc,

Thanks for answering my question on gitter (duplicating the conversation here for posterity's sake).

```
Negdayen (Negdayen)
Is there a way to open processes that contain Unicode characters in their arguments? Toy example:

$ ./test
*** ERROR IN "test"@3.1-3.48 -- (Argument 1) Can't convert to C char-string
(open-process '(path: "echo" arguments: ("🐰")))

On Unix Gambit's open-process is implemented with execvp which does not specify the
encoding of the strings it receives.  That is because the data is just passed to the
process as a sequence of bytes (char in C) with no specific interpretation.  So one
workaround is to perform the specific encoding you want, presumably UTF8:

Most programs started with open-process probably expect the command line arguments to be
strings encoded with UTF8, but some probably don't, so there needs to be a way to pass raw
bytes when needed. So perhaps the list passed as arguments: should allow both strings
(which will be UTF8 encoded) and u8vectors that are passed as-is. I'll accept PRs for this
feature.
```

Does this meet the requirements?

### DIFF
--- a/doc/gambit.txi
+++ b/doc/gambit.txi
@@ -15864,9 +15864,9 @@ This setting indicates the location of the executable program in the
 filesystem.  There is no default value for this setting.
 
 @item
-@code{arguments:} @var{list-of-strings}
+@code{arguments:} @var{list-of-strings-or-u8vectors}
 
-This setting indicates the string arguments that are passed to the
+This setting indicates the string or u8vector arguments that are passed to the
 program.  The default value of this setting is the empty list (i.e. no
 arguments).
 
--- a/lib/_io.scm
+++ b/lib/_io.scm
@@ -7124,11 +7124,31 @@
   (define (fail)
     (##fail-check-string-or-settings 1 prim path-or-settings arg2))
 
+  (define (utf8 string)
+    (u8vector->string (##string->utf8 string)))
+
+  (define (u8vector->string u8vect)
+    (##list->string (##map ##integer->char (##u8vector->list u8vect))))
+
   (##make-process-psettings
    direction
    (if (##string? path-or-settings)
        (##list 'path: path-or-settings)
-       path-or-settings)
+       (let normalize-settings ((settings path-or-settings))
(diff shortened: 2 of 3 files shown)


feeley:
Thank you for your quick response aiming to solve the issue with the encoding of the arguments of `open-process`.

As you can see from the failing CI your solution is Unix specific (it fails on Windows) because it relies on the `execvp` function using the `char*` type (sequence of bytes) that are passed to the process unchanged. On Windows `open-process` is implemented with `CreateProcess` where strings are encoded using UCS2, so the forced UTF8 conversion done by `open-process` will pass garbage to `CreateProcess`.

The Scheme-level implementation of `open-process` should probably propagate the string or u8vectors all the way to the low-level `___device_stream_setup_from_process` function (in `lib/os_io.c`). This function can then decide how to encode strings (UTF8 on Unix, and UCS2 string on Windows), or pass the u8vector unchanged.

But this does not feel totally right because for Windows a u8vector does not make sense... it should be a u16vector. But that means the API of `open-process` depends on the OS, something I want to avoid.

So maybe a better approach is to add a `argument-char-encoding:` option in the settings of `open-process`, and have a default that is OS specific (UTF8 on Unix, UCS2 on Windows). And then the `___device_stream_setup_from_process` function can do the appropriate encoding. There is also a need for a `environment-char-encoding:` option to know how to encode the process' environment.

While writing this response I have noticed that, strangely, the environment is encoded with UTF8 on Unix whereas the arguments are encoded natively (i.e. `char*`) which means one byte per character. See the definition of `___ENVIRON_CE_SELECT` in `lib/os_shell.h`. This seems inconsistent.

So perhaps the simplest temporary solution is to change `___STREAM_OPEN_PROCESS_CE_SELECT` to use UTF8 encoding. Then we can work on supporting an option to choose other char encodings (the `arguments-char-encoding:` and `environment-char-encoding:` options).

PR #998, opened 2026-05-07
implemented srfi 95

chouettevan:
This PR implements [SRFI 95](https://srfi.schemers.org/srfi-95/srfi-95.html). Sorting and Merging. Merge sort is used for lists,
and Heapsort is used to sort vectors and string in-place.


lib/_std#.scm
```diff
@@ -883,6 +929,28 @@
...
+                                   (,name ,name)
+                                   (start (index-range-incl
+                                            0
+                                            (,prim-vect-length ,name))
+                                          0)
+                                   (end (index-range-incl
+                                          start
+                                          (,prim-vect-length ,name))
+                                        (,prim-vect-length ,name)))
+                (,prim-vect-sort! less? ,name))
```
feeley: The "start" and "end" parameters are missing on the call to `prim-vect-sort!`. There should be unit tests to catch this error.

lib/_std#.scm
```diff
@@ -883,6 +929,28 @@
...
+              ) (write 1))
```
feeley: Remove the stray code `(write 1)`. Shouldn't there be an empty list here for the false case of `(eq? name 'vector)`?

lib/_std#.scm
```diff
@@ -883,6 +929,28 @@
...
                  (,prim-vect-concatenate ,vect (macro-deleted-obj)))))
 
+
+
+        ,@(if (eq? name 'vector)
+             `((define-prim (,prim-vect-sort! less? ,name
+                                              #!optional
+                                              (start 0)
+                                              (end (,prim-vect-length ,name)))
+                            (macro-vect-mergesort! less? ,name start end ,prim-make-vect ,prim-vect-ref ,prim-vect-set!))
```
feeley: Should be indented under `define-prim` with 2 spaces.

The call to `macro-vect-mergesort!` should be protected by an `(if (fx< start end) ...)` so that the primitive can be called with end < start and not generating a segfault.

The vector should be returned from the primitive.

lib/_std#.scm
```diff
@@ -883,6 +929,28 @@
                (define-prim&proc (,vect-append ,vect ...)
                  (,prim-vect-concatenate ,vect (macro-deleted-obj)))))
 
+
+
+        ,@(if (eq? name 'vector)
+             `((define-prim (,prim-vect-sort! less? ,name
```
feeley: Avoid putting 2 parameters on the same line when the others each have a separate line.

lib/_std#.scm
```diff
@@ -383,6 +424,7 @@
     (define vect-set!-fixnum       (sym name '-set!-fixnum))
     (define vect-set               (sym name '-set))
     (define vect-set-small         (sym name '-set-small))
+    (define vect-sort!              (sym name '-sort!))
```
feeley, line 427: Remember that this will add the `##vector-sort!` and `vector-sort!` procedures to the Gambit runtime system. Because the documentation (file `doc/gambit.txi`) has not been extended to document this procedure it will fail a `make check` that verifies that all Gambit procedures are documented.

So before submitting a PR, make sure that `make check` succeeds.

The changes to `_std#.scm` should be in a PR that does just that.

lib/_std#.scm
```diff
@@ -510,6 +552,8 @@
 
        (define-primitive (,vect-ref ,name k))
 
+
```
feeley, line 555: Why the extra space?

lib/_std#.scm
```diff
@@ -526,6 +570,8 @@
 
        (define-primitive (,vect-set! ,name k ,elem-name))
 
+
```
feeley, line 572: Remove extra space.

lib/_std#.scm
```diff
@@ -883,6 +929,28 @@
                (define-prim&proc (,vect-append ,vect ...)
                  (,prim-vect-concatenate ,vect (macro-deleted-obj)))))
 
+
```
feeley, line 930: Remove extra space.

gambiteer:
Marc told me that you're working with him this summer, so I'll let him comment on the code.  I do have some general recommendations, though.

First is to mimic, to a good extent, the style of existing Gambit code; I'd recommend the bitwise operations beginning at https://github.com/gambit/gambit/blob/dbdd3be454e483121e03922d9d8ec275aab9db95/lib/_num.scm#L6411

To get correct indentation I'd use emacs and load the file `misc/gambit.el` that comes with the Gambit distribution.  I'm not an emacs expert, I use only a few commands, but I do use the autoindent and reformatting services of emacs.

I may have other comments later.

Brad

PR #999, opened 2026-05-14
added vector-sort! primitive and builtin procedure

chouettevan:
This PR adds the `vector-sort!` builtin procedure, the `##vector-sort!` primitive and a few unit tests for them

### DIFF
--- a/doc/gambit.txi
+++ b/doc/gambit.txi
@@ -5727,6 +5727,7 @@ The procedures in this section are not yet documented.
 
 @deffn procedure list-sort @var{proc} @var{list}
 @deffnx procedure list-sort! @var{proc} @var{list}
+@deffnx procedure vector-sort! @var{proc} @var{vect} @var{start} @var{end}
 @end deffn
 
 @deffn procedure any @var{pred} @var{clist1} @var{clist2} @dots{}
--- a/doc/gambit.txi
+++ b/doc/gambit.txi
@@ -5728,6 +5728,7 @@ The procedures in this section are not yet documented.
 @deffn procedure list-sort @var{proc} @var{list}
 @deffnx procedure list-sort! @var{proc} @var{list}
 @deffnx procedure vector-sort! @var{proc} @var{vect} @var{start} @var{end}
+@deffnx procedure vector-sort @var{proc} @var{vect} @var{start} @var{end}
 @end deffn
 
 @deffn procedure any @var{pred} @var{clist1} @var{clist2} @dots{}
--- a/doc/gambit.txi
+++ b/doc/gambit.txi
@@ -5727,7 +5727,9 @@ The procedures in this section are not yet documented.
 
 @deffn procedure list-sort @var{proc} @var{list}
 @deffnx procedure list-sort! @var{proc} @var{list}
-@deffnx procedure vector-sort! @var{proc} @var{vect} @var{start} @var{end}
+@end deffn
+
+@deffn procedure vector-sort! @var{proc} @var{vect} @var{start} @var{end}
 @deffnx procedure vector-sort @var{proc} @var{vect} @var{start} @var{end}
 @end deffn
 
--- a/lib/_std#.scm
+++ b/lib/_std#.scm
@@ -935,10 +935,8 @@
                    #!optional
                    (start 0)
                    (end (,prim-vect-length ,name)))
-                 (cond
-                   ((fx<= end start) ,name)
-                 (else (macro-vect-mergesort! less? ,name start end ,prim-make-vect ,prim-vect-ref ,prim-vect-set!)
-                 ,name)))
+                 (macro-vect-mergesort! less? ,name start end ,prim-make-vect ,prim-vect-ref ,prim-vect-set!)
+                 ,name)
                (define-procedure 
                  (,vect-sort!
                    (less? procedure)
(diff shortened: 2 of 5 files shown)


lib/_std#.scm
```diff
@@ -883,6 +927,31 @@
...
+                   (start (index-range-incl
+                            0
+                            (,prim-vect-length ,name))
+                          0)
+                   (end (index-range-incl
+                          start
+                          (,prim-vect-length ,name))
+                        (,prim-vect-length ,name)))
+                 (,prim-vect-sort! less? ,name start end))
+              ) '())
```
feeley: the part ` '())` should be on the next line

lib/_std#.scm
```diff
@@ -510,6 +552,7 @@
 
        (define-primitive (,vect-ref ,name k))
 
+
```
feeley, line 555: remove this empty line

lib/_std#.scm
```diff
@@ -526,6 +569,7 @@
 
        (define-primitive (,vect-set! ,name k ,elem-name))
 
+
```
feeley, line 572: remove this empty line

lib/_std#.scm
```diff
@@ -883,6 +927,31 @@
                (define-prim&proc (,vect-append ,vect ...)
                  (,prim-vect-concatenate ,vect (macro-deleted-obj)))))
 
+
```
feeley: remove this empty line

lib/gambit#.scm
```diff
@@ -506,6 +506,7 @@ list->u8vector
 list-set
 list-sort
 list-sort!
+vector-sort!
```
feeley: the names should be in alphabetical order

tests/unit-tests/07-vector/sort.scm
```diff
@@ -0,0 +1,13 @@
...
+(vector-sort! < a)
+(check-equal? a #(1 2 3 4 5))
+
+(define a #(1 3 6 4 2 5))
+(vector-sort! < a 1 4)
+(check-equal? a #(1 3 4 6 2 5))
+
+(define a #(21 33 65 54 2 5))
+(vector-sort! < a 1 4)
+(check-equal? a #(21 33 54 65 2 5))
```
feeley, line 16: These tests are rather minimal. Tests should be added for:
- `>`
- `(lambda (x y) (< (car x) (car y)))`
- verification that sort is stable (i.e. equal keys stay in the same ordering as the initial vector)

doc/gambit.txi
```diff
@@ -5727,6 +5727,7 @@ The procedures in this section are not yet documented.
 
 @deffn procedure list-sort @var{proc} @var{list}
 @deffnx procedure list-sort! @var{proc} @var{list}
+@deffnx procedure vector-sort! @var{proc} @var{vect} @var{start} @var{end}
```
feeley: The parameters `start` and `end` are optional and should be marked that way.

Also, a `vector-sort` procedure would be needed to be symmetrical with `list-sort`/`list-sort!`. It can be defined roughly as

```
(define (vector-sort less? vector) (vector-sort! less? (vector-copy vector))
```

but with the `start` and `end` parameters added, and the type checks of a `define-procedure`.

The documentation should have a separate `@deffn` for the list and vector sorts.

lib/gambit/gambit.sld
```diff
@@ -858,6 +858,7 @@ list->u8vector
 list-set
 list-sort
 list-sort!
+vector-sort!
```
feeley: Move to correct alphabetical order.

lib/_std#.scm
```diff
@@ -313,6 +311,48 @@
...
+       (let ((n (fx- hi lo)))
+         (if (fx< 1 n)
+             (let ((mid (fx+ lo (fxarithmetic-shift-right n 1))))
+               (sort! lo mid)
+               (sort! mid hi)
+               (merge! lo mid hi)))))
+
+     (sort! ,start ,end)))
+
+
```
feeley: Remove this empty line.

lib/_std#.scm
```diff
@@ -883,6 +927,31 @@
...
+        ,@(if (eq? name 'vector)
+             `((define-prim 
+                 (,prim-vect-sort!
+                   less? 
+                   ,name
+                   #!optional
+                   (start 0)
+                   (end (,prim-vect-length ,name)))
+                 (macro-vect-mergesort! less? ,name start end ,prim-make-vect ,prim-vect-ref ,prim-vect-set!))
```
feeley: There should be an additional test `(if (fx< start end) ...)` to avoid segfaults in that case.

Also the primitive should return the vector that is being sorted.

lib/gambit/gambit.sld
chouettevan: Done!

lib/_std#.scm
chouettevan: Done!

doc/gambit.txi
chouettevan: Done!

lib/_std#.scm
chouettevan: it this check worth the (likely tiny or null) performance impact it incurs? is it not expected that primitives segfault or misbehave when given  invalid inputs?

chouettevan: I finally decided to add the check in a separate commit, so we can revert it if the check is to be removed

PR #1001, opened 2026-05-19
SRFI 143

chouettevan:
This PR implements SRFI 143, which is largely present in  gambit. it introduces very little code, but lots of tests to ensure complete spec conformance


gambiteer:
I prefer formatting like
```
(define-procedure (fxcopy-bit (i (index-range-incl 0 fx-width))
                              (j number)
                              (boolean boolean))
  (copy-bit i j boolean))

(define-macro (redefine-2 procedures)
  
  (define (sym . lst)
    (string->symbol
     (apply string-append
            (map (lambda (s) (if (symbol? s) (symbol->string s) s))
                 lst))))
  
  `(begin
     ,@(map (lambda (proc)
              `(define-procedure (,proc (i number)
                                        (j number))  
                 (,(sym "##" proc) i j)))
            procedures)))
```
The way you formatted the code I find it difficult to see at a glance which parts are arguments, type names, etc.

Also, the types of  arguments should be `fixnum` instead of `number`.

I just looked at SRFI 143, and I think you should use `index-range`, not `index-range-incl`.

I could be wrong about one or more of these things, so you need to check what I say.

gambiteer:
The fixnum "constants" depend on the backend, and are defined in `_kernel.scm` and `_univlib.scm`; you should use these instead of defining your own.  This is from `_kernel.scm`:
```
(define ##min-fixnum
  (##c-code "___RESULT = ___FIX(___MIN_FIX);"))

(define ##max-fixnum
  (##c-code "___RESULT = ___FIX(___MAX_FIX);"))

(define ##fixnum-width
  (##c-code "___RESULT = ___FIX(___FIX_WIDTH);"))
```

Define `fxneg` in terms of the builtin `fx-`, not the more general `-`.  This will check that the result is a fixnum.

gambiteer:
I'm not sure `let*-values` works, you should ask Marc.

(Sorry these comments are not more organized.)

Issue #756, opened 2022-07-03
Add unit-tests for all the Gambit predefined procedures


feeley:
This issue has a bounty of $100 CAD.

Many of the Gambit predefined procedures have extensive unit tests in the `tests/unit-tests` directory.  For example each file in the `tests/unit-tests/04-list` directory contains unit tests for a specific list processing procedure or set of related procedures, and `tests/unit-tests/09-io` is for I/O related procedures.

These don't cover all the predefined procedures.  For example `tests/unit-tests/09-io` does not contain unit tests for `print`, `println` and `read-line` (which are rather useful and frequently used predefined procedures).

There needs to be a way to list all the predefined procedures that are tested by the unit tests so that this list can be compared with `lib/gambit/gambit.sld` which has a list of all the procedures exported by Gambit (that are not in the `##` namespace).  This could be done by extending the definitions of the unit testing macros in `tests/unit-tests/#.scm` which are used by all the unit tests.

Unit tests need to be added for each predefined procedure not yet tested.

tzh476:
Hi,

I've been working through the coverage analysis this issue asks for (which predefined procedures are not yet tested), using the exported symbols in `lib/gambit/gambit.sld` (1,532 symbols) against the full `tests/unit-tests/` tree (391 files), doing exact token matching on comment-stripped sources.

Summary of what I found:

- 354 exported symbols never appear anywhere in the test corpus.
- 247 of those 354 are already marked `;;unimplemented;;` in the test tree (360 such markers overall), so they are not missing tests — Gambit does not implement them yet, and writing tests for them would be out of scope here.
- That leaves a small set of implemented-but-untested procedures. The cleanest group is the 8 R7RS `bytevector` names exported at `gambit.sld` lines 274-281 (`bytevector`, `bytevector-append`, `bytevector-copy`, `bytevector-copy!`, `bytevector-length`, `bytevector-u8-ref`, `bytevector-u8-set!`, `bytevector?`), plus `make-bytevector` at line 324: the existing tests only cover the `u8vector` spellings, and none of these 9 names appear (unmarked) anywhere in the test sources.

I have a candidate `tests/unit-tests/07-vector/bytevector.scm` written in the direct-assertion style of the existing `07-vector/u8vector.scm` — 100 assertions covering normal, boundary, and error cases (distinguishing type-exception from range-exception from wrong-number-of-arguments-exception), and not touching the groups covered by the existing open PRs.

[...]

PR #1009, opened 2026-06-22
Add print and println unit tests

Peter1169:
This adds direct `09-io` unit tests for `print` and `println`.

The new tests cover:
- default output port
- explicit `port:` output
- zero, one, and multiple arguments
- invalid output port errors

This is a small step toward #756. It does not touch the tested-procedure listing infrastructure; #797 already explores that broader direction.

Tests:
- `gsi -:=../.. 09-io/print.scm`
- `GAMBIT_COVERAGE=yes gsi -:=../.. 09-io/print.scm`
- `gsi -:=../.. 09-io/display.scm`
- `gsi -:=../.. 09-io/read_line.scm`
- `git diff --cached --check`

Disclosure: I used Codex (GPT-5.5 xHigh) for a final code review pass after human programming.

### DIFF
--- a/tests/unit-tests/09-io/print.scm
+++ b/tests/unit-tests/09-io/print.scm
@@ -13,6 +13,11 @@
  (with-output-to-string (lambda () (print 1 2 3))))
 
 (test-equal
+ "1234567"
+ (with-output-to-string
+  (lambda () (print 1 (list (cons 2 3) (vector 4 5) 6) 7))))
+
+(test-equal
  "123"
  (call-with-output-string (lambda (port) (print port: port 1 2 3))))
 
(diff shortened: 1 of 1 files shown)


feeley:
Thanks for your contribution! You should add a test for one of the useful features of `print`, flattening the data that is printed:

```
> (println 1 (list (cons 2 3) (vector 4 5) 6) 7)
1234567
```

This is useful for generating text that is logically the concatenation of many strings, numbers, etc. Instead of using `string-append`, which is typically cumbersome and slow, a concatenation of the texts X and Y can be written `(list X Y)` if it is eventually written out with `print`, or converted to a string with `(let ((z (list X Y))) (with-output-to-string (lambda () (print z))))`.

Peter1169:
Thanks, that makes sense. I added explicit flattening coverage for both print and println using the nested list/pair/vector case. Validated with 09-io/print.scm, coverage for print.scm, 09-io/display.scm, 09-io/read_line.scm, and git diff --check.

feeley:
Thanks for your first accepted PR!

PR #1015, opened 2026-06-23
tests rewritten to use SRFI 64 form

chouettevan:
This PR finishes the rewrite of the unit-tests present in gambit to use SRFI 64 compatible procedures


feeley:
Thanks for this large rewrite of the unit-tests!

Issue #760, opened 2022-08-04
Bounty offer for bug fixes in the SMP support


adam-7:
For fixes to bugs in the SMP support, 15x 100 EUR bounties are allocated.

Bounty is due when validated that the bug was real and the solution is adopted by Gambit: ie an official Gambit function per the Gambit manual, that works in unicore Gambit and did not work in SMP Gambit, and with the claimant’s fix the bug is solved, and the particular PR offered is viewed as reference grade so much that it’s accepted for adoption to Gambit.

Method for transfer of the bounty eg paypal, bank, cash, crypto etc. discussed when bounty payment is due. No particular expiry date on this offer eg 10 years (1 Jan 2032).

(Offer is nonbinding in the sense subject to current sanctions etc applicable regulations.)

Supposedly there are some easy to find and solve bounties in the SMP support now. If one person intends to claim more than 2-3 bounties please discuss with donor in advance. Bigger bounties are possible. A bit down the line difficulty may go up and the bounty offer may be raised.

This issue should track the status of this bounty offer.

ober:
I'd like to put in for the first bounty as donor.

Mirochill:
Before investigating or preparing a patch, could a maintainer confirm that this bounty offer is still accepted for current Gambit SMP fixes, and could the sponsor confirm how many awards remain and that a qualifying accepted merge can be paid directly via PayPal? Please also point to the current SMP validation command or a bounded unclaimed regression if one should be used for acceptance.

op-simoneromeo:
I found a focused SMP startup/GC bug and prepared a minimal patch for review.

Patch: https://gist.github.com/op-simoneromeo/d01d7b2ca1803efcad59b755e664e79e
Raw patch: https://gist.githubusercontent.com/op-simoneromeo/d01d7b2ca1803efcad59b755e664e79e/raw/9929e1a3c5f7970baa19c17f5500c359ad119759/gambit-smp-heap-lower-bound.patch

Summary: `resize_heap` computes `lower_bound_nb_sections = compute_nb_msections_min(target_processor_count)`, but currently enforces that lower bound only inside the `max_heap > 0` path. With default heap settings, a resize to 4 processors can leave too few msections for the processor count. The next GC then exhausts the free msection list while assigning stack/heap msections during startup and crashes before user code runs. The patch applies the lower bound before the optional max-heap clamp.

Repro config:
[...]

PR #1022, opened 2026-09-06
Fix SMP thread priority setter hangs

artoshirei:
Calling `thread-base-priority-set!` or `thread-priority-boost-set!` followed by its getter hangs in an SMP build, while the same documented calls return normally in unicore. For example:

```scheme
(thread-base-priority-set! (current-thread) 12.3)
(write (thread-base-priority (current-thread)))
(newline)
(exit)
```

The SMP setters acquire the thread lock but do not release it. The priority downgrade loop also recurses when it reaches the thread sentinel, instead of advancing only while visiting owned queues. This patch releases the lock on both setter paths, preserves the void return value, and fixes that loop's continuation branch.

The regression test covers setter return values and readback, increasing/decreasing/unchanged priorities, current/not-started/finished threads, execution inside a worker, and a current thread owning two mutexes.

Validation on Apple Silicon macOS 26.5.2, source `ab6f255c3ca8757816549ff092829d46896927b3`:

* Rebuilt with `--enable-smp --enable-multiple-threaded-vms` and `-O1`.
* Original setter/getter reproductions: four SMP timeouts before; all nine unicore, SMP p1 and SMP p2 setter/control checks pass after.
* All 25 existing non-stress files in `tests/unit-tests/06-thread`, plus the new regression: 78/78 interpreted runs pass across unicore, SMP p1 and SMP p2.
* The new regression times out on the original SMP binary at p1 and p2; it passes after the fix, both interpreted and compiled to native code.
* SMP processes use `-:p1,m64M` or `-:p2,m64M` to avoid the independent default-heap startup problem. Each regression child was bounded by an external timeout.

This fixes the reproduced hangs; it does not claim to complete the existing SMP priority-inheritance and rescheduling TODOs, or validate all concurrent scheduler behavior. Generated C files are omitted, following the repository's source workflow.

AI-assisted implementation and testing by Codex on behalf of Argo. Related to the ordinary SMP development bounty in #760. Please consider this as one candidate submission, subject to maintainer adoption and sponsor confirmation of an available award. PayPal would be the preferred payment method; no award or payment agreement is assumed.


artoshirei:
@feeley @adam-7 Could we move my priority-setter fix in #1022 to review? The patch and regression tests are ready: the original SMP calls hang, and the report includes 78 passing thread-test runs across unicore and SMP after the fix.

@feeley, is there a specific change or additional check you need before adoption? @adam-7, can you confirm whether adoption of this fix would qualify for one EUR100 award, payable through PayPal?

I'd like to finish this submission before taking on another bug.

Argo

PR #1024, opened 2026-09-07
Fix concurrent SMP thread-group insertion

cainandanton:
Concurrent calls to `make-thread-group` with the same parent corrupt the
parent's child-group deque in SMP builds. Each insertion updates shared deque
pointers without taking the parent's existing low-level lock.

This change takes that lock around the insertion and adds a regression test
that creates child groups from four concurrent workers.

Before the fix, an 8-worker stress reproduction created and retained 80,000
groups independently, while `thread-group->thread-group-list` found only 137.
After the fix, it found all 80,000 groups in 10 consecutive runs.

Validation:

- Regression test passes with 2, 4, and 8 VM processors.
- Regression test passes through its non-SMP branch in a unicore build.
- SMP `06-thread` suite: 25/26 pass. The existing
  `mutex_race_timeout.scm` test exited with signal 11; this patch does not
  touch mutex timeout handling.
- `git diff --cached --check` passed before commit.

Related to the SMP fix bounty in #760. I would like this fix considered for
one EUR 100 award if it is accepted and adopted.

The investigation and patch were AI-assisted and validated against a locally
built SMP runtime on macOS arm64.


cainandanton:
Superseded by #1026, submitted from the correct contributor account.

PR #1029, opened 2026-09-07
Fix abandoned-mutex exceptions in SMP mutex-lock!

dominiktaux:
`mutex-lock!` in the SMP runtime silently returns `#t` when acquiring an abandoned mutex for a live thread or for anonymous ownership. The manual (`doc/gambit.txi`, the `mutex-lock!` entry) requires raising `abandoned-mutex-exception` after installing the requested owner whenever the mutex was abandoned before acquisition.

The slow path already saves the old owner/state in `owner`, but selects the exception using the newly installed `state`. This changes that comparison to `owner`, preserving the existing ownership updates and lock release order. The new regression covers default, explicit live, terminated, anonymous, and finite-timeout acquisitions, checks ownership after the exception, and verifies successful reuse after unlock.

For example:

```scheme
(define m (make-mutex))
(define owner (make-thread (lambda () #f)))
(mutex-lock! m #f owner)
(thread-terminate! owner)
(write
 (with-exception-catcher
  (lambda (exc)
    (if (abandoned-mutex-exception? exc) 'abandoned (raise exc)))
  (lambda () (mutex-lock! m))))
(newline)
(write (eq? (mutex-state m) (current-thread)))
(newline)
(mutex-unlock! m)
```

Unicore prints `abandoned` and `#t`. Before the fix, SMP prints `#t` and `#t`; after the fix it agrees with unicore.

Validation on macOS ARM64 with Apple Clang 17, based on `ab6f255c3ca8757816549ff092829d46896927b3`:

| Check | Before | After |
| --- | --- | --- |
| New regression, unicore, interpreted and compiled | Pass | Pass |
| New regression, SMP p1/p2/p4, interpreted and compiled | Six exception assertions fail in each run | Pass |
| All 26 non-stress thread test files, unicore, interpreted | 26/26 | 26/26 |
| All 26 non-stress thread test files, SMP p2, interpreted | 25/26, only new regression fails | 26/26 |

The complete 26-file thread directory also passes compiled after the change in both unicore and SMP p2. The runtime processor-count primitive reports 1, 2, and 4 respectively for the SMP runs.

Configurations:

- Unicore: `./configure --enable-c-opt=-O1 --prefix=<unicore-prefix>`
- SMP: `./configure --enable-smp --enable-multiple-threaded-vms --enable-c-opt=-O1 --prefix=<smp-prefix>`

All runtime C was regenerated from the current Scheme sources using a unicore host compiler rebuilt from the current compiler Scheme sources. Both configurations completed `make bootclean` followed by `make -j4 core`; after the source change, both completed another `make -j4 core`. The repository's stale release-generated C and dummy `gsc-boot` were not used as final runtime evidence.

Run each test mode separately from `tests/`, for example:

```sh
GAMBOPT=p2,m64M ../gsi/gsi -:tl,~~bin=../bin,~~lib=../lib,~~include=../include -f ./run-unit-tests.scm -gsi unit-tests/06-thread
GAMBOPT=p2,m64M ../gsi/gsi -:tl,~~bin=../bin,~~lib=../lib,~~include=../include -f ./run-unit-tests.scm -gsc unit-tests/06-thread
```

For the focused p1/p4 checks, replace `p2` and use `unit-tests/06-thread/mutex_abandoned.scm`. Unicore uses `GAMBOPT=m64M`. The explicit heap setting avoids the separate SMP default-startup heap issue. Full `make check`, stress tests, and other platforms were not run locally.

This contribution was developed and tested with OpenAI Codex on behalf of @dominiktaux. It is submitted for consideration for one of the fixed EUR 100 awards advertised in #760, subject to adoption and confirmation that an award remains available. Award availability and eligibility have not been confirmed. If approved, payment can be arranged using the direct payment methods stated in #760; payout details would be shared privately.


dominiktaux:
@adam-7, could you confirm that one EUR 100 award under #760 remains available for this specific fix if the Gambit maintainers accept it as reference-grade? The PR includes the documented unicore/SMP difference, a minimal correction, and before/after interpreted and compiled test results. The work is AI-assisted as disclosed in the description; please also confirm that it is eligible under the offer. No award or payment is being claimed as approved.

